PR #161, opened 2023-12-21
Fix wrong database name in document.

MOTOO11:


tomohisa (approved):
Thank you for fixing my mistake!

PR #169, opened 2023-12-26
I made it so that a default instance is created even when the type is IEnumerable.

MOTOO11:


tomohisa (approved):
Looks Good, thanks. Let's see if test passes.

PR #198, opened 2024-01-04
Give default values to the public properties of the SekibanAzureOption class

### LINKED ISSUE #196: Suggestion: Give default values to the public properties of the SekibanAzureOption class
Some of the properties of `Sekiban.Infrastructure.Cosmos.SekibanAzureOption` class have default values defined. However, some properties without default values also have default values passed at runtime. To add uniformity to the coding, how about defining the default values in the property declaration whenever possible if the property has a default value?

The following properties are likely to be able to define default values.
- BlobConnectionStringName
- CosmosConnectionStringName
- CosmosDatabase

tkawae:
## What type of PR is this? (check all applicable)

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related Tickets & Documents

- Related Issue #196 
- Closes #196 

## QA Instructions, Screenshots, Recordings

I wrote a sample program and confirmed that it works without any problems.

## Added/updated tests?
_We encourage you to keep the code coverage percentage at 80% and above._

- [ ] Yes
- [x] No, and this is why: I have only refactored the code a little. The changes are verifiable with the existing tests.
- [ ] I need help with writing tests

### DIFF
--- a/src/Sekiban.Infrastructure.Cosmos/SekibanAzureOption.cs
+++ b/src/Sekiban.Infrastructure.Cosmos/SekibanAzureOption.cs
@@ -1,23 +1,32 @@
+using LanguageExt.TypeClasses;
 using Microsoft.Extensions.Configuration;
 using Sekiban.Core.Setting;
 namespace Sekiban.Infrastructure.Cosmos;
 
 public record SekibanAzureOption
 {
+    public static readonly string CosmosEventsContainerDefaultValue = "events";
+    public static readonly string CosmosEventsContainerDissolvableDefaultValue = "dissolvableevents";
+    public static readonly string CosmosItemsContainerDefaultValue = "items";
+    public static readonly string CosmosItemsContainerDissolvableDefaultValue = "dissolvableitems";
+    public static readonly string CosmosConnectionStringNameDefaultValue = "SekibanCosmos";
+    public static readonly string CosmosDatabaseDefaultValue = "SekibanDb";
+    public static readonly string BlobConnectionStringNameDefaultValue = "SekibanBlob";
+
     public string Context { get; init; } = SekibanContext.Default;
 
-    public string CosmosEventsContainer { get; init; } = "events";
-    public string CosmosEventsContainerDissolvable { get; init; } = "dissolvableevents";
-    public string CosmosItemsContainer { get; init; } = "items";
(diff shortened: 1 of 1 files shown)


src/Sekiban.Infrastructure.Cosmos/SekibanAzureOption.cs
```diff
@@ -5,19 +5,21 @@ namespace Sekiban.Infrastructure.Cosmos;
...
 
-    public string? BlobConnectionString { get; init; }
-    public string? BlobConnectionStringName { get; init; }
-
-    public string? CosmosConnectionString { get; init; }
-    public string? CosmosConnectionStringName { get; init; }
     public string? CosmosEndPointUrl { get; init; }
     public string? CosmosAuthorizationKey { get; init; }
-    public string? CosmosDatabase { get; init; }
+    public string CosmosConnectionStringName { get; init; } = "SekibanCosmos";
```
tomohisa, line 16: I think this is good opportunity to make this as a public const. (Since it was use twice as literal string)

src/Sekiban.Infrastructure.Cosmos/SekibanAzureOption.cs
```diff
@@ -5,19 +5,21 @@ namespace Sekiban.Infrastructure.Cosmos;
...
+    public string? CosmosConnectionString { get; init; }
+    public string CosmosDatabase { get; init; } = "SekibanDb";
```
tomohisa, line 18: I think this is good opportunity to make this as a public const. (Since it was use twice as literal string)

src/Sekiban.Infrastructure.Cosmos/SekibanAzureOption.cs
```diff
@@ -5,19 +5,21 @@ namespace Sekiban.Infrastructure.Cosmos;
...
+
+    public string BlobConnectionStringName { get; init; } = "SekibanBlob";
```
tomohisa, line 20: I think this is good opportunity to make this as a public const. (Since it was use twice as literal string)

tomohisa (changes requested):
Can you please make public const and use it?

tomohisa (approved):
Looks Good to me! thanks. Let's wait for the test to merge.

Issue #656, opened 2025-06-22
Add ACA capability in template


tomohisa:
**Is your feature request related to a problem? Please describe.**
Currently only we can work with app service 

**Describe the solution you'd like**
ACA templates in template under
templates/Sekiban.Pure.Templates/content/Sekiban.Orleans.Aspire/infrastructure

**Describe alternatives you've considered**

**Additional context**

PR #685, opened 2025-07-11
Add ACA capability in template

cyan-drop:
### Changes:
- Added Bicep templates
- Created Dockerfiles for building containers for both backend and frontend
- Added docker-compose.yml for local testing and corresponding documentation

### Deployment Order:
1. Create Resource Group
2. Create Log Analytics workspace
3. Create Azure Container Registry (ACR) to store images
4. Build Docker images for backend and frontend, then push them to ACR
5. Full infrastructure build:
    1. Provision services like Storage, Key Vault, etc.
    2. Create the Container App Environment with VNet and Log Analytics integration
    3. Create the Azure Container App linked to the Container App Environment

### Other Notes:
- Azure Container Apps (ACA) does not support partial updates; redeployment of the full configuration is required even for minor adjustments
- Backend ports are not exposed externally; accessible only internally from the frontend container within the same environment
- Scale-in/Scale-out are configured using replicas, based on connection count and CPU usage

## References:
- https://learn.microsoft.com/en-us/dotnet/orleans/deployment/deploy-to-azure-container-apps
- https://learn.microsoft.com/en-us/dotnet/orleans/deployment/docker-deployment
- #656

### DIFF
--- a/templates/Sekiban.Pure.Templates/content/Sekiban.Orleans.Aspire/OrleansSekiban.ApiService/Dockerfile
+++ b/templates/Sekiban.Pure.Templates/content/Sekiban.Orleans.Aspire/OrleansSekiban.ApiService/Dockerfile
@@ -1,0 +1,27 @@
+FROM mcr.microsoft.com/dotnet/sdk:9.0 AS build
+
+WORKDIR /src
+
+COPY *.sln ./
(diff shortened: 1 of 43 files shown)


tomohisa (approved):
thanks @cyan-drop It looks good!

PR #1054, opened 2026-07-12
Add a non-destructive Cosmos tag index repair service

### LINKED ISSUE #1053: SEK-G4 Cosmos DCB: tag index repair service (bounded scan, checkpoint, dry-run)
# SEK-G4 Cosmos DCB: tag index repair service (bounded scan, checkpoint, dry-run)

## Goal

Add a Cosmos-provider administrative repair service that scans the events
container over a `sortableUniqueId` range, derives the expected tag rows from each
event's `tags` array, and writes missing rows idempotently — with dry-run,
checkpoint/resume, and an auditable report.

## Why This Slice Exists Now

Issue #1046: a crash can leave durable events without their tag rows, making them
invisible to tag-scoped reads and tag projectors. Because every Cosmos event
document carries its complete `tags` array, the tags container is a derivable
index — this slice turns that fact into an operational recovery surface for
existing damage and future crash residue.

## Current Observed State

- No repair or maintenance surface exists in `Sekiban.Dcb.CosmosDb`.
- Missing tag rows are silently invisible: `ReadEventsByTagAsync`,
  `GetLatestTagAsync`, tag projectors, and the tag-consistent actor's concurrency
  baseline all under-report; nothing compares the events and tags containers.
- The events container uses per-event partition keys, so a range scan over
  `sortableUniqueId` is a cross-partition query (RU cost, continuation tokens).

- The deterministic tag-row derivation rule and the idempotent tag-write primitive
  (identical row = success; mismatch = structured corruption error) exist from the
  previous slice (SEK-G2).
- Roll-forward retry in the write path (SEK-G3) may or may not have landed;
  this service must not depend on it.
- **Legacy tag-row identity (discovered in SEK-G2 / PR #1050):** rows written
  before SEK-G2 used a non-deterministic `Guid.NewGuid()` row id and
  `DateTime.UtcNow` createdAt. SEK-G2 derives id=`{eventId}`, pk=`{serviceId}|{tag}`,
  tagGroup=prefix-before-colon, createdAt=event SortableUniqueId timestamp. So a
  legacy row for an `(event,tag)` has a **different id** than the derived row, and
  earlier duplicate re-executions may have left **more than one** legacy row for the
  same `(event,tag)`. Repair must not mistake a legacy-format row for "missing".

## Target Repo / Path / Part

Target paths: `dcb/src/Sekiban.Dcb.CosmosDb dcb/tests`

Target part: new Cosmos-provider administrative service that scans events and repairs missing tag rows

## In Scope

- A provider-specific service (e.g. `CosmosDbTagRepairService`) — NOT added to
  `IEventStore` — exposing repair over a `sortableUniqueId` range.
- Bounded, resumable scan: continuation tokens, opaque caller-held checkpoint
  in/out, configurable parallelism, RU-throttling awareness (429/Retry-After),
  CancellationToken.
- Derivation via the deterministic rule; writes via the idempotent primitive;
  mismatches reported as corruption, never overwritten.
- **Legacy-row recognition (non-destructive):** lookup confined to
  pk=`{serviceId}|{tag}`; a legacy row is matched to the derived row on the
  semantic key using ordinal serviceId/tag comparison and canonical Guid
  parse+compare for eventId. A different event with the same tag has a different
  eventId and never matches. On a match the `(event,tag)` is treated as already
  indexed — no second deterministic-id row is written, and the legacy row is
  **never deleted or rewritten** by G4.
- **Post-match validation:** immutable event-derived fields (sortableUniqueId,
  eventType) are checked against the source event; a mismatch is **corrupt**, not
  legacy-present. tagGroup drift is classified explicitly (corrupt /
  legacy-format-drift), never hidden.
- **Separate legacy comparator:** the strict SEK-G2 `ContentEquals` (compares id
  and createdAt) is NOT reused for legacy rows — legacy Guid id and wall-clock
  createdAt are expected differences, reported as informational migration
  metadata, never corruption. Deterministic-id rows still use strict
  `ContentEquals` (incl. id/createdAt).
- **Multiple legacy rows:** reported only (duplicate set) with a per-key cap and
  overflow reporting; the default G4 run stays non-mutating for them. Destructive
  reduction is SEK-G4b, not G4.
- Dry-run mode (no mutation) and normal mode returning an auditable report, with
  categories separated: **missing / legacy-present / duplicate / corrupt / overflow**
  (plus repaired counts in normal mode).
- Idempotent re-run and concurrent-run behavior, including a concurrent normal
  writer landing between classification and repair.
- Tests: seeded missing rows repaired; dry-run non-mutating; checkpoint/resume
  across interruption; idempotency.
- Docs: service usage, dry-run, checkpointing, and the caveat that reads
  immediately after repair may observe stale state depending on the Cosmos
  consistency level.

## Out Of Scope

- Automatic scheduling (startup/periodic sweep is a follow-up slice).
- Change-feed-based resident processor (future optional).
- Repairing the reverse direction (tag rows whose event is missing).
- **Destructive dedup / migration of legacy rows** (delete, canonicalize, or
  rewrite legacy `Guid.NewGuid()`-id rows). G4 is strictly non-destructive:
  it classifies and reports legacy-present / duplicate rows and repairs only
  truly-missing keys. Any delete/canonicalization lives in the separately
  authorized **SEK-G4b** packet and is never invoked by the G5 sweep.

The child PR must deliver a Cosmos-provider administrative tag-repair service with
a bounded, checkpointable, RU-aware, cancellable scan of the events container that
derives expected tag rows deterministically and repairs missing ones idempotently;
dry-run and auditable reporting are required; corruption (existing row differing
from the derived row) is reported and never overwritten; re-runs and concurrent
runs are idempotent; behavior is covered by tests and documented.

## Acceptance Criteria

- Provider-specific admin service (not on IEventStore) repairs missing tag rows over a sortableUniqueId range using the deterministic derivation + idempotent writes.
- Wiring specified for both registration styles: DI via AddSekibanDcbCosmosDb AND manual/custom-factory construction; the service binds to one (serviceId, events container, tags container) lineage per instance so cross-lineage repair is structurally impossible.
- Docs include least-privilege permission + RU guidance and a per-lineage operational example (operator-only job, not a public endpoint).
- Scan is bounded and resumable (continuation tokens, opaque checkpoint in/out), parallelism-configurable, RU-throttling aware, cancellable.
- Dry-run reports without mutating; normal runs return scanned/missing/repaired/legacy-present/duplicate/corrupt/overflow counts.
- G4 is strictly non-destructive: legacy rows are classified/reported and never deleted, canonicalized, or rewritten (destructive dedup/migration is SEK-G4b).
- Legacy recognition uses ordinal serviceId/tag + canonical Guid eventId comparison, partition-key confined; post-match validation of sortableUniqueId/eventType (mismatch = corrupt, not legacy-present).
- A distinct legacy comparator ignores legacy id/createdAt after a semantic-key match; deterministic-id rows keep strict SEK-G2 ContentEquals (still detects createdAt mismatch).
- Mismatched existing rows are reported as corruption, never overwritten.
- Re-run and concurrent runs are idempotent (including a concurrent writer between classification and repair).
- Tests (per orchestrator L1/L2 review): two events sharing one tag → no false match; Guid formatting/casing normalization; matching key + differing id/createdAt → legacy-present not corrupt, no write; matching key + different sortableUniqueId/eventType → corrupt, no write; multiple legacy rows → duplicate report only, non-mutating; deterministic row strict ContentEquals still detects createdAt mismatch; concurrent writer no duplicate/error. Docs updated (EN + JA).

## Verification

## Related Links

## Knowledge Maintenance

- ADR candidate: none (admin-surface decision recorded in host design doc)
- Diagram candidate: none
- Docs update: yes — repair service usage + post-repair consistency caveat
- Closeout writeback expected: yes — report observed RU cost characteristics in the PR description (input for sweep defaults)

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

Adds `CosmosDbTagRepairService`: an operator tool that scans the events container over a `sortableUniqueId` range, derives each expected `(event, tag)` row with the same deterministic rule the write path uses, and creates the ones that are missing — with dry-run, checkpoint/resume, and an auditable report.

Closes #1053

## Non-destructive by construction

G4 **only ever creates rows that do not exist**. It never deletes, rewrites, or canonicalizes one, and it never invokes G4b.

That is enforced structurally, not by discipline: the repair's store seam (`ICosmosTagRepairStore`) exposes `ReadRowsForEventAsync`, `TryCreateRowAsync`, and `TryReadRowAsync` — **the type cannot express a delete or a replace**. Operators can enforce the same thing at the platform level by granting the job a Cosmos role without `deleteItem`/`replaceItem` on the tags container (documented).

## Legacy row recognition (L1/L2)

Rows written before the deterministic-id scheme sit at a random id, so treating them as "missing" would write a *second* row for a pair that is already indexed. Recognition is:

- **Partition-confined**: the lookup queries only `pk = {serviceId}|{tag}`, so another event's rows are never even read.
- **Semantic key** = (serviceId, tag, eventId): serviceId and tag compared **ordinally**; eventId **parsed to a `Guid` and compared canonically**, so Guid formatting/casing drift still matches. A different event carrying the same tag has a different eventId and can never match.
- **Post-match validation**: `sortableUniqueId`, `eventType`, and `tagGroup` must still agree with the source event. A legacy id excuses the id and the timestamp — **and nothing else**. Drift in any of those is `Corrupt`, not `LegacyPresent`, and `tagGroup` drift is classified explicitly rather than hidden.
- **Separate comparator**: `CosmosLegacyTagRowMatcher` ignores the legacy id/createdAt *only after* a semantic-key match, reporting them as migration metadata. The strict SEK-G2 `CosmosTagIdentity.ContentEquals` (id + createdAt included) still governs rows sitting at the derived id — a drifted `createdAt` there is still corruption.
- **Multiple legacy rows** → `Duplicate`, reported only, with a per-key cap (`MaxRowsPerKey`, default 16) and `Overflow` reporting beyond it.

## Report categories

`Present` / `Missing` / `LegacyPresent` / `Duplicate` / `Corrupt` / `Overflow`, plus `Repaired` in normal mode. **`Missing` is the only category that is ever written.** Counts are exact; only the findings detail list is capped.

## Wiring — both registration styles

`AddSekibanDcbCosmosDbTagRepair()` for DI, or `new CosmosDbTagRepairServiceFactory(context, containerResolver)` for hosts using a custom factory. It is **not** registered by `AddSekibanDcbCosmosDb` alone — an operator asks for it — and it is not on `IEventStore`. An instance binds to one `(serviceId, events container, tags container)` lineage at construction with the service id passed **explicitly** (never pulled from an ambient `IServiceIdProvider`), so cross-lineage repair is structurally impossible.

## RU cost characteristics (closeout writeback — input for the sweep defaults)

- The events scan is **cross-partition** (events are partitioned per event) and paged with continuation tokens; cost scales with the range width and `PageSize` (default 100).
- Each `(event, tag)` key costs **one partition-confined query** against the tags container. So **RU scales with keys, not events**: roughly `events × distinct-tags-per-event` cheap single-partition queries, plus the scan itself. A repair write costs one create.
- `MaxParallelism` (default 4) is the RU-rate dial; `MaxEventsToScan` (default 10,000) bounds one run's total cost, after which the run hands back a checkpoint. **Suggested sweep defaults**: keep `MaxParallelism` low (2–4) so a background sweep yields to live traffic, and size `MaxEventsToScan` to the RU headroom of the quietest window.
- Throttling is expected and handled: Cosmos `Retry-After` is honored **in full** (never shortened by a client cap).

## Checkpointing

The opaque checkpoint carries the **last `sortableUniqueId` reached, not a Cosmos continuation token** — a continuation token is bound to the query that produced it and expires, so it cannot be handed across runs. The scan is ordered by `sortableUniqueId`, which is monotonic, so "everything after the last one I saw" is both exact and durable. Continuation tokens are still used for paging *within* a run, where the query is fixed. (An earlier version threaded both and double-skipped on resume; the resume test caught it.)

## Test plan

- [x] 24 new tests. Per the L1/L2 review list: two events sharing one tag → **no false match** (one `LegacyPresent`, one `Missing`, repaired independently); Guid formatting/casing (`D`/`N`/`B`, upper-cased) still matches; matching key + differing id/createdAt → `LegacyPresent`, **zero create attempts**, legacy row byte-identical afterwards; matching key + drifted `sortableUniqueId` / `eventType` / `tagGroup` → `Corrupt`, no write; multiple legacy rows → `Duplicate`, non-mutating; deterministic-id row with drifted `createdAt` → still `Corrupt` under the strict comparator; concurrent writer landing between classification and repair → no duplicate, no error, counted `Present`.
- [x] Plus: missing row repaired; dry-run writes nothing; re-run idempotent; checkpoint/resume across an interrupted scan (no gap, no re-repair); range honored; overflow cap; cancellation; tag repeated within one event = one key; lineage binding; both registration styles.
- [x] `Sekiban.Dcb.WithResult.Tests` 546 passed / 0 failed
- [x] `Sekiban.Dcb.WithoutResult.Tests` 30 passed — no regressions
- [x] `DcbOrleans.ApiService` (downstream Cosmos consumer) builds
- [x] Docs updated (EN + JA): usage, dry-run, checkpointing, report categories, least-privilege + RU guidance, operator-only job example, and the post-repair consistency-level caveat
- [x] `git diff --check` clean

### DIFF
--- a/dcb/src/Sekiban.Dcb.CosmosDb/Repair/CosmosContainerRepairStores.cs
+++ b/dcb/src/Sekiban.Dcb.CosmosDb/Repair/CosmosContainerRepairStores.cs
@@ -84,10 +84,10 @@ internal sealed class CosmosContainerRepairStore : ICosmosTagRepairStore
         int maxRows,
         CancellationToken cancellationToken)
     {
-        // Partition-confined and event-scoped: rows of other events carrying the same tag are never read.
-        var query = new QueryDefinition("SELECT * FROM c WHERE c.pk = @pk AND c.eventId = @eventId")
-            .WithParameter("@pk", partitionKey)
-            .WithParameter("@eventId", eventId.ToString());
+        // The query is a superset prefilter, confined to the tag's partition. It is NOT what decides which
+        // rows index this event — a row stored with a different Guid rendering must not be able to slip past
+        // the server and be mistaken for a missing row. See CosmosRepairRowQuery.
+        var query = CosmosRepairRowQuery.BuildCandidateQuery(partitionKey, eventId);
 
         var requestOptions = new QueryRequestOptions
         {
@@ -104,7 +104,10 @@ internal sealed class CosmosContainerRepairStore : ICosmosTagRepairStore
         {
             var response = await iterator.ReadNextAsync(cancellationToken).ConfigureAwait(false);
             requestCharge += response.RequestCharge;
-            rows.AddRange(response);
+
+            // The correctness gate: canonical Guid comparison, client-side. Format and casing cannot change
+            // the answer, and anything the prefilter over-returned is rejected here.
+            rows.AddRange(response.Where(row => CosmosRepairRowQuery.IsRowForEvent(row, eventId)));
         }
 
         if (rows.Count > maxRows)
--- a/dcb/src/Sekiban.Dcb.CosmosDb/Repair/CosmosContainerRepairStores.cs
+++ b/dcb/src/Sekiban.Dcb.CosmosDb/Repair/CosmosContainerRepairStores.cs
@@ -1,0 +1,153 @@
+using Microsoft.Azure.Cosmos;
+using Sekiban.Dcb.CosmosDb.Models;
+using System.Net;
+namespace Sekiban.Dcb.CosmosDb.Repair;
+
+/// <summary>
+///     The production events reader: a bounded, continuation-token-paged, cross-partition range scan over
+///     the events container, scoped to one service id.
+/// </summary>
+internal sealed class CosmosContainerRepairEventSource : ICosmosRepairEventSource
+{
+    private readonly Container _container;
+    private readonly string _serviceId;
+
+    public CosmosContainerRepairEventSource(Container container, string serviceId)
+    {
+        _container = container;
+        _serviceId = serviceId;
+    }
+
+    public async Task<CosmosRepairEventPage> ReadEventPageAsync(
+        string? fromSortableUniqueIdExclusive,
+        string? toSortableUniqueIdInclusive,
+        int pageSize,
+        string? continuationToken,
+        CancellationToken cancellationToken)
+    {
+        var sql = "SELECT * FROM c WHERE c.serviceId = @serviceId";
+        if (fromSortableUniqueIdExclusive != null)
+        {
+            sql += " AND c.sortableUniqueId > @from";
+        }
+
+        if (toSortableUniqueIdInclusive != null)
+        {
+            sql += " AND c.sortableUniqueId <= @to";
+        }
+
+        sql += " ORDER BY c.sortableUniqueId";
+
+        var query = new QueryDefinition(sql).WithParameter("@serviceId", _serviceId);
+        if (fromSortableUniqueIdExclusive != null)
+        {
+            query = query.WithParameter("@from", fromSortableUniqueIdExclusive);
+        }
+
+        if (toSortableUniqueIdInclusive != null)
+        {
+            query = query.WithParameter("@to", toSortableUniqueIdInclusive);
+        }
+
+        using var iterator = _container.GetItemQueryIterator<CosmosEvent>(
+            query,
+            continuationToken,
+            new QueryRequestOptions { MaxItemCount = pageSize });
+
+        if (!iterator.HasMoreResults)
+        {
+            return new CosmosRepairEventPage(Array.Empty<CosmosEvent>(), null, 0);
+        }
+
+        var response = await iterator.ReadNextAsync(cancellationToken).ConfigureAwait(false);
+        return new CosmosRepairEventPage(
+            response.ToList(),
+            response.ContinuationToken,
+            response.RequestCharge);
+    }
+}
+
+/// <summary>
+///     The production tag-row store for repair. Every lookup is confined to one tag partition, so a repair
+///     can only ever see the rows of the (service, tag) it is examining.
+///     Note there is no delete or replace here — the type cannot express a destructive operation.
+/// </summary>
+internal sealed class CosmosContainerRepairStore : ICosmosTagRepairStore
+{
+    private readonly Container _container;
+
+    public CosmosContainerRepairStore(Container container) => _container = container;
+
+    public async Task<CosmosRepairRowLookup> ReadRowsForEventAsync(
+        string partitionKey,
+        Guid eventId,
+        int maxRows,
+        CancellationToken cancellationToken)
+    {
+        // Partition-confined and event-scoped: rows of other events carrying the same tag are never read.
+        var query = new QueryDefinition("SELECT * FROM c WHERE c.pk = @pk AND c.eventId = @eventId")
+            .WithParameter("@pk", partitionKey)
+            .WithParameter("@eventId", eventId.ToString());
+
+        var requestOptions = new QueryRequestOptions
+        {
+            PartitionKey = new PartitionKey(partitionKey),
+            MaxItemCount = maxRows + 1 // one over the cap, so overflow is detectable
+        };
+
+        using var iterator = _container.GetItemQueryIterator<CosmosTag>(query, requestOptions: requestOptions);
+
+        var rows = new List<CosmosTag>();
+        var requestCharge = 0.0;
+
+        while (iterator.HasMoreResults && rows.Count <= maxRows)
+        {
+            var response = await iterator.ReadNextAsync(cancellationToken).ConfigureAwait(false);
+            requestCharge += response.RequestCharge;
+            rows.AddRange(response);
+        }
+
+        if (rows.Count > maxRows)
+        {
+            return new CosmosRepairRowLookup(rows.Take(maxRows).ToList(), true, requestCharge);
+        }
+
+        return new CosmosRepairRowLookup(rows, false, requestCharge);
+    }
+
+    public async Task<(bool Created, double RequestCharge)> TryCreateRowAsync(
+        string partitionKey,
+        CosmosTag row,
+        CancellationToken cancellationToken)
+    {
+        try
+        {
+            var response = await _container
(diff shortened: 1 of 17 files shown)


tomohisa:
Blocking implementation finding — accepted from Copilot inline signal discussion_r3565489245 and confirmed against the SEK-G4 L1/non-destructive contract. CosmosContainerRepairStore.ReadRowsForEventAsync filters c.eventId by eventId.ToString() before CosmosLegacyTagRowMatcher can perform canonical Guid parse+compare. Cosmos string equality is format/case-sensitive, so a legacy row stored as uppercase D, N, or B format is not returned, the key becomes Missing, and normal mode creates a deterministic row beside an already-indexed legacy row. That amplifies duplicates and violates the explicit canonical-Guid and Missing-only/non-destructive acceptance criteria. Fix the production lookup so all rows in the derived pk partition that canonically match the event Guid can be classified client-side (while preserving MaxRowsPerKey/Overflow bounds), and add a production-store-level test proving non-D/case variants are retrieved/classified without a create. The current fake-store matcher test does not exercise the SQL prefilter.

tomohisa:
CI repair required — SonarCloud Code Analysis is failing at head db1373615f1143c7493033740fb2c2172300bf3d. The prior semantic blocker is resolved; this request-update is limited to the red quality gate findings below:\n\n1. MAJOR vulnerability S2077 — dcb/src/Sekiban.Dcb.CosmosDb/Repair/CosmosRepairRowQuery.cs:50: parameterize the query rather than constructing predicate text via string formatting.\n2. CRITICAL code smell S1994 — dcb/src/Sekiban.Dcb.CosmosDb/Repair/ThrottleAware.cs:24: rewrite the loop so its condition tests the attempt bound instead of using an unbounded condition while the incrementer updates attempt; preserve Retry-After/cancellation semantics.\n3. MINOR code smell S3458 — dcb/src/Sekiban.Dcb.CosmosDb/Repair/CosmosDbTagRepairService.cs:460: remove the empty case clause.\n\nPlease make the narrow fixes, run the focused repair tests, and return with SonarCloud Code Analysis green. Do not alter the resolved canonical-Guid/non-destructive contract.

PR #1056, opened 2026-07-12
Add an opt-in sweep that repairs recent Cosmos tag residue

### LINKED ISSUE #1055: SEK-G5 Cosmos DCB: opt-in startup sweep + readiness non-guarantee documentation
# SEK-G5 Cosmos DCB: opt-in startup sweep + readiness non-guarantee documentation

## Goal

Add an opt-in hosted sweep that automatically runs the Cosmos tag-index repair
(previous slice) over a recent window at startup and optionally on an interval —
and document, honestly, that the sweep provides EVENTUAL repair only and does not
gate tag readers.

## Why This Slice Exists Now

Issue #1046: manual repair alone leaves routine crash residue unrepaired until an
operator notices. An automatic sweep closes that gap. At the same time, an
arbitrary sweep cannot guarantee repair-before-tag-read (multiple writers/readers,
post-startup crashes, sweep/reader races, Cosmos consistency levels), so the docs
must state the non-guarantee explicitly instead of implying safety.

## Current Observed State

- The repair service (previous slice) exists but must be invoked manually.
- `CosmosDbInitializer` is an existing `IHostedService` suitable for hosting
  background startup work.
- `GeneralTagConsistentActor` rebuilds its optimistic-concurrency baseline from
  the tags container; missing tag rows silently regress that baseline until
  repaired.

- The bounded, idempotent, checkpointable repair service (SEK-G4) is available;
  overlap safety comes from its idempotency.

## Target Repo / Path / Part

Target paths: `dcb/src/Sekiban.Dcb.CosmosDb dcb/tests docs/dcb_llm/11_storage_providers.md docs/dcb_llm_ja/11_storage_providers.md`

Target part: opt-in hosted sweep running the tag repair over a recent window + readiness non-guarantee documentation

## In Scope

- Opt-in (disabled by default) hosted sweep: startup run over a configurable
  recent `sortableUniqueId` window, optional interval runs; options for window
  size, interval, parallelism, bounded startup budget, cancellation on shutdown.
- Failures logged + telemetry, never crashing the host or blocking startup
  indefinitely.
- Idempotent and safe to overlap with normal writes and concurrent manual repair
  (verified by test).
- Documentation (EN + JA): the sweep is eventual repair; it does NOT gate tag
  readers; a missing-tag window can remain; the tag-consistent actor baseline can
  be regressed until repair lands (known limitation); readiness gating / commit
  protocol is explicitly out of scope / future work.
- Tests: startup sweep repairs seeded crash residue; disabled by default; overlap
  safety.

## Out Of Scope

- Readiness gate / read-time verification / commit protocol (future work).
- Change-feed resident processor.

The child PR must add a disabled-by-default hosted sweep that runs the existing
Cosmos tag repair over a configurable recent window at startup (and optionally on
an interval) with bounded startup impact, telemetry, shutdown cancellation, and
overlap safety — plus documentation that states the eventual-repair non-guarantee
and the remaining known limitations exactly as scoped above.

## Acceptance Criteria

- Opt-in sweep (disabled by default) runs repair over a configurable recent window at startup and optionally on an interval, with window/interval/parallelism/startup-budget/cancellation options.
- Package reference/upgrade alone adds no hosted service, network scan, startup delay, or required configuration.
- Opt-in wiring defined for both AddSekibanDcbCosmosDb and manual/custom-factory consumers; sweep options and checkpoints are per lineage (serviceId + container names).
- Guidance: manual dry-run repair + RU observation before enabling interval sweeps; jitter and leader/lease patterns documented for replicated services.
- Sweep failures are logged and surfaced via telemetry; the host never crashes and startup is never blocked indefinitely.
- Sweep runs are idempotent and overlap-safe with normal writes and manual repair (tested).
- **Legacy-row safety:** the automatic sweep is statically restricted to SEK-G4's non-destructive surface (missing-row backfill + classification/telemetry only). Repeated sweeps over legacy-present/duplicate sets create zero rows and delete zero rows; the sweep cannot reach SEK-G4b destructive dedup/migration or any legacy-row delete even by configuration. Corruption/overflow surface via telemetry without attempting migration.
- Docs (EN + JA) explicitly state: eventual repair only, no tag-reader gating, remaining missing-tag window, tag-consistent actor baseline regression until repair, readiness gating is future work.
- Tests cover startup repair convergence, default-off, overlap safety, and no-duplicate-amplification when legacy rows are present.

## Verification

## Related Links

## Knowledge Maintenance

- ADR candidate: none
- Diagram candidate: none
- Docs update: yes — sweep configuration + readiness non-guarantee statement
- Closeout writeback expected: no

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

Adds `CosmosTagSweepService`: a disabled-by-default hosted sweep that runs the existing non-destructive tag repair (#1053 / PR #1054) over a recent `sortableUniqueId` window shortly after startup, and optionally on an interval — plus documentation of the **eventual-repair non-guarantee**.

Closes #1055

## Opt-in twice over

A package reference or upgrade adds **no hosted service, no network scan, no startup delay, and no required configuration**:

1. `AddSekibanDcbCosmosDb` does **not** register the sweep. You call `AddSekibanDcbCosmosDbTagSweep(...)` yourself.
2. Even registered, it stays inert until `CosmosTagSweepOptions.Enabled` is set (default `false`).

Both are covered by tests (`AddSekibanDcbCosmosDb_Alone_Should_Register_No_Hosted_Sweep`, `A_Disabled_Sweep_Should_Not_Touch_Storage` — zero reads).

## Legacy-row safety: structural, not configurable

The sweep's **only** route to storage is `CosmosDbTagRepairService`, whose store seam (`ICosmosTagRepairStore`) exposes exactly `ReadRowsForEventAsync`, `TryCreateRowAsync`, `TryReadRowAsync`. **The type cannot express a delete, a replace, an upsert, or a patch** — so "the sweep cannot be configured into destructiveness" is a property of the types, not a promise in a comment. There is no G4b code path from here, and `CosmosTagSweepOptions` has no mode/dedup/migrate switch to flip.

Two tests pin this structurally by reflection (asserting the seam's member list, and that neither the seam nor the options carry any `Delete`/`Replace`/`Upsert`/`Patch`/`Dedup`/`Migrat`/`Canonical`/`Mode` member), plus behavioral proof: **three consecutive sweeps over a `LegacyPresent` set and over a `Duplicate` set create zero rows and delete zero rows**, and the legacy rows come out byte-identical (id and `createdAt` unchanged). `Corrupt`/`Overflow` surface via telemetry and logs with no migration attempted.

## Bounded, safe, per-lineage

- **Startup is never blocked**: the sweep runs on a background task; `StartAsync` returns immediately (tested with a blocking runner). `RunBudget` (default 5 min) bounds any single run — an overrun stops and resumes from its checkpoint next turn.
- **A failing sweep never takes the host down**: it is logged, counted, and retried on the next turn (tested).
- **Per lineage**: `ServiceIds` selects which lineages to sweep (empty = the host's own); each has its own window and its own checkpoint.
- **Idempotent and overlap-safe** with live writes and with a concurrent manual repair — a row the write path lands between classification and repair is recognized as the row the sweep was about to write, counted `Present`, not duplicated (tested).
- **Jitter**: `MaxStartupJitter` (default 30s) keeps replicas from sweeping in lockstep. Leader/lease guidance is documented for interval sweeps across many replicas.
- **RU**: `MaxParallelism` defaults to **2** so a background sweep yields to live traffic; docs tell operators to run a manual dry-run and observe RU before enabling an interval.

Wiring works for both registration styles: `AddSekibanDcbCosmosDbTagSweep(...)` for DI, and the sweep is constructible directly for hosts that build their stores with a custom factory.

## Telemetry

`tag_sweep.runs` (label `outcome`: `completed` | `budget_exhausted` | `failed`), `tag_sweep.repaired_rows`, `tag_sweep.corrupt_keys`, `tag_sweep.overflow_keys`. Bounded labels only.

## What it does not guarantee (documented EN + JA)

Stated plainly in the docs rather than letting an automatic sweep imply a safety it cannot provide:

- It is **eventual repair only**. It **does not gate tag readers** — nothing waits for a sweep.
- A **missing-tag window remains** open from the crash until a run reaches it.
- **`GeneralTagConsistentActor`'s optimistic-concurrency baseline can be regressed** for that whole window (known limitation).
- Post-repair reads can still be stale depending on the Cosmos consistency level.
- **Readiness gating / read-time verification / commit protocol are future work**, explicitly out of scope. Workloads that cannot tolerate the window should use **Postgres**, which has none of these gaps.

## Test plan

- [x] 16 new tests: default-off; disabled sweep touches nothing; `AddSekibanDcbCosmosDb` alone registers no hosted sweep; startup sweep repairs seeded crash residue; only the recent window is scanned; repeated sweeps idempotent; **legacy rows — 3 sweeps, 0 creates, 0 deletes, rows unchanged**; **duplicate legacy rows — 3 sweeps, not reduced**; concurrent writer → no duplicate/error; failing sweep does not crash the host; startup not blocked; jitter spreads replicas; parallelism/event-budget plumbed through with `DryRun = false` meaning only "create what is missing"; two reflection guards on the non-destructive surface.
- [x] `Sekiban.Dcb.WithResult.Tests` 577 passed / 0 failed — no regressions
- [x] `DcbOrleans.ApiService` (downstream Cosmos consumer) builds
- [x] No analyzer warnings; `git diff --check` clean


tomohisa:
Blocking implementation findings for SEK-G5:\n\n1. Run-budget checkpoint/resume contract is not implemented. When RunBudget cancels the runner, SweepLineageAsync catches OperationCanceledException but receives no CosmosTagRepairReport and does not update _checkpoints (the code explicitly says checkpoint is not advanced). The next interval therefore re-scans from the same window start/current checkpoint rather than resuming from progress made during the overrun. This contradicts the PR/docs claim that an overrun stops and resumes from its checkpoint next turn and the review requirement for run-budget checkpoint/resume. Please provide a bounded-run path that durably returns/captures partial progress before budget exhaustion (or revise the runner/service protocol accordingly), and add a deterministic test proving the next turn starts from the newly advanced checkpoint rather than the prior start.\n\n2. Accepted Copilot docs signal: EN/JA list shorthand metric names (tag_sweep.*), while CosmosDbTelemetry publishes fully qualified sekiban.dcb.cosmos.tag_sweep.* instruments. Update both docs to the exact queryable names so operators can find them.\n\nKeep the resolved G4 Missing-only/non-destructive route, double opt-in, and eventual-only/no-readiness contract unchanged.

Issue #1046, opened 2026-07-10
Cosmos DCB event and tag writes are not crash-atomic


joemphilips:
## Summary

`Sekiban.Dcb.CosmosDb` can persist event documents without the tag rows required for normal tag-scoped replay. This occurs when the process or host stops after the event-write phase and before all tag batches have committed.

This is a DCB correctness issue: the event is present in `ReadAllEventsAsync`, but absent from `ReadEventsByTagAsync`, so tag projectors and tag-consistent actors can reconstruct a state that omits a durable event.

## Affected version

- `Sekiban.Dcb.CosmosDb` 10.2.2 (`dcb-v10.2.2`)
- The same two-phase design is also present in 9.2.0.

## Evidence in the v10.2.2 implementation

`CosmosDbEventStore.WriteEventsAsync` does the following:

1. `WriteEventsInParallelAsync` creates each event in the events container. Event documents use distributed event-id partition keys.
2. `WriteTagsWithBatchAsync` then creates tag records, in transactional batches only within each individual tag partition.
3. Only an ordinary caught tag-write exception triggers `TryRollbackEventsAsync`, which deletes already-written event documents best-effort.

Relevant source:

- https://github.com/J-Tech-Japan/Sekiban/blob/dcb-v10.2.2/dcb/src/Sekiban.Dcb.CosmosDb/CosmosDbEventStore.cs#L423-L480
- https://github.com/J-Tech-Japan/Sekiban/blob/dcb-v10.2.2/dcb/src/Sekiban.Dcb.CosmosDb/CosmosDbEventStore.cs#L510-L705

Cosmos transactional batches cannot cover the distributed event partitions and the separate tag container. Therefore a host termination in that window bypasses the in-process compensation path. A partial tag write can also leave only some tag associations visible.

## Reproduction scenario

1. Configure a DCB Cosmos event store and submit a command that emits one event with one or more tags.
2. Allow the event `CreateItemAsync` calls to complete.
3. Terminate the process before `WriteTagsWithBatchAsync` has completed every tag batch.
4. Restart and compare reads:
   - `ReadAllEventsAsync` returns the event.
   - `ReadEventsByTagAsync` for an affected tag does not return it.

A deterministic provider-level fault-injection test at that boundary would demonstrate this without relying on timing.

## Expected behavior / requested contract

For a successful-or-crashed `WriteEventsAsync`, DCB needs one of these guarantees:

[...]

PR #1058, opened 2026-07-12
Add the crash-window fault-injection E2E matrix

### LINKED ISSUE #1057: SEK-G6 Cosmos DCB: crash-window fault-injection E2E test matrix
# SEK-G6 Cosmos DCB: crash-window fault-injection E2E test matrix

## Goal

Deliver the consolidated, deterministic fault-injection E2E test matrix for the
Cosmos DCB crash window — verifying that the contracts implemented in the
preceding slices (idempotent tag writes, roll-forward retry, repair service,
startup sweep, documented non-guarantees) hold end to end.

## Why This Slice Exists Now

Issue #1046 explicitly requested a deterministic provider-level fault-injection
test at the event/tag write boundary. Individual slices carry their own scoped
tests; this slice closes the loop with cross-cutting crash → restart → repair →
convergence scenarios and asserts that documented behavior matches real behavior.

## Current Observed State

Before this work, the Cosmos write/rollback path had zero test coverage. The
preceding slices added scoped tests and a deterministic fault-injection seam for
the tag-write stage; the cross-scenario matrix does not exist yet.

- Fault-injection seam (SEK-G2), roll-forward retry + partial-failure contract
  (SEK-G3), repair service (SEK-G4), and startup sweep (SEK-G5) are available.
- Scenarios covering only SEK-G2..G4 behavior may already have been implemented
  early alongside those slices; do not duplicate them — consolidate and fill gaps.

## Target Repo / Path / Part

Target paths: `dcb/tests`

Target part: end-to-end fault-injection matrix for the Cosmos crash window and repair convergence

## In Scope

- Scenario: events written, crash before any tag batch → after restart, repair converges; tag-scoped reads see the event.
- Scenario: crash mid-tag-batches → repair converges without duplicates.
- Scenario: multi-event partial event-create success → no deletion; structured error identifies the partially-visible set (contract from the retry slice).
- Scenario: same-tag re-execution has no duplicates; content-mismatched rows are detected as corruption.
- Scenario: repair re-run and concurrent runs are idempotent.
- Scenario: large scan checkpoint/resume under RU throttling and cancellation.
- Scenario: repair completion racing tag readers — observed behavior matches the documented eventual-repair non-guarantee (assert nothing stronger).
- Consumer-contract scenarios for both downstream registration styles (AddSekibanDcbCosmosDb and manual options + store construction).
- Compatibility assertion: disabled sweep + legacy defaults → no new background work, no behavior change after upgrade.
- Two-lineage isolation scenario: repair/sweep cannot cross management/runtime container lineages.
- All scenarios deterministic via the seam (no timing dependence).

## Out Of Scope

- New runtime features (tests only; real defects found are reported as separate bugs).
- A permanent live Cosmos emulator CI harness (optional).

The child PR must add a deterministic, seam-based E2E test matrix covering the
seven scenario families above, consolidated with (not duplicating) the scoped
tests that landed with earlier slices, all green and free of timing dependence.

## Acceptance Criteria

- All seven scenario families implemented and green, deterministic via the fault-injection seam.
- No duplication of earlier scoped tests — consolidation and gap-filling only.
- Any real defect discovered is filed separately and referenced, not silently fixed in this PR.

## Verification

Full matrix green in CI; no regression; `git diff --check` clean.

## Related Links

## Knowledge Maintenance

- ADR candidate: none
- Diagram candidate: none
- Docs update: none (unless a defect changes a documented contract — handled with the fix)
- Closeout writeback expected: yes — report which scenarios exposed real defects (input for closing issue #1046)

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

The consolidated, deterministic fault-injection matrix for the Cosmos crash window. Tests only — **no `dcb/src` changes**.

The scoped tests that landed with SEK-G2..G5 check each piece in isolation. This checks the thing those pieces exist for: after a crash in the two-phase write, **a tag-scoped reader eventually sees the event** — and nothing stronger than "eventually" is true, because that is all the documentation claims.

Closes #1057

## The harness

An in-memory Cosmos (`InMemoryCosmosClient` / `InMemoryCosmosDatabase` / `InMemoryCosmosContainer` / transactional batch) so the **real** `CosmosDbEventStore`, the **real** `CosmosDbTagRepairService`, and the **real** sweep run end to end with no emulator.

It models the two behaviors the crash contracts hang on — a create **conflicts** on an existing `(pk, id)`, and a transactional batch is **all-or-nothing** — and it **refuses to answer a query shape it does not recognize** rather than returning nothing and turning a real bug into a green test. (That failure mode is exactly what bit PR #1054: a fake that was more forgiving than Cosmos.)

Faults go through the seams (`ICosmosTagWriteFaultInjector`, injected `ICosmosRetryScheduler`, an `OnWrite` hook to act on the Nth write). **Nothing sleeps, nothing polls, nothing depends on timing.**

## Scenarios (20)

**Crash window, through the real store and the real repair (16)** — `CosmosCrashWindowE2ETests`

| Family | What it proves |
|---|---|
| Crash **before any** tag batch | Event durable, not deleted, no tag rows. All-events reader sees it; **tag reader does not**. After repair: tag reader sees it. |
| Crash **mid** tag batches | The row that landed is recognized (`Present`), the two that did not are backfilled — 3 rows, no duplicates. |
| Roll-forward retry | A transient tag failure retried through the real store converges to exactly 3 rows, deletes nothing. |
| Multi-event partial create | `CosmosPartialEventWriteException` names both sets; **zero deletes**. |
| Re-execution / corruption | Same write replayed → no duplicate rows. A row disagreeing with its event → `Corrupt`, never overwritten. |
| Repair idempotency | Two repairs raced: exactly one writes (`Repaired`=1), the other finds it (`Present`=1), neither reports corruption. A third finds nothing to do. |
| Checkpoint / resume | Bounded runs resume with no gap and no re-repair, **through a 429 whose Retry-After is honored**. A cancelled run keeps exactly the progress it settled (3 of 4), and resuming does the remaining 1. |
| **Reader racing repair** | Before repair: `ReadEventsByTag` empty, `TagExists` false, `GetLatestTag` empty — the tag-consistent baseline is regressed — while all-events readers see the event throughout. Only after repair does it converge. **Asserts the documented window, rather than wishing it away.** |
| Both registration styles | DI and manual/custom-factory both reach the same non-destructive surface; the manual path (what SekibanWasmRuntime / SekibanAsAService use) gets the same end-to-end repair proof. |
| Legacy defaults | Unchanged after upgrade: no retry, legacy rollback still deletes. |
| Two-lineage isolation | Repairing the runtime lineage leaves the management lineage **untouched** (zero creates); each repairs only itself. |
| Dry run | Reports the residue, writes nothing. |

**The sweep, through the types that actually run (4)** — `CosmosSweepE2ETests`

The sweep's own tests substitute a fake runner, which proves the scheduling and nothing about whether the wiring reaches storage. These drive the **registered `CosmosTagSweepService`** through the **real `CosmosTagRepairRunner`**, the **real repair service**, and the harness — nothing faked below the sweep. Only `ISweepClock` is substituted; turns are driven by starting the service, and the budget expiry is injected at an exact write.

| Scenario | What it proves |
|---|---|
| DI wiring reaches production repair | `AddSekibanDcbCosmosDbTagSweep(Enabled = true)` → the hosted service the host would start → real runner → real repair → the missing rows are written, and a tag-scoped read then sees the event. |
| Budget-exhausted turn resumes | Turn 1 cancelled at the 3rd tag write (2 settled); turn 2 **resumes from the kept checkpoint** and finishes the remaining 2; turn 3 starts from the window again (checkpoint cleared) and finds all 4 present, creating nothing. |
| Lineage isolation through sweep | Sweeping runtime repairs its rows and leaves management with **zero creates**; sweeping management repairs only management. Every row sits under its own `serviceId` partition prefix. |
| Disabled default | Both containers pinned on both dimensions: **no queries, no creates, no deletes**, either side. |

## Do these tests actually bite?

**Mutation-checked, and it earned its keep.**

- Disabling the repair's write → **8 of the 16** crash-window scenarios fail.
- Making the sweep's runner never reach repair → **3 of the 4** sweep scenarios fail.
- Removing the sweep's `Enabled` guard → the disabled-default test fails.
- Dropping the budget-exhausted checkpoint → the resume test fails **(only after it was fixed — see below)**.

## Defects revealed (closeout writeback)

**Zero new product defects.** All twenty scenarios pass against `main` as it stands. The contracts from SEK-G1..G5 hold end to end.

Two distinct classes of defect are worth separating in the record for #1046:

**1. Three real product defects — caught by review, not by the scoped tests.** Each slipped through because a test exercised a fake instead of the production path:
- **G3**: a Cosmos 429 `Retry-After` was capped by the client's `MaxBackoff`, so the client retried before the server was ready.
- **G4**: the production SQL matched legacy rows by the *string form* of the event id, so a row stored in another Guid rendering looked *missing* and would have been duplicated. The unit test passed because its fake parsed Guids client-side.
- **G5**: a budget-exhausted sweep turn threw its partial checkpoint away, so a tight budget re-scanned the same prefix forever and never advanced.

**2. One defect in this PR's own coverage — caught by the mutation check.** The first version of the sweep-resume test asserted only the final row count, and **passed with the checkpoint persistence deleted**: a turn that drops the checkpoint re-scans from the window, finds the first two present, repairs the last two, and converges on the same four rows. The test proved nothing. It now counts the *work* instead — the repair issues one tag lookup per key it examines, so a resumed turn examines 2 and a restarted one examines 4 — and the mutant now fails it. That is a test defect, not a product defect, and the class of mistake this matrix exists to stop.

## Test plan

- [x] 20 new E2E scenarios (16 crash-window + 4 sweep), all green, deterministic
- [x] Mutation-checked: disabling repair writes fails 8; breaking the sweep runner fails 3; removing the Enabled guard fails the disabled test; dropping the budget checkpoint fails the resume test
- [x] `Sekiban.Dcb.WithResult.Tests` 600 passed / 0 failed — no regressions
- [x] No `dcb/src` changes (tests only, per the packet)
- [x] No warnings; `git diff --check` clean

### DIFF
--- a/dcb/tests/Sekiban.Dcb.WithResult.Tests/Cosmos/CosmosCrashWindowE2ETests.cs
+++ b/dcb/tests/Sekiban.Dcb.WithResult.Tests/Cosmos/CosmosCrashWindowE2ETests.cs
@@ -1,0 +1,631 @@
+using Dcb.Domain;
+using Microsoft.Extensions.DependencyInjection;
+using Sekiban.Dcb.Common;
+using Sekiban.Dcb.CosmosDb;
+using Sekiban.Dcb.CosmosDb.Models;
+using Sekiban.Dcb.CosmosDb.Repair;
+using Sekiban.Dcb.CosmosDb.Sweep;
+using Sekiban.Dcb.CosmosDb.Tags;
+using Sekiban.Dcb.Events;
+using Sekiban.Dcb.ServiceId;
+using Sekiban.Dcb.Tags;
+
+namespace Sekiban.Dcb.Tests.Cosmos;
+
+/// <summary>
+///     The crash-window matrix: crash → restart → repair → convergence, driven end to end through the real
+///     <see cref="CosmosDbEventStore" />, the real <see cref="CosmosDbTagRepairService" />, and the real
+///     sweep, against in-memory Cosmos containers.
+///     The scoped tests that landed with the earlier slices check each piece in isolation. These check the
+///     thing those pieces exist for: that after a crash in the two-phase write, a tag-scoped reader
+///     eventually sees the event — and that nothing stronger than "eventually" is true, because that is all
+///     the documentation claims.
+///     Everything here is deterministic. Faults are injected through the seams; nothing sleeps, and nothing
+///     depends on timing.
+/// </summary>
+public class CosmosCrashWindowE2ETests
+{
+    private const string ServiceId = "svc";
+    private const string OtherServiceId = "other";
+
+    /// <summary>One service's containers plus the store that writes to them.</summary>
+    private sealed class Lineage
+    {
+        public Lineage(string serviceId, InMemoryCosmosClient client, CosmosDbEventStoreOptions options)
+        {
+            ServiceId = serviceId;
+            Client = client;
+            Options = options;
+
+            var context = new CosmosDbContext(client, "test-db", null, options);
+            var resolver = new DefaultCosmosContainerResolver(options);
+            Store = new CosmosDbEventStore(
+                context,
+                DomainType.GetDomainTypes().EventTypes,
+                new FixedServiceIdProvider(serviceId),
+                resolver);
+            RepairFactory = new CosmosDbTagRepairServiceFactory(context, resolver);
+        }
+
+        public string ServiceId { get; }
+        public InMemoryCosmosClient Client { get; }
+        public CosmosDbEventStoreOptions Options { get; }
+        public CosmosDbEventStore Store { get; }
+        public CosmosDbTagRepairServiceFactory RepairFactory { get; }
+
+        public InMemoryCosmosContainer Events => Client.Container(Options.EventsContainerName);
+        public InMemoryCosmosContainer Tags => Client.Container(Options.TagsContainerName);
+
+        public Task<CosmosDbTagRepairService> RepairAsync() => RepairFactory.CreateAsync(ServiceId);
+    }
+
+    private sealed class FixedServiceIdProvider : IServiceIdProvider
+    {
+        private readonly string _serviceId;
+
+        public FixedServiceIdProvider(string serviceId) => _serviceId = serviceId;
+
+        public string GetCurrentServiceId() => _serviceId;
+    }
+
+    /// <summary>Fails the tag write before batch N, deterministically — the seam, not a race.</summary>
+    private sealed class FailBeforeBatch : ICosmosTagWriteFaultInjector
+    {
+        private readonly int _batchIndex;
+
+        public FailBeforeBatch(int batchIndex) => _batchIndex = batchIndex;
+
+        public Task OnBeforeBatchAsync(int batchIndex, string partitionKey, IReadOnlyList<CosmosTag> rows)
+        {
+            if (batchIndex >= _batchIndex)
+            {
+                throw new InvalidOperationException($"Injected crash before tag batch {batchIndex}");
+            }
+
+            return Task.CompletedTask;
+        }
+    }
+
+    private static CosmosDbEventStoreOptions NewOptions() =>
+        new()
+        {
+            EventsContainerName = "events",
+            TagsContainerName = "tags",
+            // Roll-forward with a single attempt reproduces exactly what a crash leaves behind: the events
+            // are durable and nothing is deleted, but the tag write never completed.
+            WriteFailurePolicy = CosmosWriteFailurePolicy.RollForward,
+            TagWriteRetry = new CosmosTagWriteRetryOptions { MaxAttempts = 1, JitterRatio = 0 }
+        };
+
+    private static Lineage NewLineage(string serviceId = ServiceId, CosmosDbEventStoreOptions? options = null) =>
+        new(serviceId, new InMemoryCosmosClient(), options ?? NewOptions());
+
+    private static SerializableEvent NewEvent(params string[] tags) =>
+        new(
+            System.Text.Encoding.UTF8.GetBytes("""{"Name":"test"}"""),
+            SortableUniqueId.GenerateNew(),
+            Guid.NewGuid(),
+            new EventMetadata("causation", "correlation", "user"),
+            tags.ToList(),
+            "TestEventPayload");
+
+    private sealed record TestTag(string Value) : ITag
+    {
+        public bool IsConsistencyTag() => false;
+        public string GetTagGroup() => Value.Split(':')[0];
+        public string GetTag() => Value;
+        public string GetTagContent() => Value.Split(':')[1];
+    }
+
+    private static async Task<CosmosTagRepairReport> RunRepairAsync(Lineage lineage, bool dryRun = false)
+    {
+        var repair = await lineage.RepairAsync();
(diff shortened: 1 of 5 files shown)


tomohisa:
Blocking review finding — SEK-G6 sweep-path acceptance evidence is missing.

Failed acceptance criterion / packet goal: the packet requires the consolidated matrix to cover the startup sweep, both downstream registration styles, and two-lineage repair/sweep isolation. The review focus also requires the real CosmosDbEventStore, repair service, and sweep paths to be exercised rather than bypassed by test doubles.

Current evidence: all 16 new tests pass on net9.0 and net10.0, and they exercise the real CosmosDbEventStore and CosmosDbTagRepairService. However, CosmosCrashWindowE2ETests never constructs or starts CosmosTagSweepService and never routes through CosmosTagRepairRunner. A_Disabled_Sweep_Should_Add_No_Background_Work only verifies the disabled/no-registration default; Repair_Should_Not_Cross_Lineages calls repair directly. Therefore the consolidated gate cannot detect broken enabled-sweep DI wiring, serviceId/container lineage routing, or checkpoint/resume propagation through the hosted sweep.

Required unblock action: add deterministic test coverage that enables the registered sweep and drives the real sweep/runner path against the in-memory Cosmos containers. It must prove (1) enabled DI registration reaches the production repair path, (2) checkpoint is carried into a later sweep turn and cleared at completion, and (3) two serviceId/container lineages remain isolated through sweep execution. Keep it seam/virtual-scheduler driven with no sleep or polling. The existing disabled-default and direct-repair tests can remain as separate compatibility evidence.

False-claim boundary: the current green matrix verifies repair convergence but does not verify enabled hosted-sweep correctness or sweep lineage isolation.

tomohisa:
Re-review follow-up — the production sweep blocker is substantially resolved, but two acceptance-evidence gaps remain.

1. Disabled sweep zero-I/O assertion is still incomplete. `A_Disabled_Sweep_Should_Not_Touch_The_Containers` snapshots/asserts `lineage.Events.Queries` and asserts zero tag creates, but never snapshots/asserts `lineage.Tags.Queries`. A disabled sweep that performs tag reads would still pass. The required contract is zero reads and zero writes. Add an assertion that tag query count is unchanged as well (and retain the existing event-query/create assertions).

2. Required closeout writeback is stale/incomplete. The PR body still labels the matrix and test plan as 16 scenarios and says “all sixteen”, although the repaired matrix now has 20. It records zero new product defects and the earlier three fake-vs-production review catches, but does not record the fourth review finding that the original G6 matrix was vacuous for the real hosted sweep path (the reason these four production-chain tests were added). Update the durable writeback to map/report all 20, retain “zero new product defects”, and explicitly distinguish the G6 vacuous-test coverage finding from the three earlier product defects caught during review.

Verified resolved: registered enabled DI reaches real `CosmosTagSweepService -> CosmosTagRepairRunner -> CosmosDbTagRepairService`; query work-count proves checkpoint resume vs rescan and completion clearing; sweep execution isolates the two lineages; no sleeps or polling; focused 20-test matrix passes on net9.0 and net10.0.

PR #1060, opened 2026-07-12
Add the operator-authorized legacy tag-row migration

### LINKED ISSUE #1059: SEK-G4b Cosmos DCB: legacy tag-row dedup/migration (destructive, operator-authorized)
# SEK-G4b Cosmos DCB: legacy tag-row dedup/migration (destructive, operator-authorized)

## Goal

Add a **destructive** dedup/migration mode to the Cosmos tag-repair surface that
canonicalizes or removes legacy non-deterministic-id tag rows — gated behind
explicit operator authorization, with ETag-guarded deletes, a deterministic
survivor policy, a backup/recovery path, and an audit log. Expose it **both** as a
provider administrative service API **and** as a standalone operator CLI tool that
wraps the same service. This is the only packet permitted to delete or rewrite tag
rows.

## Why This Slice Exists Now

Split out of SEK-G4 per the orchestrator design review (L3, 2026-07-12). SEK-G4
recognizes and reports legacy rows (pre-SEK-G2 `Guid.NewGuid()` ids) but is
strictly non-destructive. Actually collapsing duplicate legacy rows or migrating
them to the deterministic-id format is destructive: it materially increases blast
radius (deletes) and review surface, so it must be a separate, explicitly
authorized operation — never a default, never reachable from the automatic sweep.

## Current Observed State

- SEK-G4 (merged) ships a strictly non-destructive Cosmos tag-repair administrative
  service: it backfills truly-missing tag rows and classifies/reports
  legacy-present / duplicate / corrupt / overflow — but never deletes,
  canonicalizes, or rewrites any tag row.
- Legacy tag rows (pre-SEK-G2, non-deterministic `Guid.NewGuid()` ids) keep working
  for reads because SEK-G4's comparator recognizes them by semantic key; migration
  is therefore optional, not required for correctness.
- No destructive dedup/migration surface exists, and there is no operator CLI for
  the repair/maintenance path today.
- During the G1–G6 closeout, no real legacy-row population was observed, so this
  capability is prospective (built ahead of a confirmed cleanup need).

- SEK-G4's non-destructive repair + classification exists: legacy recognition via
  semantic-key match (ordinal serviceId/tag, canonical Guid eventId, partition-key
  confined), event-derived-field validation, and the `LegacySemanticEquals`
  classifier. Categories: missing / legacy-present / duplicate / corrupt / overflow.
- SEK-G2 deterministic derivation + strict `ContentEquals`.
- The SEK-G5 automatic sweep is statically forbidden from invoking this surface.

## Target Repo / Path / Part

Target paths: `dcb/src/Sekiban.Dcb.CosmosDb dcb/tests`
Target part: destructive dedup/migration mode on the Cosmos tag-repair surface.

## In Scope

- A distinct destructive operation (separate method/mode from SEK-G4 repair),
  requiring an explicit confirm flag/mode — never the default.
- Deterministic survivor / canonical-row policy (prefer the SEK-G2
  deterministic-id row; else a documented deterministic tiebreak); reproducible.
- ETag-guarded (optimistic-concurrency) deletes; a lost race is retried or
  reported, never forced.
- Backup/export of rows to be deleted (or an equivalent documented recovery path)
  produced from the dry-run artifact before any deletion.
- Bounded per-`(event,tag)` cap with overflow reported; checkpoint-resumable;
  defined, safe concurrent-writer behavior.
- Audit log of every canonicalization/delete (survivor, removed, keys, counts).
- Lineage safety identical to SEK-G4 (one serviceId + container lineage per run).
- A standalone operator CLI tool (e.g. `dotnet tool` / console utility) that wraps
  the administrative service and exposes the same safety flow from the terminal
  (confirm flag, default dry-run, backup/export, per-lineage config, range,
  checkpoint/resume, parallelism, cancellation). The CLI is a thin front-end over
  the one service implementation — no destructive logic duplicated.
- Tests and EN + JA docs.

## Out Of Scope

- The non-destructive repair/classification itself (SEK-G4).
- Automatic scheduling — this operation is operator-only and never automatic.
- Reverse-direction repair (tag rows whose event is missing).

The child PR must deliver an operator-authorized destructive dedup/migration mode,
separate from the SEK-G4 non-destructive repair and unreachable from the SEK-G5
sweep, that reduces multiple legacy/deterministic rows for one `(event,tag)` to a
deterministic canonical survivor using ETag-guarded deletes, produces a dry-run
artifact and backup/recovery path before mutating, records an audit log, is
bounded and resumable, defines safe concurrent-writer behavior, and is covered by
tests (including a structural test that the automatic sweep cannot reach it) and
EN + JA docs.

## Acceptance Criteria

- Destructive dedup/migration is a distinct, explicitly-authorized surface; never default; never reachable from the SEK-G5 sweep (structural test).
- Delivered both as a service API and a standalone operator CLI tool sharing one implementation (CLI is a thin front-end; no duplicated destructive logic; CLI defaults to dry-run and requires an explicit confirm flag to mutate).
- Explicit operator authorization (confirm flag/mode) required; a dry-run artifact is produced and is the basis for the real run.
- Deterministic survivor/canonical-row policy (prefer SEK-G2 deterministic-id row; else documented tiebreak); reproducible.
- ETag-guarded deletes; concurrent writer not silently clobbered (retry or report).
- Backup/export or documented recovery path exists before deletion.
- Bounded per-key cap with overflow reported; checkpoint-resumable; safe concurrent-writer behavior.
- Audit log of every canonicalization/delete.
- Lineage safety identical to SEK-G4.
- Tests: deterministic reduction of a multi-legacy (event,tag); ETag-guarded delete respects concurrent writer; dry-run non-mutating; partial re-run idempotent/resumable; sweep-cannot-invoke structural test. Docs updated (EN + JA).

## Verification

## Related Links

- Split from SEK-G4 per orchestrator design review L3 (host design doc §6)

## Knowledge Maintenance

- Intent placement: host intent node 01 storage-consistency.
- ADR candidate: yes — "destructive legacy-row dedup/migration is a separate operator-authorized packet, never automatic" (design §6 / review L3).
- Diagram candidate: none.
- Docs update: yes — destructive-mode usage, authorization, survivor policy, backup/recovery, operator-only never-automatic warning (EN + JA).
- Closeout writeback expected: yes — report real legacy-row populations found and survivor-policy behavior.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

A **destructive** dedup/migration mode that reduces the tag rows of an `(event, tag)` to the single canonical row — as a service API and a standalone operator CLI over one shared implementation.

Closes #1059

**It deletes documents. It is the only thing in Sekiban that does**, so it is built as a thing that would rather refuse. Note the framing: legacy rows *work* — the repair service recognizes them by semantic key, tag reads find them, and **nothing about correctness depends on migrating them**. This is housekeeping, never a fix.

**No destructive run was executed against any real Cosmos data or account.** Everything below is exercised against the in-memory harness.

## The sweep cannot reach it — structurally

The migration is a separate service on a **separate store seam** (`ICosmosTagMigrationStore`) — the only seam in the provider that can express a delete at all. The repair service and the automatic sweep are wired to `ICosmosTagRepairStore`, which has none. So the sweep cannot delete a row by configuration, by DI, or by accident.

Two reflection tests pin this: the sweep's and repair's constructors, fields, properties, and method signatures mention **none** of the migration types; and the migration seam is `internal`, so **no assembly outside the provider — the CLI included — can implement or call a delete**. The only public route to removing a tag row is `ApplyAsync`, behind every gate below.

`AddSekibanDcbCosmosDb` + `AddSekibanDcbCosmosDbTagRepair` + `AddSekibanDcbCosmosDbTagSweep` together still yield **no** ability to delete (tested).

## Two passes, so you cannot delete what you have not seen

`PlanAsync` mutates nothing and produces an artifact naming exactly which rows would die. `ApplyAsync` takes **that artifact and nothing else**.

| Gate | Behavior — all *before* a document is touched |
|---|---|
| No plan | Refused. `ApplyAsync` has no overload that deletes without one. |
| No `Confirm` | `CosmosTagMigrationNotAuthorizedException`. No permissive default. |
| No backup writer | `CosmosTagMigrationNotAuthorizedException`. Cosmos has no undo, so the export *is* the recovery path. |
| Plan altered since produced | Fingerprint (lineage + every action) no longer matches → `CosmosTagMigrationPlanRejectedException`. An artifact that was not reviewed authorizes nothing. |
| Plan from another lineage | Refused. An instance binds one `(serviceId, events, tags)` at construction. |
| Backup write throws | **Nothing is deleted.** The backup is written first, before the first delete, on purpose. |

## Survivor policy — deterministic, reproducible

The canonical (deterministic-id) row always survives. If none exists, it is **created from the event** — so the survivor's content is what the write path would have written, and no legacy quirk outlives the migration — and created **before** any removal, so the key is never left unindexed, not even for an instant. Legacy rows are never promoted, only removed, so **there is no tiebreak to get wrong**. Planning the same unchanged world twice produces a byte-identical artifact (tested).

## Concurrency — never forced

Deletes are **ETag-guarded** against the version the plan pinned. If a row was written to since the operator reviewed it, Cosmos refuses — and so does the migration. It re-reads, and retries at the new version **only if the row still says exactly what it said when the plan called it a removable duplicate**. If any content changed, the row is left alone and the race is reported (`LostRace`). There is no code path that forces a delete past a moved row.

A row that disagrees with its event is not a duplicate — it is corruption, and what to do about that is not this service's call: `Skipped`, never deleted. Same for over-cap keys (`Overflow`). Every key produces an audit entry, **including the ones it declined to touch**.

## Two real bugs found while building this

Both in my own code, both caught by the tests before push:

1. **The survivor was never created.** When only legacy rows existed, `ApplyAsync` read for the canonical row, found nothing, and declared the key stale — it never called create. It would have deleted nothing, but the "create before remove" invariant was fiction.
2. **The ETag guard could be forced.** On a mismatch, the retry re-pinned the new ETag if the row's *identity* still matched, ignoring that its *content* had changed — so a concurrent writer's edit would have been deleted anyway. That is precisely what "never forced" forbids. It now retries only when the row is still a semantically identical legacy duplicate.

## Do the gates actually hold?

**Mutation-checked, each in isolation, with the suite green before and after:**

| Mutation | Result |
|---|---|
| Remove the `Confirm` gate | tests fail |
| Remove the backup requirement | tests fail |
| Remove the plan fingerprint check | tests fail |
| Delete without `IfMatchEtag` | tests fail |

## CLI

`tools/SekibanDcbTagMigration` — a thin front-end over the same service, with **zero destructive logic of its own** (it could not have any: the delete seam is internal). Defaults to dry-run; `apply` refuses an unconfirmed run **before it opens a connection** (verified by running it with no reachable account — no backup file is even created).

```
plan   Read-only. Writes the artifact.
apply  Destructive. Refuses without --confirm and --backup.
```

## Test plan

- [x] 21 new tests: plan mutates nothing; re-planning is byte-identical; refuses without confirm / without backup / without a plan / with an altered plan / with another lineage's plan; a failing backup deletes nothing; reduction to the canonical row with the backup written first (asserted: zero deletes at backup time); an existing canonical row survives and is not rewritten; re-running finds nothing to do; a row changed since the plan → `Stale`, left alone; a row that moves between the backup and the delete → `LostRace`, left alone; corrupt → `Skipped`; over-cap → `Overflow`; bounded plans resume; lineage isolation (the other lineage: zero deletes); **sweep has no route to the destructive surface**; the repair seam still cannot express a delete; `AddSekibanDcbCosmosDb` alone does not register it.
- [x] `Sekiban.Dcb.WithResult.Tests` 621 passed / 0 failed — no regressions
- [x] CLI builds and refuses offline; `DcbOrleans.ApiService` builds
- [x] Docs updated (EN + JA): usage, authorization, survivor policy, backup/recovery, ETag semantics, and the operator-only never-automatic warning
- [x] No analyzer warnings; `git diff --check` clean

## Closeout writeback

**No real legacy-row population was observed** during G1–G6, and none exists in this repo's data — consistent with the packet's note that this capability is prospective. Survivor-policy behavior is therefore reported from the harness only: the canonical row wins in every case, is created from the event when absent, and legacy rows are never promoted.


tomohisa:
Blocking destructive-safety review findings (no real Cosmos execution was performed):

1. Canonical survivor is not revalidated before deletes. A plan with `SurvivorExists=true` causes `ApplyActionAsync` to skip the create/read-back branch entirely. If the deterministic survivor is deleted or changed after planning, apply can still delete all planned legacy rows, leaving the key unindexed or retaining a corrupt canonical row. This violates the packet requirement that the canonical row exists from event-derived content before any removal and that concurrent writers are safe. Pin the planned survivor version/content in the artifact and, immediately before deletion, prove an exact event-derived canonical row exists; recreate it safely if absent, or report stale/lost-race without deleting if it differs. Add mutations/tests for survivor deleted and survivor content changed between plan/apply.

2. Plan fingerprint does not cover `CosmosTagMigrationAction.Tag`. `ComputeFingerprint()` hashes partition key, event id, survivor fields, and removals, but omits `action.Tag`. An edited artifact can change Tag without invalidating the fingerprint; apply then derives/creates a row for the altered tag in the original partition and deletes the reviewed legacy rows. This violates tampered-artifact rejection and lineage/key binding. Fingerprint every mutation-relevant action field (including Tag) with unambiguous encoding, and add a tampered-Tag rejection test.

3. ETag mismatch retry does not compare the full planned removable row. `DeleteWithEtagGuardAsync` re-pins a new ETag whenever `CosmosLegacyTagRowMatcher.Classify(...)=LegacyPresent`. That semantic comparator intentionally ignores legacy metadata such as createdAt and other non-event-derived content, so a row changed after operator review may be deleted under a new ETag even though its full contents differ from the reviewed/exported version. The required boundary says mismatch re-read compares full removable content and never unsafe re-pins. Carry a complete immutable snapshot/hash in the plan and retry only on exact full-content equality (excluding only Cosmos-owned version metadata as explicitly documented); otherwise audit LostRace and leave it alone. Add a mutation test where ignored metadata/content changes under the ETag.

The focused migration suite is green (21/21 on net9.0 and net10.0), but these unsafe mutations are not covered. Existing positive evidence: destructive API is separate/confirm-gated, backup writer is invoked before deletes, CLI refuses no-confirm before building the Cosmos service, and no real account/data was accessed.

tomohisa:
Re-review: fingerprint and full-snapshot ETag findings are resolved, but the canonical-survivor invariant is still not race-safe.

`ApplyActionAsync` calls `EnsureCanonicalSurvivorAsync` once per action, then loops over `RowsToRemove` and deletes them. The survivor is not re-read immediately before each delete. A concurrent actor can delete or corrupt the canonical row after that one proof (including between two legacy-row deletes), and the migration will continue deleting the remaining legacy rows. The new tests mutate/delete the survivor before `ApplyAsync`, so they do not exercise this post-proof race and remain green if the survivor changes during the delete loop.

This still violates the required safety boundary: canonical survivor strictly re-read/`ContentEquals` immediately before every delete; changed survivor causes zero further deletes/no unindexed interval. Move the proof into the per-row destructive boundary (or use an equivalent atomic partition operation that conditionally proves the canonical row and deletes the victim together). Add deterministic mutation hooks/tests for (a) survivor removed after proof but before the first delete and (b) survivor changed between two deletes; assert no victim is deleted after the proof fails and audit the stale-survivor/lost-race outcome completely.

Verified resolved: collision-safe fingerprint includes Tag and full snapshots; ETag re-pin requires exact planned snapshot excluding only Cosmos-owned `_etag`/`_ts`; backup includes planned plus live snapshots; focused tests pass 26/26 on net9.0 and net10.0. No real Cosmos account was accessed.

tomohisa:
Atomic-batch rereview: the prior survivor/delete race is resolved in production, but two blocking gaps remain.

1. A canonical-id row that is already corrupt at plan time is not classified as corruption. `CosmosTagSurvivorPolicy.Plan` selects any row whose id equals the deterministic id as the survivor, while `AllRowsAreSafeToRemove` validates only legacy victims. The plan then records that corrupt survivor ETag, and `ExecuteReduceAsync` performs an IfMatch `ReplaceItem` with the derived row before deleting victims. This silently rewrites/canonicalizes corruption instead of skipping it, contrary to “corrupt/overflow skipped” and the operator-reviewed plan boundary. At plan time require strict `CosmosTagIdentity.ContentEquals(derived, deterministicSurvivor)`; if false, emit `Corrupt`, no action/backup/mutation. Add a deterministic test with a corrupt canonical-id row plus legacy duplicates and assert plan skipped, zero batch operations/deletes/replaces.

2. The in-memory transactional harness does not enforce the batch partition. `CreateTransactionalBatch(partitionKey)` discards `partitionKey`; `InMemoryTransactionalBatch` stores only the container; and delete lookup derives a partition by scanning globally for the first matching id. It therefore accepts cross-partition operations that real Cosmos rejects and can target the wrong same-id row. This makes same-partition/lineage mutation tests permissive. Store the batch partition key, require every create/replace document pk to equal it, resolve deletes strictly as `(batchPk,id)`, and reject any mismatch atomically. Add mutation tests showing a wrong action partition/document pk and same id in two partitions cannot commit.

Verified resolved: production uses one transactional batch per action (survivor Create or IfMatch Replace first, all victims IfMatch Delete); failed batches audit empty RemovedIds; standalone create/delete removed from migration seam; status mapping separates survivor/victim rejection; 99 victims succeeds and 100 is skipped at plan time; prior fingerprint/full-snapshot/CLI/G5 gates remain. Focused tests pass 30/30 on net9.0 and net10.0. No real Cosmos account was accessed.

Issue #1045, opened 2026-07-10
NullReferenceException in CosmosDbEventStore.WriteEventsAsync (Sekiban.Dcb.CosmosDb 9.2.0)


joemphilips:
## Description

`CosmosDbEventStore.WriteEventsAsync` returns a null/unset `ResultBox` instead of a proper error result when the Cosmos write fails. Calling `.UnwrapBox()` on the result throws `System.NullReferenceException: Object reference not set to an instance of an object.`

## Stack trace

```
WriteEventsAsync failed: Object reference not set to an instance of an object.
Stack trace:    at Sekiban.Dcb.CosmosDb.CosmosDbEventStore.WriteEventsAsync(IEnumerable`1 events)
warn: Failed to dispatch Event Sourcing mintd-mirror refresh
      System.NullReferenceException: Object reference not set to an instance of an object.
         at ResultBoxes.UnwrapExtensions.UnwrapBox[TValue](ResultBox`1 result)
         at EventSourcingMintdMirrorWriter.RefreshAsync(...)  // calls executor.ExecuteAsync(command, handler, ct).UnwrapBox()
```

## Environment

- Sekiban.Dcb.CosmosDb: 9.2.0
- ResultBoxes: 0.3.31
- Microsoft.Azure.Cosmos: 3.61.0
- .NET: 9.0 (on Azure App Service Linux, DOTNETCORE|9.0)
- Cosmos DB: Azure Cosmos DB (serverless), partition key `/aggregateId`

## Reproduction

1. Configure Sekiban with Cosmos DB event store
2. Call `executor.ExecuteAsync(command, handler, cancellationToken)` which internally calls `CosmosDbEventStore.WriteEventsAsync`
3. The `ResultBox` returned is null, causing `.UnwrapBox()` to throw NRE

## Expected behavior

`WriteEventsAsync` should return a proper `ResultBox` with the error information (e.g., `ResultBox.FromException(ex)`) instead of null, so that `.UnwrapBox()` can throw a meaningful exception.

## Questions

1. Is this a known issue in 9.2.0?
2. Is there a configuration step that could cause `WriteEventsAsync` to silently return null?
3. Does Sekiban 10.x fix this? (We tried upgrading but the API breaking changes were too large for a quick migration.)

## Additional context

[...]

PR #1066, opened 2026-07-14
Guard every WithoutResult ResultBox boundary

### LINKED ISSUE #1065: SEK-G9 dcb: guarded unwrap + unified exception policy for WithoutResult boundaries
# SEK-G9 dcb: guarded unwrap + unified exception policy for WithoutResult boundaries

## Goal

Replace every bare `UnwrapBox` boundary in the WithoutResult packages with a single
contextual guarded unwrapper (null-box / failed-box / success-null detection, inner
exception preserved, operation context attached) and unify the exception-mapping
policy, so WithoutResult failures are always descriptive instead of bare
`NullReferenceException`s.

## Why This Slice Exists Now

Issue #1045's reporter lost significant diagnosis time because a null/failed
`ResultBox` surfaced through `ResultBoxes.UnwrapBox` as a bare NRE with no message —
and explicitly flagged "a guarded unwrap with context would be a real robustness
win". Investigation (2026-07-14) confirmed the structural cause: WithoutResult is a
facade that computes ResultBox internally and strips it at 20 context-erasing
boundaries, while Core's exception→ResultBox conversion is spread across ~114 catch
sites with non-uniform null guards. This slice is the DECIDED P0 of the
WithoutResult-core inversion discussion; the inversion itself (P1–P3) is an
undecided future aspiration and is out of scope.

## Current Observed State

- 20 bare `UnwrapBox` call sites: 18 in `Sekiban.Dcb.WithoutResult`, 2 in
  `Sekiban.Dcb.Orleans.WithoutResult`.
- A null ResultBox, a failed ResultBox, or a success-with-null-value can surface as
  a bare NRE without operation/tag/projector context.
- Success-null guards exist in SOME Core query paths ("HandleQuery returned null
  result", "ResultBox value is null") but protection is not uniform.
- Cancellation and inner-exception preservation behavior differs by boundary
  because catch/wrap/unwrap policies grew organically.

- Core/Core.Model are ResultBox-native by design; removing ResultBox from Core
  contracts is explicitly NOT this slice (P1–P3, undecided).
- WithResult consumers rely on current ResultBox semantics — their observable
  behavior must not change.
- dcb-v10.3.0 is released; this slice targets main after it.

## Target Repo / Path / Part

Target paths: `dcb/src/Sekiban.Dcb.WithoutResult dcb/src/Sekiban.Dcb.WithoutResult.Model dcb/src/Sekiban.Dcb.Orleans.WithoutResult dcb/src/Sekiban.Dcb.Core dcb/tests`
Target part: guarded unwrap helper + the 20 boundary call sites + exception-mapping policy + tests + troubleshooting docs.

## In Scope

- One shared guarded-unwrap helper (prefer INTERNAL) taking an operation context
  (operation kind, command/query/projector type name, tag when available):
  1. null ResultBox → descriptive exception naming the operation (never bare NRE);
  2. failed ResultBox → per the precedence rule below;
  3. success-with-null-value → descriptive contextual exception.
  The helper must NOT delegate to `UnwrapBox` — it inspects
  `IsSuccess`/`GetException`/`GetValue` directly.
- **Exception precedence rule**: cancellation is the deliberate exception to
  contextual wrapping — a failure carrying `OperationCanceledException` rethrows
  the ORIGINAL instance via `ExceptionDispatchInfo` (stack and CancellationToken
  preserved, never wrapped). ALL other failures throw ONE defined contextual
  boundary exception type with the original as `InnerException`.
- Replace all 20 bare UnwrapBox sites with the helper. **Mechanical gate**: rg over
  both WithoutResult projects returns no production `UnwrapBox` invocation at all.
- **Completeness guard**: the PR lists the 20 pre-change sites mapped each to a
  boundary-category test. If the discovered count differs from 20, STOP and reply
  blocked to the orchestrator — do not silently widen scope.
- Tests: per-boundary-category null-box; success-null via a controlled test double
  (deliberately constructed null-success shape); cancellation
  reference/type/token/stack preservation; wrapped-case inner-exception
  type/stack/context assertions; issue-1045-style regression.
- Docs: short note in `docs/dcb_llm(_ja)/13_common_issues.md` describing the new
  contextual exceptions. The note must NOT imply issue #1045 was a Sekiban
  functional defect (it was stale state; the fix here is diagnostic quality).

## Out Of Scope

- **Core-wide catch-site normalization (G9-B1)** — explicitly out of scope. A
  `Sekiban.Dcb.Core` edit is permitted ONLY when strictly necessary for a named
  guarded boundary or its test, and the PR must enumerate and justify every Core
  file changed. Prefer implementing everything in
  `Sekiban.Dcb.WithoutResult` / `Sekiban.Dcb.Orleans.WithoutResult`.
- Native ports / partial or full inversion (P1–P3 — undecided future aspiration).
- Any removal/change of ResultBox in Core public contracts.
- WithResult behavior changes.
- Orleans surrogate wire-format changes.

The child PR must add one shared contextual guarded-unwrap helper, replace all 20
bare UnwrapBox boundaries in the WithoutResult packages with it, define and apply a
consistent exception-mapping policy (cancellation passthrough, inner-exception
preservation), keep public API changes additive-only with WithResult behavior
unchanged, cover the behavior with boundary tests including an issue-1045-style
regression, and add the troubleshooting docs note (EN + JA).

## Acceptance Criteria

- Shared guarded-unwrap helper (prefer internal) with operation context; it inspects IsSuccess/GetException/GetValue directly and never delegates to UnwrapBox.
- All 20 bare UnwrapBox sites replaced; mechanical gate: rg over the two WithoutResult projects returns no production UnwrapBox invocation at all.
- PR evidence lists the 20 pre-change sites mapped to boundary-category tests; a count other than 20 is a stop-and-report (blocked reply), not silent expansion.
- Exception precedence applied: cancellation rethrows the ORIGINAL OperationCanceledException via ExceptionDispatchInfo (reference/type/token/stack preserved, never wrapped); all other failures throw ONE contextual boundary exception with the original as InnerException.
- Core-wide normalization out of scope; any Core file change is enumerated and justified in the PR as strictly necessary for a named boundary/test.
- Public API changes additive-only (boundary exception type may be public so consumers can catch it); WithResult observable behavior unchanged (existing suites green).
- Tests: null-box per boundary category, success-null via controlled test double, cancellation preservation assertions, wrapped-case exception-shape assertions, issue-1045 regression.
- docs/dcb_llm(_ja)/13_common_issues.md note added (EN + JA) without implying issue #1045 was a functional defect.

## Verification

All new and existing dcb tests green; `git diff --check` clean; no public signature removals (API-diff or reviewer check).

## Related Links

- https://github.com/J-Tech-Japan/Sekiban/issues/1045 (diagnostic-defect origin)
- Host design record: intents/sekiban/design/withoutresult-core-inversion.md

## Knowledge Maintenance

- Intent placement: host design doc withoutresult-core-inversion.md (P0 decided).
- ADR candidate: none (decision recorded host-side).
- Diagram candidate: none.
- Docs update: 13_common_issues.md (EN + JA) — in scope.
- Closeout writeback expected: yes — report any additional unguarded sites/policy inconsistencies found; evidence feeds the P1–P3 inversion discussion.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1065

## What was actually wrong

WithoutResult is a facade over the ResultBox-native core: internally every operation returns a `ResultBox<T>`, and at the edge the box is opened. It was opened with `UnwrapBox()` — and `UnwrapBox()`'s behaviour depends on the shape of `T`. I probed ResultBoxes 0.4.0 directly rather than assume:

| box | before |
|---|---|
| failed, `T` reference type | rethrows the carried failure — correct |
| failed, `T` **value type** (`bool`, `int`) | returns `default`, **throws nothing** |
| null box | bare `NullReferenceException`, no message, no operation name |

Row 3 is what issue #1045 hit. Per the reporter's own closing comment, #1045 had **two causes and only one was ours**: their staging carried stale projection state (no projector-versioning discipline) so reads returned null — not a Sekiban defect — and *our* `UnwrapBox()` turned that null into a bare `NullReferenceException` with no message, so an environment problem looked exactly like a library bug. The diagnostics half is ours to fix.

**Row 2 is a correctness defect of our own, which nobody had reported, found while fixing row 3.** `ICommandContext.TagExistsAsync` returns `bool`, so an event store that could not be reached came back as `false` — indistinguishable from *"no, that tag does not exist"*. A command handler guarding on it goes on to create the entity it was just told, wrongly, was absent. `ISekibanExecutor.QueryAsync<TResult> where TResult : notnull` admits structs, so a failed count query answered `0`. The Orleans `ToTypedResult<TResult>()` sites had the same hole.

## The policy

All **20** production `UnwrapBox` call sites — 6 `CommandContextAdapter`, 9 `GeneralSekibanExecutor`, 3 `Generic*TagMultiProjector`, 2 `OrleansDcbExecutor` — now go through `GuardedUnwrap`, which never calls `UnwrapBox`:

1. **A failure is rethrown as itself** — same instance, same type, same stack, via `ExceptionDispatchInfo` — whether it was carried *inside the box* or thrown by *the Task meant to produce the box*, and whether `T` is a reference type or a value type. Cancellation keeps its original token and needs no special case: it is just a carried failure. What changed is that a value-typed boundary now throws it instead of swallowing it.
2. **The boundary is recorded on `Exception.Data`** (`Sekiban.Boundary.Operation`, `Sekiban.Boundary.Target`) — context without a type change. Read and both writes happen under one gate, so an exception crossing concurrent boundaries can never end up with an operation from one and a target from another. First writer wins, and that is the innermost boundary.
3. **When there is no failure to rethrow** (null box, null `Task`), the caller gets a `SekibanBoundaryException` naming the operation, instead of an unexplained `NullReferenceException`.

### Deviation from the original packet, stated plainly

The packet asked for *all* non-cancellation failures to be wrapped in `SekibanBoundaryException` with the original as `InnerException`. **I did not do that, because I implemented it and the repository told me it was wrong**: three existing tests fail — `ExecuteCommandAsync_HandlerThrows_Throws`, and both `Command_WithInvalidData_Should_ThrowException` / `Command_WithDuplicateStudent_Should_ThrowException`, which expect their domain exception. Those tests *are* the WithoutResult contract: this package exists so that a `SekibanValidationException` arrives as a `SekibanValidationException`. Wrapping every failure would break every `catch` block our users have written, in a slice whose purpose is diagnostics. Wrapping is therefore reserved for the case that has nothing to wrap — which is also the only case that was ever silent.

## Evidence

- `dcb/tests/Sekiban.Dcb.WithoutResult.Tests` — **66 passed** (30 pre-existing, 36 new), 0 failed. CI green on net9 and net10; SonarCloud green.
- The new tests **assert the old behaviour where it differed**, so they cannot pass against it: `Assert.False(ResultBox<bool>.Error(ex).UnwrapBox())`, `Assert.Equal(0, general.ToTypedResult<int>().UnwrapBox())` and `Assert.Throws<NullReferenceException>(() => ((ResultBox<string>)null!).UnwrapBox())` sit next to the guarded assertions.
- **Mutation-tested — six mutations, all killed**: delete the null-box branch (5 fail), drop the `Data` annotation (7 fail), wrap carried failures instead of rethrowing (17 fail), revert `TagExistsAsync` to `UnwrapBox` (3 fail), drop the faulted-Task catch (3 fail), remove the annotation lock (2 fail, 3 runs out of 3).
- Not regressed: `Sekiban.Dcb.WithResult.Tests` **634 passed** (no behaviour or signature change), `Sekiban.Dcb.Orleans.Tests` 80 passed / 1 skipped, `MaterializedView.Tests` 21, `ColdEvents.Tests` 85. `Sekiban.slnx` builds with 0 errors.
- Mechanical gate: `rg '\.UnwrapBox\(\)'` over `Sekiban.Dcb.WithoutResult` and `Sekiban.Dcb.Orleans.WithoutResult` returns **nothing**.
- Diff hygiene: the three projector files are **4 additions / 1 deletion each**. Two of them are CRLF on `main`, and `core.autocrlf=input` had flattened them to LF on my first commit (271/268 of pure line-ending churn); their original endings are restored.

## Site → test map

| sites | boundary | covered by |
|---|---|---|
| `CommandContextAdapter` ×6 | `ICommandContext.*` | `CommandContextBoundaryTests` — failed value-typed box, cancellation + token, null-box, per-operation `Operation`/`Target` |
| `GeneralSekibanExecutor` ×9 | `ISekibanExecutor.*` | `ExecutorBoundaryTests` (validation failure stays a `SekibanValidationException` **and** is annotated), `StructQueryBoundaryTests` (a failed `QueryAsync<int>` throws instead of answering `0`) |
| `OrleansDcbExecutor` ×2 | Orleans typed-result | `StructQueryBoundaryTests` — the production expression `general.ToTypedResult<int>()` on a cast failure: `UnwrapBox()` returned `0`, the guard throws the carried `InvalidCastException` |
| all async sites | Task-carried failures | `TaskCarriedFailureTests` — faulted Task (identity, stack), faulted-with-cancellation (instance + token), runtime-cancelled Task (type + token + annotation) |
| all sites | concurrent annotation | `ConcurrentAnnotationTests` — see below |
| `Generic*TagMultiProjector` ×3 | `Project` | uniformity only — **`ProjectInternal` has no failure path today** (every branch returns `FromValue`), so I make no coverage claim for a failure that cannot currently occur |

### On the concurrency test

My first attempt was worthless and I am not going to hide it: 24 threads on a barrier, and it **passed against a lock-free guard, three runs out of three**. Racing and hoping is not a test. The replacement does not hope — the exception's own `Data` dictionary **parks the first writer mid-write** until a second thread has got past the `Contains` check, which it can only do if the read-then-write is not atomic, and then counts the writes. Atomic: exactly one write. Not atomic: two, deterministically, 3 runs out of 3.

## Honest notes

- `ResultBoxes` 0.4.0 cannot represent "success carrying no value": a null value reads as `IsSuccess == false`, carrying a synthesised `ResultValueNullException`. The guard keeps a branch for it anyway; `BoxWithNoValue_IsAFailure_NotASuccess_InResultBoxes` pins that invariant, so if a future ResultBoxes changes it, the test fails and the branch stops being dead code.
- `GuardedUnwrap` is `internal`; `InternalsVisibleTo` is granted to `Sekiban.Dcb.Orleans.WithoutResult` (same boundary, other assembly) and the test project. The only new **public** type is `SekibanBoundaryException` — additive.
- No `Sekiban.Dcb.Core` edits (SEK-G9 constraint). No WithResult behaviour or signature change.
- Docs: `docs/dcb_llm{,_ja}/13_common_issues.md` state the #1045 history as it actually happened — stale state on the reporter's side, a diagnostics defect on ours — and keep the swallowed value-typed failure as a separately discovered correctness defect.


tomohisa:
SEK-G9 strict semantic review — request update

The green CI and focused net9/net10 tests are necessary but not sufficient. The amended host contract at 2c61d2f still has blocking gaps:

1. **Faulted/cancelled Task failures are not boundary-annotated.** `GuardedUnwrap.UnwrapAsync` checks only a null Task, then directly awaits it. If the Task itself faults or cancels (rather than successfully returning a failed ResultBox), the original exception propagates without `Sekiban.Boundary.Operation` / `Target`. The amended contract explicitly covers failures carried by **box or task**: preserve the original exception instance/type/token/stack via EDI and annotate the boundary. Add faulted-Task and cancelled-Task identity/token/Data regressions.

2. **`Exception.Data` annotation is not concurrency-safe or atomic.** `Contains(OperationDataKey)` followed by separate Operation and Target writes races when the same exception is annotated across concurrent boundaries/tasks. Two callers can both pass `Contains`, overwrite one another, or leave an operation/target pair from different contexts; `Exception.Data` itself is not promised thread-safe. Implement serialization/atomic ownership for each exception so the innermost/first boundary wins as one context, and add a deterministic concurrent same-exception regression that verifies one coherent pair and no annotation failure replaces the original exception.

3. **Required struct-query regression is absent.** The tests cover failed `ResultBox<bool>` at the helper/TagExists boundary, but no failed struct `ISekibanExecutor.QueryAsync<TResult>` / Orleans typed-result boundary exists. Add a real struct query regression proving failure throws the original instance instead of returning `default(T)`.

4. **Minimize line-ending churn.** `GenericIntTagMultiProjector.cs` and `GenericStringTagMultiProjector.cs` each show 271 additions / 268 deletions solely because CRLF was normalized to LF. `--ignore-space-at-eol` reduces each semantic change to 4 additions / 1 deletion. Preserve the original line endings so review/audit sees only the guarded unwrap delta.

5. **Correct EN/JA issue-1045 history.** The new docs describe the bare NRE diagnostics symptom and separately describe swallowed value-type failures, but omit that #1045 itself involved stale state plus a diagnostics defect. State that historical cause explicitly; keep the value-type swallowed failure identified as a separately discovered correctness defect.

Verified non-blocking facts: all 20 production UnwrapBox sites are replaced; no production UnwrapBox remains in the two target packages; no Core or WithResult edit; carried box failures use EDI and preserve identity/type/token/stack; null box/null Task/success-null policy is guarded; friend assemblies are limited to Orleans.WithoutResult and tests; current CI is green. PR-body G7/G8 status is ignored for closeout as directed.

PR #1069, opened 2026-07-14
DCB: durability/runtime descriptors, fail-closed production guard, startup banner

### LINKED ISSUE #1068: SEK-G10 dcb: durability/runtime descriptors + fail-closed production guard + startup banner
# SEK-G10 dcb: durability/runtime descriptors + fail-closed production guard + startup banner

## Goal

Make volatile storage and test-only execution machine-detectable and impossible to
run in Production silently: additive capability descriptors on stores and
executors, an explicit opt-in `AddSekibanDcbProductionGuard` that fails closed, and
a structured startup banner. Zero default behavior change.

## Why This Slice Exists Now

A downstream production system registered `InMemoryDcbExecutor` as
`ISekibanExecutor`; its one-arg constructor silently created a private volatile
store, and nothing at startup could detect that events were not reaching Cosmos.
Names are human-checkable only — prevention needs runtime-resolved capability
metadata plus a fail-closed guard. Operator policy (2026-07-14): real environments
always use a distributed-runtime executor; the InMemory executor is test-only.

## Current Observed State

- `IEventStore` and projection state stores expose no durability metadata; a
  volatile store is indistinguishable from a durable one at composition time.
- `InMemoryDcbExecutor` (WithResult and WithoutResult variants) uses in-process
  actors (`InMemoryObjectAccessor`) with no Orleans involvement; its one-arg
  constructor silently creates a private volatile store.
- No startup validation or banner reports executor mode / store durability.
- Existing docs place in-memory types under testing, but nothing enforces it.

- Operator taxonomy (design doc): production/real envs = distributed-runtime
  executor + durable store; local dev = Orleans localhost + optionally volatile
  store; unit tests = InMemory executor + volatile store.
- The g1 "DirectDcbExecutor" alternative was rejected; do not add new supported
  non-distributed production executors in this slice.
- Renaming/moving the testing stack is the NEXT slice (SEK-G11); this slice only
  adds descriptors/guard/banner and the one-arg [Obsolete].

## Target Repo / Path / Part

Target part: capability descriptors, production guard hosted validator, startup banner, one-arg ctor obsolete, tests, docs.

## In Scope

- **Two runtime-resolved capability axes** (additive interfaces/registration
  metadata, not names/attributes):
  - store durability: `Durable | Volatile | Unknown` + ProviderName — on event
    stores AND projection/multi-projection state stores;
  - executor runtime: `DistributedRuntime | TestingInProcess | Unknown` +
    RuntimeName.
  Built-ins self-declare (Cosmos/Postgres/Sqlite/DynamoDB = Durable; InMemory =
  Volatile; Orleans executor = DistributedRuntime; InMemoryDcbExecutor =
  TestingInProcess). Decorators/factories propagate the underlying descriptor; the
  guard evaluates the actually resolved instances.
- **`AddSekibanDcbProductionGuard(options)`** — explicit opt-in hosted startup
  validator (readiness gate) evaluated after registrations. Production fails
  closed on Volatile/Unknown store or TestingInProcess/Unknown executor. Never
  auto-injected into existing registrations.
- **Override separation**: `AllowVolatileStorageInProduction` permits volatile
  STORAGE only — it never authorizes a testing executor. No default testing
  executor override; any break-glass is separately named, audited, default-false,
  and absent from templates.
- **Structured startup banner**: environment, resolved executor type/runtime,
  store providers + durability (event + projection), override names used, loud
  data-loss warning when volatile; no connection strings.
- **`[Obsolete]` on the one-argument `InMemoryDcbExecutor` constructor** with a
  migration message (behavior unchanged).
- Tests (matrix below) and EN/JA docs.

## Out Of Scope

- Renaming/moving InMemory types or Testing packages (SEK-G11).
- Any behavior change without opt-in; auto-registration of the guard.
- Analyzer / deployment canary (deferred P2).
- Non-distributed production executor support (rejected by operator policy).

The child PR must add the two additive runtime-resolved capability axes with
built-in self-declarations and decorator propagation, the explicit opt-in
production guard hosted validator that fails closed on volatile/unknown storage
and testing/unknown executors with separated overrides (storage override can never
authorize a testing executor), the structured startup banner, and the one-arg
constructor obsoletion — with zero default behavior change, the full test matrix,
and EN + JA docs.

## Acceptance Criteria

- Two capability axes exist, runtime-resolved, propagated through decorators/factories; built-ins self-declare correctly.
- AddSekibanDcbProductionGuard is explicit opt-in; in Production it fails closed on Volatile/Unknown stores and TestingInProcess/Unknown executors; not auto-injected.
- AllowVolatileStorageInProduction authorizes storage only; an explicit test proves it cannot authorize a testing executor; no default testing-executor override exists.
- Banner logs environment, executor runtime, store durability (event + projection), overrides used, data-loss warning; no secrets.
- One-arg InMemoryDcbExecutor constructor is [Obsolete] (behavior unchanged).
- Zero default behavior change; existing suites green without modification.
- Test matrix: Distributed+Durable pass / Distributed+Volatile prod fail (storage override → pass) / TestingExecutor+Durable prod fail / TestingExecutor+Volatile prod fail / Unknown prod fail / Development unchanged / decorator propagation / override-separation proof.
- Docs (EN + JA) updated.

## Verification

## Related Links

- Host design record: intents/sekiban/design/inmemory-misuse-prevention.md
- Incident distillation (downstream): SekibanAsAServiceHost ADR-0007 / SKS-G808/G812/G814

## Knowledge Maintenance

- Intent placement: host design doc inmemory-misuse-prevention.md (P0).
- ADR candidate: none (policy recorded host-side).
- Diagram candidate: none.
- Docs update: 11_storage_providers.md EN/JA — in scope.
- Closeout writeback expected: yes — unclassified descriptor candidates and nontrivial decorator paths, feeding SEK-G11.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1068

## The thing that happened

A downstream production system registered `InMemoryDcbExecutor` as its `ISekibanExecutor`. The one-argument constructor quietly created a private in-memory event store. Every command succeeded. No event ever reached the Cosmos account that was configured — sitting there, correct, and empty. Nothing at startup could detect it, **because nothing could ask**.

Names are not evidence. This PR adds the thing that can be asked.

## Two axes, resolved from the live instance

| Axis | Values | Answered by |
|---|---|---|
| Storage durability | `Durable` / `Volatile` / `Unknown` + provider name | event stores **and** projection state stores, plus their per-service factories |
| Executor runtime | `DistributedRuntime` / `TestingInProcess` / `Unknown` + runtime name | executors — by asking the actor accessor they actually run on |

Built-ins self-declare: Postgres / Cosmos / DynamoDB → `Durable`, InMemory → `Volatile`, Orleans → `DistributedRuntime`, `InMemoryDcbExecutor` → `TestingInProcess`.

Two cases are the argument for resolving this at runtime instead of inferring it from a type name, an attribute, or which `Add…` method was called:

- **Sqlite answers `Durable` for a file and `Volatile` for `:memory:`.** Same class, same name, opposite guarantee. Only the live instance knows which one it got.
- **`HybridEventStore` reports where the data actually lands.** Wrapping a volatile store is *volatile*. A decorator cannot launder durability by being wrapped around it.

`Unknown` is not a neutral value: it means the component declined to say, and the guard treats it as unsafe. Silence is not a promise of durability.

## The guard

`AddSekibanDcbProductionGuard()` — explicit opt-in, never auto-registered by a provider or a template. At startup, after every registration has had its say and before the host serves anything, it resolves **what the container actually built**, asks each thing what it is, logs the banner, and in Production refuses to start on a non-distributed executor or a non-durable store. Outside Production it validates nothing.

It resolves rather than reading the `IServiceCollection` — so a later `AddSingleton<IEventStore>` that swaps a durable store for a volatile one is still caught. There is a test for exactly that.

### The one override, and how narrow it is

`AllowVolatileStorageInProduction` is narrow in two directions:

- **Storage only.** Set it, register `InMemoryDcbExecutor` in Production, and the host still refuses to start.
- **`Volatile` only — never `Unknown`.** Setting it means *"I looked at a store that said it was volatile, and I meant it."* A store that says nothing has not given you anything to mean, so `Unknown` stays fail-closed with the override on. (This was **wrong in the first push** — Volatile and Unknown were one condition and the override skipped both. The review caught it; it now has its own enforcement branch and its own regression.)

There is **no** override that permits a testing executor in Production. Not off by default, not behind a flag — it does not exist, and `NoOverrideExistsForATestingExecutor` fails if a property on the options type is so much as *named* like one. A volatile store in Production can be a decision (a cache-shaped service, a throwaway environment). A test executor in Production is not a decision, it is an accident.

Any override that *is* on is named, by name, in the banner at `Warning`.

### Resolved in a scope

Sekiban's own Cosmos registrations are **scoped**. A hosted service holds the *root* provider, and asking the root provider for a scoped service throws under `ValidateScopes` — so opting into the guard broke a supported composition outright, in Development too. A guard that breaks the host it is protecting is worse than no guard. It now resolves inside a scope that lives exactly as long as the check, and a `ValidateScopes=true` test fails if it ever resolves from the root again.

### The banner

Environment, executor type + runtime, both store providers + durability, overrides in use, and a loud data-loss warning when storage is volatile. Never a connection string: a banner that leaks a secret into every log sink would be a worse bug than the one it exists to prevent. `AddSekibanDcbStartupBanner()` gives the report without the enforcement, for local development.

## Test matrix (all green)

| case | result |
|---|---|
| Production + Distributed + Durable | starts |
| Production + Distributed + Volatile | **refuses** |
| …with `AllowVolatileStorageInProduction` | starts |
| Production + Testing executor + Durable | **refuses** |
| Production + Testing executor + Volatile | **refuses** |
| **Production + Testing executor + storage override** | **still refuses** — override-separation proof |
| Production + Unknown executor / Unknown store | **refuses** |
| **Production + Unknown store + storage override** | **still refuses** — the override authorises `Volatile`, not silence |
| Production + Volatile store + storage override | starts — the pair that shows what the override *does* |
| scoped executor/stores, `ValidateScopes=true` | guard reports (does not throw); a scoped volatile store is still caught |
| Production + nothing registered | **refuses** |
| Development + Testing + Volatile | starts, unchanged |
| banner-only in Production | never fails the host |
| operator-named production env (`prod-eu`) | guarded |
| store swapped *after* the guard is registered | **refuses** (resolution, not registration) |
| decorator propagation (volatile / durable / silent inner) | reports the inner |
| the incident, reproduced end to end with the real `InMemoryDcbExecutor` | **refuses** |

Every guard case builds a real host and really starts it: the guard's whole job is to stop a host from starting, and asserting on a method in isolation would prove something weaker than the claim.

**Mutation-tested — six mutations, all killed**: let the storage override excuse a testing executor (1 fail), treat `Unknown` as acceptable (2), let the decorator claim durability on its own authority (2), make Sqlite always claim `Durable` (2), **fold `Unknown` back under the volatile override** (1), **resolve from the root provider again** (2).

## Review round 2

Three blockers, all real, all fixed:

1. **`Unknown` storage was authorised by the volatile-only override.** Fixed as described above, with the regression the review asked for.
2. **The validator resolved from the root provider**, which throws under `ValidateScopes` for the scoped Cosmos registrations. Now resolved in a scope, with a `ValidateScopes=true` test.
3. **`TheSameCompositionInDevelopment_StillStarts` asserted nothing** (Sonar S2699, BLOCKER). It was also too weak a claim to leave a test on — a guard that let the host start but broke the executor on the way past would have satisfied it. It now starts the host **and executes a command through the resolved executor**. SonarCloud is green.

## The obsolete constructor, and what obsoleting it caught

`new InMemoryDcbExecutor(domainTypes)` is `[Obsolete]`. Behaviour unchanged — only its silence is deprecated.

All **33** call sites in this repo now pass the store explicitly, so the solution builds with **zero** obsolete warnings and the migration we ask users to make is one we made first. That migration then caught itself: my first pass substituted a bare `new InMemoryEventStore()`, which uses `ReflectionEventTypes` instead of the domain's registered event types, and **21 existing tests failed**. A silent store substitution changing behaviour under a green-looking build is precisely what this slice exists to prevent, so it is worth saying that it happened here, to me, and that the tests are what said so.

Template content (`templates/**`) is outside the issue's stated target paths (`dcb/src dcb/tests`), and I edited it anyway, deliberately: leaving it would have made every newly generated project emit obsolete warnings on first build. The edit is mechanical — the same explicit-store migration — and I would rather declare it than ship that.

## Evidence

- `Sekiban.slnx`: **0 errors, 0 obsolete warnings** on a clean (`--no-incremental`) build.
- WithResult 654 passed (634 pre-existing + 20 new), WithoutResult 70 (66 + 4), ColdEvents 88 (85 + 3), Orleans 80 / 1 skipped, MaterializedView 21, MaterializedView.MultiProvider 9, BlobStorage.AzureStorage 4. All pre-existing suites green **unmodified** in behaviour.
- `git diff --check` clean. No public signature removals; everything added is additive.

## Closeout notes (feeding SEK-G11)

- **Unclassified descriptor candidates**: `IBlobStorageSnapshotAccessor` (`InMemoryBlobStorageSnapshotAccessor` volatile; Azure Blob and S3 durable) is *not* covered — the issue scoped the axes to event stores and projection state stores. A snapshot accessor that is volatile in Production is a smaller but real hole, and it is the obvious next descriptor.
- **Nontrivial decorator paths**: `HybridEventStore` is the only `IEventStore` decorator in-tree and it propagates. `EventStoreCacheSync` composes two stores but is not itself an `IEventStore`, so it is invisible to the guard — worth a look if it ever becomes one.
- The multi-tenant factories (`IEventStoreFactory`, `IMultiProjectionStateStoreFactory`) self-declare, but the guard describes the *factory*, not the per-service store it builds. That is correct for the built-ins (every store a factory builds shares its backend) and would not be for a factory that mixes backends per service.

### DIFF
--- a/dcb/src/Sekiban.Dcb.Core/Capabilities/ExecutorRuntimeDescriptor.cs
+++ b/dcb/src/Sekiban.Dcb.Core/Capabilities/ExecutorRuntimeDescriptor.cs
@@ -1,0 +1,53 @@
+namespace Sekiban.Dcb.Capabilities;
+
+/// <summary>
+///     What kind of runtime an executor actually runs on.
+/// </summary>
+public enum ExecutorRuntimeKind
+{
+    /// <summary>
+    ///     The executor did not say. Treated as unsafe: in Production the guard fails closed on it, because an executor
+    ///     that cannot state its runtime cannot be shown to be a real one.
+    /// </summary>
+    Unknown = 0,
+
+    /// <summary>
+    ///     In-process actors, no cluster, no distribution — a unit-test runtime. Nothing here survives the process or
+    ///     coordinates with another one.
+    /// </summary>
+    TestingInProcess = 1,
+
+    /// <summary>
+    ///     A real distributed runtime (Orleans today): grains placed across a cluster, state coordinated between hosts.
+    /// </summary>
+    DistributedRuntime = 2
+}
+
+/// <summary>
+///     What an executor says about itself when something asks at runtime.
+/// </summary>
+/// <param name="Runtime">The kind of runtime this executor actually uses.</param>
+/// <param name="RuntimeName">
+///     A human name for the banner and the failure message: "Orleans", "InMemory (in-process actors)". Never a
+///     connection string.
+/// </param>
+public sealed record ExecutorRuntimeDescriptor(ExecutorRuntimeKind Runtime, string RuntimeName)
+{
+    /// <summary>For an executor whose runtime cannot be determined — e.g. a generic executor over an unknown accessor.</summary>
+    public static ExecutorRuntimeDescriptor Unknown(string runtimeName) =>
+        new(ExecutorRuntimeKind.Unknown, runtimeName);
+
+    /// <summary>e.g. <c>Orleans (DistributedRuntime)</c>.</summary>
+    public override string ToString() => $"{RuntimeName} ({Runtime})";
+}
+
+/// <summary>
+///     Implemented by executors — and by the actor accessors they run on — that can state their own runtime.
+///     Executors that are generic over an accessor answer by asking the accessor, so the answer describes the runtime
+///     that will actually execute the command, not the class that was registered.
+/// </summary>
+public interface IExecutorRuntimeDescriptorProvider
+{
+    /// <summary>Describes this executor as it is actually composed, at the moment it is asked.</summary>
+    ExecutorRuntimeDescriptor DescribeRuntime();
+}
--- a/dcb/src/Sekiban.Dcb.Core/Capabilities/SekibanDcbCapabilityExtensions.cs
+++ b/dcb/src/Sekiban.Dcb.Core/Capabilities/SekibanDcbCapabilityExtensions.cs
@@ -1,0 +1,57 @@
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+namespace Sekiban.Dcb.Capabilities;
+
+/// <summary>
+///     Registration for the startup banner and the production guard.
+///     Both are opt-in and neither changes how anything executes. The guard is never auto-registered by a provider or
+///     a template: a library that decides on its own when your host may not start is a library that will one day be
+///     wrong about it.
+/// </summary>
+public static class SekibanDcbCapabilityExtensions
+{
+    /// <summary>
+    ///     Registers the guard. The package that knows what <c>ISekibanExecutor</c> is passes in the lookup — see
+    ///     <c>AddSekibanDcbProductionGuard</c> in Sekiban.Dcb.WithResult / Sekiban.Dcb.WithoutResult, which is what
+    ///     applications call.
+    /// </summary>
+    public static IServiceCollection AddSekibanDcbProductionGuardCore(
+        this IServiceCollection services,
+        Func<IServiceProvider, object?> resolveExecutor,
+        Action<SekibanDcbProductionGuardOptions>? configure = null) =>
+        services.AddSekibanDcbStartupValidator(resolveExecutor, configure, true);
+
+    /// <summary>
+    ///     Registers the banner alone: it reports what was resolved and warns loudly about volatile storage, and it
+    ///     never fails the host. For local development, where a volatile store is the point.
+    /// </summary>
+    public static IServiceCollection AddSekibanDcbStartupBannerCore(
+        this IServiceCollection services,
+        Func<IServiceProvider, object?> resolveExecutor,
+        Action<SekibanDcbProductionGuardOptions>? configure = null) =>
+        services.AddSekibanDcbStartupValidator(resolveExecutor, configure, false);
+
+    private static IServiceCollection AddSekibanDcbStartupValidator(
+        this IServiceCollection services,
+        Func<IServiceProvider, object?> resolveExecutor,
+        Action<SekibanDcbProductionGuardOptions>? configure,
+        bool enforce)
+    {
+        ArgumentNullException.ThrowIfNull(services);
+        ArgumentNullException.ThrowIfNull(resolveExecutor);
+
+        var options = new SekibanDcbProductionGuardOptions();
+        configure?.Invoke(options);
+
+        services.AddSingleton<IHostedService>(sp => new SekibanDcbStartupValidator(
+            sp,
+            sp.GetRequiredService<IHostEnvironment>(),
+            options,
+            resolveExecutor,
+            enforce,
+            sp.GetRequiredService<ILogger<SekibanDcbStartupValidator>>()));
+
+        return services;
+    }
+}
--- a/dcb/src/Sekiban.Dcb.Core/Capabilities/SekibanDcbCapabilityReport.cs
+++ b/dcb/src/Sekiban.Dcb.Core/Capabilities/SekibanDcbCapabilityReport.cs
@@ -1,0 +1,30 @@
+namespace Sekiban.Dcb.Capabilities;
+
+/// <summary>
+///     What was actually resolved out of the container, and what it said about itself.
+///     Assembled from live instances, not from registrations: a decorator, a factory or a lambda that quietly swaps the
+///     real store for something else is visible here, because here is where we ask the thing that got built.
+/// </summary>
+/// <param name="EnvironmentName">The host environment, e.g. "Production".</param>
+/// <param name="IsProductionEnvironment">Whether the guard considers that environment a production one.</param>
+/// <param name="ExecutorTypeName">The concrete executor type that was resolved, or null if none was registered.</param>
+/// <param name="Executor">What the executor said about its runtime.</param>
+/// <param name="EventStore">What the event store said about its durability.</param>
+/// <param name="ProjectionStore">What the projection state store said about its durability.</param>
+/// <param name="UsedOverrideNames">Overrides the operator turned on, by name.</param>
+public sealed record SekibanDcbCapabilityReport(
+    string EnvironmentName,
+    bool IsProductionEnvironment,
+    string? ExecutorTypeName,
+    ExecutorRuntimeDescriptor Executor,
+    StorageDurabilityDescriptor EventStore,
+    StorageDurabilityDescriptor ProjectionStore,
+    IReadOnlyList<string> UsedOverrideNames)
+{
+    /// <summary>True when any resolved store does not durably keep what it is given.</summary>
+    public bool HasVolatileOrUnknownStorage =>
+        EventStore.Durability != StorageDurability.Durable || ProjectionStore.Durability != StorageDurability.Durable;
+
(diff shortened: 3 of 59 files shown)


tomohisa:
SEK-G10 semantic review — request update

I checked the packet/review-context against head `1c65de3f4dda243fec3d76bd699b46aa566c1f55`. The runtime descriptors, TestingInProcess separation, Hybrid/Sqlite propagation, explicit opt-in registration, constructor compatibility, templates migration, and EN/JA docs are broadly aligned, but three current-PR blockers remain:

1. **Unknown storage is incorrectly authorized by the volatile-only override** (implementation finding; packet AC “Unknown fails closed in Production” + “AllowVolatileStorageInProduction permits ONLY volatile storage”). In `SekibanDcbStartupValidator.Enforce`, line 74 combines `Volatile` and `Unknown` under `HasVolatileOrUnknownStorage`, then skips the entire fault when `AllowVolatileStorageInProduction=true`. Thus DistributedRuntime + Unknown store + the storage override starts successfully. Unknown is not an operator-confirmed volatile store and must remain fail-closed. Separate Unknown from Volatile in enforcement and add the missing matrix regression: Unknown + storage override still refuses.

2. **The hosted validator resolves scoped executor/stores from the root provider** (implementation finding; packet AC “guard evaluates the ACTUALLY RESOLVED executor and stores” and “Development behavior unchanged”). `SekibanDcbStartupValidator` retains the root provider supplied to the singleton `IHostedService`; `SekibanDcbCapabilityResolver.Resolve` calls `GetService<IEventStore>()` / projection / executor directly on it. Built-in Cosmos Full/JWT registrations are scoped. With scope validation enabled, opting into the guard/banner throws “Cannot resolve scoped service from root provider” before producing a report—even outside Production—so those supported registration styles cannot use the feature and Development is not unchanged. Resolve the capability set inside a created scope (and test a scoped executor/event/projection composition with `ValidateScopes=true`) while retaining startup-only lifetime.

3. **Current required quality gate is red** (implementation finding). SonarCloud Code Analysis reports BLOCKER S2699 at `InMemoryExecutorCapabilityTests.cs:64`: `TheSameCompositionInDevelopment_StillStarts` has no assertion. Make the Development proof explicit/non-vacuous so the companion Sonar gate passes. The two S1133 informational notices are expected consequences of the packet-required additive `[Obsolete]` compatibility constructor and are not separate semantic blockers.

Evidence: local `dotnet test dcb/tests/Sekiban.Dcb.WithResult.Tests/Sekiban.Dcb.WithResult.Tests.csproj -- --filter-class '*Capabilities*'` completed with 650/650 passing on both net9 and net10; passing tests are necessary but not sufficient because the missing Unknown+override and scoped-resolution cases are not covered. GitHub dcbTestsNet9/net10 are green, but SonarCloud Code Analysis is currently FAILURE.

Out-of-scope boundary assessment: the six template edits are limited to warning-prevention migration from the newly obsolete one-arg constructor, preserve the registered domain event types, and do not perform the G11 rename/package move; I consider them justified and behavior-preserving. No merge was performed.

PR #1071, opened 2026-07-14
DCB: Testing packages, and a localhost Orleans composition that actually works

### LINKED ISSUE #1070: SEK-G11 dcb: split testing stack into Testing packages + localhost Orleans composition guide
# SEK-G11 dcb: split testing stack into Testing packages + localhost Orleans composition guide

## Goal

Give the test-only in-memory stack a structural (compile/reference) boundary —
`Sekiban.Dcb.Core.Testing` / `WithResult.Testing` / `WithoutResult.Testing` with
Obsolete forwarding aliases — and ship the consumable localhost Orleans
composition (web / worker / CLI lifecycles) that makes "distributed-runtime
executor everywhere, even locally" a practical instruction rather than a slogan.

## Why This Slice Exists Now

Operator policy (2026-07-14): real environments always use a distributed-runtime
executor; local dev uses Orleans localhost with in-memory state; the InMemory
executor is test-only — including its two-argument form. Naming alone cannot
enforce this; a package boundary can (a runtime project that does not reference
`*.Testing` cannot even compile the volatile executor). The orchestrator review
set one hard precondition: do NOT declare non-Orleans production unsupported until
a documented, tested localhost Orleans composition exists for web, worker, and
short-lived CLI/batch lifecycles.

## Current Observed State

- All in-memory types live inside the production packages (`Sekiban.Dcb.Core`,
  `WithResult`, `WithoutResult`) under `Sekiban.Dcb.InMemory`.
- `InMemoryDcbExecutor` exists per facade; SEK-G10 marks it `TestingInProcess` via
  descriptor and obsoletes the one-arg constructor, but the types remain in
  production packages.
- No documented single-silo localhost Orleans composition exists for CLI/batch;
  templates run Orleans locally via Aspire (web-centric).
- Templates already confine InMemory usage to `*.Unit` projects (preserve this).

- SEK-G10 descriptors/guard/banner are merged: descriptors are authoritative;
  names are secondary.
- The DirectDcbExecutor alternative is rejected; do not add new supported
  non-distributed production executors.
- Production-internal in-memory caches (e.g. `GeneralTagStateActor`'s
  `InMemoryTagStatePersistent`) are implementation details, not authority stores —
  they stay where they are.

## Target Repo / Path / Part

Target paths: `dcb/src dcb/tests templates/Sekiban.Dcb.Templates docs/dcb_llm docs/dcb_llm_ja`
Target part: Testing packages + forwarders, localhost Orleans composition docs/tests, migration inventory, template/docs updates, release-workflow packaging.

## In Scope

- **Testing packages (dependency-safe split)**: `Sekiban.Dcb.Core.Testing`
  (volatile Core stores/helpers for tests), `Sekiban.Dcb.WithResult.Testing` and
  `Sekiban.Dcb.WithoutResult.Testing` (facade-specific testing executors with
  explicit names such as `InMemoryDcbExecutorForTesting`, declaring
  `TestingInProcess`). WithResult/WithoutResult executors are not forced into one
  ambiguous package.
- **Compatibility**: old types/locations kept as Obsolete forwarding surfaces;
  binary + source compatibility preserved; physical removal only at next major.
- **Localhost Orleans composition (hard precondition)**: documented and tested for
  web host, worker host, and short-lived CLI/batch — single silo, deterministic
  StartAsync/StopAsync, no external clustering dependency, explicit
  durable-or-volatile store selection, graceful drain, sample commands; CLI/batch
  silo startup cost documented as a known trade-off.
- **Migration inventory**: enumerate downstream two-arg
  `InMemoryDcbExecutor(domainTypes, durableStore)` usages and give each concrete
  migration guidance to the localhost composition.
- **Docs (EN + JA)**: the taxonomy (real env / local dev / unit test), Testing
  packages, composition guide, migration guidance.
- **Templates**: `*.Unit` projects reference Testing packages; runtime projects do
  not; README local-dev guidance points at the localhost composition.
- **Release workflow**: pack/publish the new Testing packages in packagesDcb.yml.
- **Source-scan test**: no runtime project references `*.Testing`.

## Out Of Scope

- Removing/moving old types physically (next major).
- Moving production-internal in-memory caches.
- Analyzer / deployment canary (P2, deferred).
- New non-distributed production executors (rejected).

The child PR must add the three Testing packages with dependency-safe contents and
explicit testing-named executors declaring TestingInProcess, keep old surfaces as
Obsolete forwarders with binary/source compatibility proven by tests, ship the
documented and tested localhost Orleans composition for web/worker/CLI lifecycles
before the docs declare non-distributed production unsupported, produce the two-arg
usage migration inventory with concrete guidance, update templates/docs (EN + JA),
extend the release workflow to publish the Testing packages, and prove via a
source-scan test that no runtime project references *.Testing.

## Acceptance Criteria

- Three Testing packages exist with dependency-safe split; testing executors have explicit names and declare TestingInProcess; WithResult/WithoutResult not merged into one ambiguous package.
- Old types/locations remain as Obsolete forwarders; binary/source compatibility proven; production-internal caches untouched.
- Localhost Orleans composition documented AND tested for web/worker/CLI: deterministic lifecycle, no external clustering, explicit store selection, graceful drain, sample commands, CLI cold-start trade-off documented.
- Downstream two-arg usage inventory with per-usage migration guidance; no silent reclassification without a replacement.
- Docs EN/JA state the taxonomy; templates confine Testing references to *.Unit; release workflow publishes the new packages.
- Source-scan test: no runtime project references *.Testing.
- Zero behavior change for existing consumers (compat tests for old constructors/package graphs).

## Verification

All new and existing dcb tests green; template smoke tests still pass; `git diff --check` clean.

## Related Links

- Host design record: intents/sekiban/design/inmemory-misuse-prevention.md
- Depends on SEK-G10 (descriptors/guard)

## Knowledge Maintenance

- Intent placement: host design doc inmemory-misuse-prevention.md (P1).
- Diagram candidate: none.
- Docs update: 11_storage_providers.md + 13_common_issues.md EN/JA — in scope.
- Closeout writeback expected: yes — migration inventory results, CLI composition ergonomics, analyzer readiness.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1070

## What this is for

Naming did not prevent the incident. `InMemoryDcbExecutor` reads like a test type, and a production system registered it as its `ISekibanExecutor` anyway — every command succeeded, and no event ever reached the database it had configured. [SEK-G10](https://github.com/J-Tech-Japan/Sekiban/pull/1069) made that **detectable at startup**. This gives the volatile stack a home a runtime project has no reason to enter, and puts the compiler across *that* door.

**Scope of the claim, since an earlier draft of this PR overstated it.** The compile-time boundary applies to the **new** `Sekiban.Dcb.Testing` entry points: a project without the Testing packages cannot reach `InMemoryDcbExecutorForTesting` or the stores beside it, so it cannot pick the testing executor up by accident. It does **not** make the legacy path uncompilable — `Sekiban.Dcb.InMemory.InMemoryDcbExecutor` and the old stores stay public in the runtime packages, still compile, still run, and the compatibility tests below prove it on purpose. For that path the protections are the `[Obsolete]` pointer and the opt-in production guard — whose two answers are not the same answer: **an in-process executor is rejected in Production, always**, with no override, whichever name it was obtained under; a **volatile store is rejected by default but can be authorised** by the named, storage-only `AllowVolatileStorageInProduction`; and a store that will not say what it is (`Unknown`) stays fail-closed either way. The old door closes at the next major; until then the guard is what stands in it, not the compiler — a wall for a testing executor, and a decision the operator has to make by name for volatile storage.

## Three Testing packages

| package | contents |
|---|---|
| `Sekiban.Dcb.Core.Testing` | the volatile stores and helpers — namespace `Sekiban.Dcb.Testing` |
| `Sekiban.Dcb.WithResult.Testing` | `InMemoryDcbExecutorForTesting` (ResultBox facade) |
| `Sekiban.Dcb.WithoutResult.Testing` | `InMemoryDcbExecutorForTesting` (exception-based facade) |

The two facades stay apart deliberately: a single ambiguous `Sekiban.Dcb.Testing` holding both executors would be a package whose meaning depends on which one you happened to import.

A project that does not reference these packages cannot reach the new testing entry points at all. That is a real boundary, and it is what stops a runtime project adopting the testing executor — see the scope note above for what it does *not* do.

## Compatibility: deprecating something is a promise that it still works

The old types are `[Obsolete]`, not removed — and not re-implemented either: the implementation stayed where it was, and the new `Sekiban.Dcb.Testing` types derive from it, so calling them "forwarders" would have been wrong and I have stopped. `LegacyInMemorySurfaceCompatibilityTests` keeps the promise honestly: the old executor still executes, the old one-argument constructor still works (private volatile store and all), and the new types are assignable to the old ones, so a helper or fixture typed against the old name takes the new one unchanged. Physical removal waits for the next major.

**`InMemoryTagStatePersistent` is not swept up.** Despite the name and the namespace, it is the tag-state actor's real in-process cache, not a test double. It stays in `Sekiban.Dcb.Core`, it is not obsolete, and a test fails if a future cleanup ever confuses the two.

## The hard precondition: a localhost Orleans composition that actually works

Telling people "the in-memory executor is test-only" is not an instruction until there is something to use instead. So:

```csharp
builder.UseOrleans(silo => silo.UseSekibanDcbLocalhost());
```

One silo, localhost clustering, memory grain storage and streams, nothing to install — and a **real** Orleans runtime, which reports itself as `DistributedRuntime` because it is one. It deliberately does **not** choose your event store: a local silo over Postgres and a local silo over an in-memory store are both legitimate, and neither should be implicit.

**All three lifecycles are tested, not merely documented** — real hosts, really started, really used, really stopped (`LocalhostCompositionTests`):

| lifecycle | what the test actually does |
|---|---|
| **web** | `WebApplication` on a real Kestrel port; an HTTP POST executes a command on the silo and the event lands in the store |
| **worker** | a `BackgroundService` executes through the silo, then drains on shutdown |
| **CLI / batch** | `Host.CreateApplicationBuilder` → `StartAsync` → one command → `StopAsync`, and the events are in the store the caller chose |

Plus one that pins the claim the taxonomy rests on: the localhost executor reports `DistributedRuntime`, not `TestingInProcess`. If that ever flipped, "use Orleans locally" would be advice with nothing behind it.

**The CLI cold start (~1s per invocation) is documented as the trade-off it is**, with the workaround, rather than left for someone to discover and feel misled by.

## The boundary, enforced against this repository

`TestingPackageBoundaryTests` reads every runtime `.csproj` under `dcb/src`, `dcb/internalUsages` and `templates/` and fails if any of them references a `*.Testing` package — including the template *content*, which no solution in this repo compiles. It also fails if the release workflow stops publishing the Testing packages: a package nobody can install is a boundary nobody can cross to. And it fails if it ever silently scans zero projects.

**Mutation-tested, both killed**: add a Testing `PackageReference` to a template ApiService (scan fails), delete the pack steps from `packagesDcb.yml` (workflow test fails).

## We migrated this repository first

58 files — tests and template content — now go through the Testing packages, and the solution builds with **zero** obsolescence warnings on a clean `--no-incremental` build. The migration we ask users to make is the one we made.

One thing the migration caught about itself, worth stating: subclassing changes `GetType()`, and a white-box test that reflected over `_actorAccessor.GetType().GetField("_actors", NonPublic)` stopped finding a private field declared on the base. It now walks the hierarchy. Nothing was broken by it, but it is the kind of thing a forwarding split can quietly do, so it is named here rather than buried.

## Evidence

- `Sekiban.slnx`: **0 errors, 0 obsolete warnings**, clean `--no-incremental` build.
- WithResult **657** (654 + 3 boundary), WithoutResult **76** (70 + 6 compatibility), Orleans **84** / 1 skipped (80 + 4 localhost lifecycles), ColdEvents 88, MaterializedView 21, MV.MultiProvider 9, BlobStorage.Azure 4. Every pre-existing suite green with **no behavioural change**.
- `git diff --check` clean. No public signature removals; the Testing packages and `UseSekibanDcbLocalhost` are additive.
- Release workflow packs all three new packages.

## Docs

New `docs/dcb_llm{,_ja}/22_localhost_orleans.md`: the taxonomy (real env / local dev / unit test), the three lifecycles with code, the Testing packages, and a migration table whose dangerous row is the one that looks fine — **an in-process executor over a *durable* store**, where the events survive but commands still run on in-process actors with no cluster coordination, so two hosts never see each other's tag reservations. `11_storage_providers`, `12_unit_testing`, `13_common_issues` and the template README point at it.

## Closeout notes

- **The migration inventory is honest about its limit.** We can enumerate the two-argument usages in *this* repository, and we did (they are all tests and template content, now migrated). We cannot enumerate yours — the doc gives the grep and explains what each shape means, and says plainly that nobody else can run it for you.
- **What still blocks the physical move** (next major): `Sekiban.Dcb.Core` itself type-checks `is InMemoryEventStore` in `GeneralTagStateActor` and `EventStoreExtensions` to fetch `IEventTypes`. Those two sites are `#pragma`-guarded and commented; removing that coupling is the prerequisite for deleting the old types rather than deprecating them.
- **Analyzer readiness**: with descriptors (G10) and the package boundary (G11) in place, an analyzer that flags a `*.Testing` reference from a non-test project would now have something unambiguous to key on. The source-scan test is the cheap version of it and covers the template content an analyzer would not.


tomohisa:
Blocking semantic/documentation finding:

The package split does not make the legacy volatile executor or stores uncompilable. `Sekiban.Dcb.InMemory.InMemoryDcbExecutor` and the old store types remain public in the runtime packages for compatibility, and `ObsoleteForwarderCompatibilityTests.TheOldExecutorStillExecutes` / `TheOldOneArgumentConstructorStillWorks_PrivateVolatileStoreAndAll` explicitly prove that a project with no `*.Testing` reference still compiles and runs that volatile path. The new Testing types derive from those retained implementations; the old types are not forwarding implementations.

Therefore the PR body statements “This makes it uncompilable” and “A project that does not reference these packages cannot compile against the volatile executor”, plus the equivalent unqualified claims in EN/JA `11_storage_providers.md`, overstate the protection and contradict the compatibility section immediately below. Please scope the claim consistently across PR metadata, EN/JA docs, and package READMEs: the compiler boundary applies to the new testing-named API and prevents runtime projects from adopting those new entry points; legacy `Sekiban.Dcb.InMemory` entry points remain usable until the next major, with `[Obsolete]` guidance and the opt-in G10 production guard as the available mitigations. Also avoid calling the retained old implementations “forwarders” unless the implementation is actually inverted.

All other reviewed areas are currently non-blocking: three dependency-safe packages and packed dependency graph; old identity/constructor behavior; untouched `InMemoryTagStatePersistent`; real localhost Orleans web/worker/CLI composition; non-vacuous runtime/template source scan; template Unit-only references; release workflow; focused tests (4 localhost + 3 boundary + 6 compatibility) and all four current GitHub checks pass.

tomohisa:
One narrow accuracy blocker remains in the repaired wording.

The scope/compiler claim and retained-base/new-derived topology are now described correctly. However, the replacement mitigation text says the opt-in G10 guard “refuses to start a Production host that resolved an in-process executor or a volatile store however it was obtained” (PR body and EN/JA localhost docs; the Core.Testing README has the same absolute statement). That is true for `TestingInProcess`, but not for volatile storage alone: G10 deliberately permits `StorageDurability.Volatile` when the operator sets `AllowVolatileStorageInProduction=true`; only `Unknown` remains fail-closed and there is no override for a testing executor.

Please preserve the current boundary/compatibility correction while making the guard statement match the settled G10 contract everywhere it appears: an in-process testing executor is always rejected in Production; volatile stores are rejected by default but can be authorized by the named storage-only override. This is documentation/XML-doc/PR-metadata accuracy only; no behavior or test change is requested.

Current-head policy note: GitHub reports dcbTestsNet9 and dcbTestsNet10 SUCCESS. No Sonar runs are attached to this prose/XML-doc/test-class-rename head, but the main branch has no required status-check protection configured, so Sonar absence is not an independent blocker under the current repository policy.

tomohisa:
Final SEK-G11 semantic rereview: **no blocking findings remain** at `8fa777a6ced654327534b72cbc5acbaf940bd3bb`.

The compiler boundary is now accurately scoped to the new `Sekiban.Dcb.Testing` APIs; legacy runtime-package surfaces are explicitly retained, compilable, and runnable until the next major. The settled G10 contract is consistent across PR metadata, EN/JA docs, package READMEs, and executor XML docs: `TestingInProcess` always fails Production with no override; `Volatile` storage fails by default and only the storage-specific `AllowVolatileStorageInProduction` can authorize it; `Unknown` remains fail-closed; and that storage override never authorizes `TestingInProcess`.

Prior full-contract verification remains valid for the three dependency-safe packages, retained API identities, untouched production cache, real localhost Orleans web/worker/CLI lifecycle, non-vacuous source scan, Unit-only template references, package dependency graph, and release workflow. `git diff --check` is clean; `Closes #1070` remains correct. Current-head `dcbTestsNet9` and `dcbTestsNet10` are SUCCESS. No Sonar run was emitted for this prose-only head; main has no required status-check protection, so this is not a routing blocker.

GitHub formal approval was attempted and rejected solely because this account is the PR author (`Review Can not approve your own pull request`). No merge was performed; routing is released for the standing design-owned admin-merge fallback.

Issue #1075, opened 2026-07-16
Sekiban.Dcb: InMemoryObjectAccessor.CreateActor catch-all swallows projection fold exceptions — failing catch-up presents as an empty successful query


tomohisa:
## Summary

`InMemoryObjectAccessor.CreateActor` wraps actor creation/catch-up in a broad catch-all. When a multi-projection fold throws while processing an event (in our case `GeneralMultiProjectionActor.AddSerializableEventsAsync` per-event fold throwing `InvalidOperationException: Value cannot be null. (Parameter 'key')`), the exception is **swallowed** and the list query returns an **empty successful result** instead of an error.

## Reproduction (verified on 10.2.2 and 10.4.0, .NET 10, real PostgreSQL store)

1. Arrange any event in the store whose payload fails to fold (easiest trigger: a payload that deserializes with null required fields — see companion issue about case-sensitive deserialization).
2. Execute a Decider list query that goes through `InMemoryObjectAccessor.CreateActor` → `IEventStore.ReadAllSerializableEventsAsync` → `GeneralMultiProjectionActor.AddSerializableEventsAsync`.
3. The per-event fold throws; `CreateActor`'s catch-all swallows it; the query returns an empty list with success semantics.

## Why this matters

The swallow converts a loud correctness bug into a silent data-visibility bug: "the fold is crashing" presents as "there is no data". In production this masked the fact that durable events existed but could not be projected — the most expensive kind of failure to notice. It also amplifies any producer-side payload defect (like the casing issue) into silent empties.

## Suggested improvement

- Rethrow (or fail the query) on fold exceptions during catch-up, or at minimum surface a distinct "projection failed" result instead of an empty success.
- If resilience-by-skipping is intentional for poison events, make it explicit and observable: a configurable policy plus a counter/log with event id + type + exception, never a bare catch-all.

Found while building Sekiban Cloud (SekibanAsAService) — companion issue: the case-sensitive deserialization null-field behavior. Full standalone repro available on request.

tomohisa:
Fixed in **[dcb-v10.5.0](https://github.com/J-Tech-Japan/Sekiban/releases/tag/dcb-v10.5.0)** (#1079).

[...]

Issue #1074, opened 2026-07-16
Sekiban.Dcb: EventTypes.DeserializeEventPayload silently yields null properties on payload-casing mismatch (case-sensitive camelCase, no fail-loud or tolerant option)


tomohisa:
## Summary

`DcbDomainTypes.EventTypes.DeserializeEventPayload` deserializes payload JSON with **case-sensitive camelCase** property matching. When the stored payload JSON uses a different casing (e.g. PascalCase `"InvitationId"` instead of `"invitationId"`), it does **not throw** — it returns a payload object with **all mismatched properties silently null/default**, even though the JSON demonstrably contains the values.

Downstream, a projector fold over such a payload typically throws (e.g. `InvalidOperationException: Value cannot be null. (Parameter 'key')`), and combined with #1075 the end result is a silently empty query — extremely hard to diagnose.

## Reproduction (verified on 10.2.2 and 10.4.0, .NET 10, real PostgreSQL store)

1. Register domain types via `DcbDomainTypes` (full registration graph, e.g. a `Create()` composing event types + projectors).
2. Store an event whose payload JSON is PascalCase. In our case this happened through `IEventStore.WriteEventsAsync` with an app-side serializer that produced PascalCase (app bug on our side — but any legacy/foreign-written rows hit the same path). Raw `dcb_events.Payload` shows `{"InvitationId": "...", ...}`.
3. Read back via `IEventStore.ReadAllSerializableEventsAsync` and call `EventTypes.DeserializeEventPayload` for the event type.
4. Result: a payload instance whose properties are all null/default. No exception, no warning.

Control: the same byte-identical JSON deserializes correctly with plain `System.Text.Json` when `PropertyNameCaseInsensitive = true`; and payloads written via `WriteSerializableEventsAsync` (pre-serialized canonical camelCase) round-trip byte-identically and fold correctly.

## Why this is worth hardening

A casing mismatch is a data/producer bug, but the failure mode makes it nearly undiagnosable: values visibly present in the stored JSON come back as null. We spent a multi-round investigation ruling out type identity, duplicate registration, call ordering, and plain STJ behavior before pinpointing this.

## Suggested improvement (either or both)

- **Fail loud**: after deserialization, if the JSON contained members that did not bind to any property (or required/key properties are null), throw a descriptive exception naming the type and the unbound members.
[...]

PR #1077, opened 2026-07-17
Event-payload binding integrity: fail loud on casing mismatch, explicit deserialization policies

### LINKED ISSUE #1076: SEK-G13 dcb: event-payload binding integrity — fail loud on casing mismatch, explicit deserialization policies
# SEK-G13 dcb: event-payload binding integrity — fail loud on casing mismatch, explicit deserialization policies

## Goal

Stop `EventTypes.DeserializeEventPayload` from returning all-null "successes" when
the stored payload's property casing does not match: fail loud by default on
case-only member mismatches (with a descriptive, secret-free exception), and add an
explicit deserialization policy with StrictUnmapped and CaseInsensitiveLegacy
opt-ins — identically for the reflection (Simple) and source-generated (AOT)
pipelines.

## Why This Slice Exists Now

Issue #1074 (operator-filed; reproduced on 10.2.2/10.4.0, .NET 10, real
PostgreSQL): a PascalCase payload demonstrably containing values deserializes into
an instance whose properties are all null/default — no exception, no warning.
Combined with #1075's swallow this surfaced as a silently empty query and took a
multi-round investigation to diagnose. The failure mode converts a producer-side
data bug into an undiagnosable reader-side mystery.

## Current Observed State

- `DcbDomainTypes.Simple` always injects camelCase **case-sensitive**
  `JsonSerializerOptions`; `SimpleEventTypes`/`AotEventTypes` call
  `JsonSerializer.Deserialize` with no post-bind validation.
- Whole-options injection already exists (constructor accepts an optional
  instance) — what is missing is explicit, safe policy semantics.
- Canonical library writes use the same options (camelCase), so correct rows
  cannot false-positive a casing check.
- `UnmappedMemberHandling.Disallow` exists but rejects genuinely unknown additive
  fields from newer writers — not acceptable as an unconditional default.

- Metadata-declared required members (C# `required` / `JsonRequired`) are the only
  safe required-enforcement mechanism; blanket null/default validation is unsafe
  and out of scope.
- SekibanWasmRuntime consumes the AOT pipeline (source-generated `JsonTypeInfo`);
  policy parity there is mandatory.
- SEK-G14 will reuse this slice's deterministic malformed-payload fixture.

## Target Repo / Path / Part

Target part: EventTypes deserialization path (Simple + AOT), policy surface, tests, docs.

## In Scope

- **Default: case-only mismatch fail-loud, top-level scope.** A policy-owned,
  side-effect-free preflight checks the payload's TOP-LEVEL member names against
  the ACTUAL `JsonTypeInfo` each pipeline uses: a member that fails to bind AND
  matches a declared property name except for casing throws a descriptive
  exception (event type, CLR type, offending JSON name, expected name, payload
  location; never payload values/secrets). Unrelated unknown fields stay ignored
  (forward compat). Recursive validation of nested objects/collections/converters/
  polymorphism is explicitly out of contract.
- **AOT-safe mechanics**: never mutate any `JsonSerializerOptions` or registered
  `JsonTypeInfo` (source-generated metadata may be bound to read-only context
  options). StrictUnmapped uses the same preflight for AOT when metadata cannot
  safely adopt `UnmappedMemberHandling.Disallow`; CaseInsensitiveLegacy uses
  policy-compatible metadata/deserialization, never read-only-context mutation.
- **Ambiguous-name rejection**: two declared JSON names equal under
  OrdinalIgnoreCase but ordinally different → deterministic failure at policy
  construction/deserialization (never silently pick a member).
- **Policy enum** on DcbDomainTypes/Simple/AOT construction:
  `FailOnCaseMismatch` (default) / `CompatibleCaseSensitive` (pre-G13 escape
  hatch) / `StrictUnmapped` (opt-in) / `CaseInsensitiveLegacy` (migration-only
  opt-in, documented risk).
- Metadata-declared required members surface through the same descriptive wrapper.
- Identical semantics + tests for Simple (reflection) and AOT (source-gen).
- Reusable deterministic malformed-payload fixture in a shared test-support
  location (consumed by SEK-G14; no production API exposure).
- EN + JA docs incl. a 13_common_issues troubleshooting entry for #1074.

## Out Of Scope

- Blanket null/default validation (unsafe).
- StrictUnmapped or CaseInsensitiveLegacy as defaults.
- Projection/actor swallow fixes (SEK-G14).
- Rewriting stored legacy data (docs guidance only).

The child PR must make case-only member mismatches fail loud by default with a
descriptive secret-free exception while continuing to ignore unrelated unknown
fields, add the four-value deserialization policy with identical Simple/AOT
semantics, enforce only metadata-declared required members, never mutate
caller-owned options, ship the reusable malformed-payload fixture, cover
everything with reflection AND source-generated tests, and update EN + JA docs
including the issue #1074 troubleshooting entry — additive public API only.

## Acceptance Criteria

- Default fails loud on TOP-LEVEL case-only mismatches (descriptive exception: event type, CLR type, offending/expected names, location; no payload values); unrelated unknown fields ignored; nested/recursive validation explicitly out of contract (pinned by test); correct camelCase rows unaffected.
- Policy-owned side-effect-free preflight operates on the ACTUAL JsonTypeInfo per pipeline; no JsonSerializerOptions or registered JsonTypeInfo is ever mutated; StrictUnmapped for AOT via the same preflight when metadata cannot adopt Disallow; CaseInsensitiveLegacy via policy-compatible metadata (no read-only-context mutation).
- Ambiguous case-folded declared names are rejected deterministically (tested).
- Policy enum with FailOnCaseMismatch default, CompatibleCaseSensitive escape hatch, StrictUnmapped opt-in, CaseInsensitiveLegacy migration-only opt-in.
- Required enforcement via metadata declarations only (delegated to STJ); docs recommend required/JsonRequired for identifiers.
- Simple and AOT pipelines have identical policy semantics with source-gen tests.
- Tests: canonical round-trip unchanged; top-level PascalCase fails (default) / binds (CaseInsensitiveLegacy); nested casing does not trigger; additive unknown ignored (default) / rejected (StrictUnmapped); ambiguous-name rejection; required-missing fails; options/typeinfo non-mutation; AOT parity; malformed fixture in shared test-support for SEK-G14.
- Docs EN + JA updated (policies + troubleshooting entry).
- Public API additive-only (DS-B1 minor-safe: only current false-success inputs change behavior).

## Verification

## Related Links

- https://github.com/J-Tech-Japan/Sekiban/issues/1074
- Host design record: intents/sekiban/design/silent-failure-hardening.md

## Knowledge Maintenance

- Intent placement: host design doc silent-failure-hardening.md (slice 1).
- Diagram candidate: none.
- Docs update: 13_common_issues.md EN/JA — in scope.
- Closeout writeback expected: yes — fixture location for SEK-G14 and any Simple/AOT semantics divergence.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1076

## The failure this ends

Issue #1074: a PascalCase payload that plainly held values deserialized into an instance whose every property was null — no exception, no warning. The domain reads camelCase, case-sensitively; System.Text.Json binds a non-matching member to its default and reports success. A producer-side data bug became an all-null instance on the reader side, and #1075's swallow turned it into a silently empty query that took a multi-round investigation to diagnose.

## The check

A **side-effect-free preflight** runs before every event-payload bind, in one shared component — `EventPayloadBinder` — that both the reflection (Simple) and the source-generated (AOT) pipeline call, so they cannot drift. It reads the declared JSON member names off the `JsonTypeInfo` and compares them to the payload's **top-level** member names:

- a member that does not bind **and** matches a declared name except for casing throws `SekibanEventPayloadBindingException` — event type, CLR type, offending JSON name, expected name, `$.location`, and **never a payload value**;
- a genuinely unknown member (a newer writer's additive field) is still **ignored** — forward compatibility is untouched;
- two declared names equal under case folding but different with it are **refused deterministically**, never silently picked.

Top-level only, by contract. Nested objects, collections, converters and polymorphism are explicitly out of scope — a half-recursive check that fired inconsistently would be worse than a clearly-bounded one. There is a test pinning that nested casing does **not** trip it.

## Four policies

| policy | top-level case mismatch | unknown field | when |
|---|---|---|---|
| `FailOnCaseMismatch` (default) | throws | ignored | new systems; catches #1074 |
| `CompatibleCaseSensitive` | binds to null (pre-G13) | ignored | temporary escape hatch |
| `StrictUnmapped` | throws | throws | reject any non-exact payload |
| `CaseInsensitiveLegacy` | binds (top-level) | ignored | read legacy mis-cased rows during migration |

Set on `DcbDomainTypesExtensions.Simple(...)` / the AOT builder. The default changes behaviour only for inputs that were already silently wrong (DS-B1 minor-safe); correct camelCase rows are never touched.

## Never mutates caller options or TypeInfo

The packet's hard constraint. The preflight only **reads** names. Each pipeline deserializes its own natural way — reflection through `options+type`, AOT through its source-gen `JsonTypeInfo` — handed to the binder as a delegate, so the shared code never touches a `JsonSerializerOptions` or a registered `JsonTypeInfo`. A test drives every policy over a source-gen type and asserts the options' `PropertyNameCaseInsensitive` / `UnmappedMemberHandling` and the typeInfo's property names are unchanged afterward.

A real hazard surfaced here and is worth stating: a fresh `JsonSerializerOptions` with no resolver **throws** on `GetTypeInfo` (unlike `JsonSerializer.Deserialize`, which auto-attaches one). My first cut resolved the names off the caller's options and silently no-op'd the entire check on the Simple pipeline whenever the options hadn't been used yet — which the reflection half of the test matrix caught red. The reflection pipeline now reads declared names via a private options that mirrors only the caller's naming policy, so the check always runs and the caller's options are still never touched.

## Required members

Enforced only through metadata — C# `required` / `[JsonRequired]` — surfaced through the same descriptive exception. Blanket null/default validation is deliberately not done: it would reject legitimately-default values. Docs recommend `required` for identifiers.

## Tests — identical over both pipelines

Every case runs twice from one theory (`Pipeline.Reflection` and `Pipeline.SourceGenerated`) because a silent failure only one pipeline catches is still a silent failure:

- canonical camelCase binds under every policy (the control);
- top-level PascalCase: fails by default / binds under `CaseInsensitiveLegacy` / binds-to-null under `CompatibleCaseSensitive`;
- nested casing does **not** trip the top-level check;
- additive unknown ignored by default / rejected under `StrictUnmapped`;
- ambiguous declared names rejected; required-missing fails, required-present binds;
- options + source-gen typeInfo unchanged after every policy runs.

**Mutation-tested, four killed**: silence the case-mismatch throw (4 fail), silence the StrictUnmapped throw (2), silence the ambiguity throw (2), drop the legacy rename (2).

The malformed payloads live in a new shared **`Sekiban.Dcb.TestSupport`** project so SEK-G14 reuses the exact same bytes; it is `IsPackable=false`, not a product package.

## One existing test was the bug in miniature

`SerializedCommitTests` serialized with a bare `JsonSerializer.Serialize` (PascalCase) and committed to a camelCase domain — it was manufacturing the #1074 payload, and the new default correctly rejected it. Fixed to serialize through the domain's options (the canonical write path). It was building the bug, not exercising the commit path.

## Evidence

- Additive public API only; no signature removals (`IEventTypes` unchanged).
- `Sekiban.slnx` builds clean (`--no-incremental`, 0 errors).
- WithResult **682** (657 + 25 new binding-integrity, both pipelines), WithoutResult 76, Orleans 84 / 1 skipped, ColdEvents 88 — every pre-existing suite green.
- `git diff --check` clean. Docs EN/JA: `13_common_issues` gains the #1074 entry and the policy table.


tomohisa:
SEK-G13 strict review — request update

Blocking implementation finding: the Simple preflight does not use metadata semantically identical to the actual deserializer, violating the packet's ACTUAL JsonTypeInfo / Simple-AOT parity acceptance criterion.

At head 9f00a461a246d60680af68d3327ffe5728d7af21, SimpleEventTypes.DeserializeEventPayload binds with JsonSerializer.Deserialize(effectiveJson, eventType, _jsonOptions) (SimpleEventTypes.cs:71), but TryResolveNamesTypeInfo always constructs a separate options object containing only PropertyNamingPolicy, PropertyNameCaseInsensitive, and a fresh DefaultJsonTypeInfoResolver (lines 146-156). Therefore caller TypeInfoResolver metadata/modifiers and converter-driven contracts are not the metadata checked by the preflight.

Concrete failures:
- A caller resolver modifier can rename a property to wireName. The real deserializer binds wireName, while the private resolver still declares the convention-derived name. StrictUnmapped can reject a valid payload as unknown.
- For a case-only mismatch of a resolver-modified declared name, FailOnCaseMismatch can treat the JSON key as genuinely unrelated and permit the same silent default/null bind this slice must prevent.
- A modifier can remove/ignore a member while the private resolver still treats it as declared, producing false case-mismatch decisions.
- If the names-only reflection resolver cannot describe the type, lines 61-72 explicitly fall back to pre-G13 unvalidated deserialization, so the default fail-loud guarantee becomes conditional without a policy-visible failure.

The current 25-test matrix passes on both net9.0 and net10.0, but Build() always supplies only camelCase options; it has no custom TypeInfoResolver/modifier/converter case and therefore cannot catch this divergence.

Required unblock:
1. Make Simple preflight consume the exact effective metadata/contract used by options+type deserialization, without mutating caller options. If exact metadata is unavailable, fail deterministically under metadata-requiring policies rather than silently reverting to pre-G13 behavior.
2. Add non-vacuous Simple tests with resolver modifiers covering renamed and ignored/removed members, StrictUnmapped validity, and case-only mismatch detection; cover converter/unsupported-metadata fallback behavior explicitly.
3. Preserve the already-correct AOT direct registered JsonTypeInfo path and all current top-level/non-mutation semantics.

Other reviewed areas were consistent: AOT directly uses registered source-generated JsonTypeInfo without mutation; top-level rewriting does not recurse; shared fixture is under dcb/tests/Sekiban.Dcb.TestSupport with IsPackable=false; SerializedCommitTests now serialize through canonical domain options; EN/JA docs and additive exception surface are in scope. No merge was performed.

tomohisa:
SEK-G13 rereview — request update (converter-only metadata remains blocking)

The resolver/modifier repair is materially correct: Simple now first uses caller _jsonOptions.GetTypeInfo, the resolver-null path uses a full JsonSerializerOptions copy plus DefaultJsonTypeInfoResolver, resolver rename tests are non-vacuous, and an actually unresolved type fails deterministically for metadata-requiring policies.

One blocking part of the prior finding remains. ResolveEffectiveTypeInfo treats any non-null JsonTypeInfo as authoritative, but a custom converter commonly produces JsonTypeInfoKind.None with an empty Properties collection. That is not a declared-name contract. The converter can bind arbitrary top-level names, while EventPayloadBinder.Preflight sees zero declared names:

- FailOnCaseMismatch treats every key as genuinely unrelated and allows the converter to run, so it can still silently default a case-mismatched member while the policy claims validation.
- CaseInsensitiveLegacy cannot derive a safe rename target.
- StrictUnmapped rejects every converter-owned key as unknown rather than applying an authoritative contract.
- This applies to Simple caller converters and to registered/source-generated AOT metadata using a custom converter.

The new tests cover resolver rename and a resolver returning null, but contain no JsonConverter or JsonTypeInfoKind.None case; the focused suite therefore passes 32/32 on net9 and net10 without exercising this boundary.

Required unblock:
1. In the shared binder/effective-metadata boundary, treat non-object/converter-only metadata (at minimum JsonTypeInfoKind.None) as unavailable for declared-name validation. Metadata-requiring policies must fail deterministically with SekibanEventPayloadBindingException unless a documented converter-specific authoritative-name mechanism exists. CompatibleCaseSensitive may continue unchecked and must prove the converter actually runs.
2. Add non-vacuous Simple and AOT/source-generated converter tests: a converter that binds a known top-level shape, metadata-requiring policies fail before claiming validation, and CompatibleCaseSensitive successfully uses the converter.
3. Preserve the now-correct exact caller resolver path, faithful resolver-null copy, modifier rename semantics, and all prior accepted behavior.

Current head: 0b0548b3ea58142d0680a197c9eba0b14e9e79ef. All GitHub checks are green, but tests are necessary and not sufficient. No merge was performed.

tomohisa:
SEK-G13 final rereview — one minimal test gap remains

Production behavior is now correct: EventPayloadBinder rejects every non-Object JsonTypeInfo before name validation for all metadata-requiring policies, with a descriptive secret-free SekibanEventPayloadBindingException; CompatibleCaseSensitive returns before the guard. Ordinary source-generated Object metadata remains accepted. Simple converter tests cover FailOnCaseMismatch, StrictUnmapped, and CaseInsensitiveLegacy, and both Simple/AOT Compatible tests assert the converter marker, proving converter execution.

Blocking evidence gap required by this rereview: Aot_ConverterOwnedTypeInfo_FailsDeterministically_UnderMetadataRequiringPolicies has InlineData only for FailOnCaseMismatch and StrictUnmapped. CaseInsensitiveLegacy is absent. The production code is shared, but the AOT construction/registration path must be proven for all three policies; shared-code inference is explicitly insufficient here.

The focused net9/net10 run passes 39 tests per TFM. Adding the missing AOT CaseInsensitiveLegacy InlineData row should make the intended matrix 40 per TFM and directly prove the final parity requirement.

Required unblock: add EventPayloadDeserializationPolicy.CaseInsensitiveLegacy to the AOT converter-owned theory and confirm it fails with the same Kind=None policy-visible exception. Preserve the current Kind guard and all accepted behavior.

Head: 72109f43f957f47a10f5476fb648de89c27eaae5. All GitHub checks are green. No merge was performed.

tomohisa:
SEK-G13 final rereview — completed, no blocking findings

Head 5556ac8522fef181a7f92d014cd5a42585732776 resolves the final evidence gap. The delta from 72109f43 is exactly one test-only InlineData row adding CaseInsensitiveLegacy to the AOT converter-owned metadata theory; there is no production change and git diff --check is clean.

The converter parity matrix now proves:
- Simple and AOT reject JsonTypeInfoKind.None deterministically for FailOnCaseMismatch, StrictUnmapped, and CaseInsensitiveLegacy.
- Simple and AOT CompatibleCaseSensitive both execute the converter, proven by the bound-by-converter marker.
- Ordinary Object metadata, exact caller resolver modifiers, resolver-null faithful fallback, top-level-only semantics, unknown/ambiguity/required handling, non-mutation, secret-free exception, shared fixture, canonical serialization, and EN/JA docs remain intact.

Independent focused run:
dotnet test dcb/tests/Sekiban.Dcb.WithResult.Tests/Sekiban.Dcb.WithResult.Tests.csproj -c Release --filter EventPayloadBindingIntegrityTests|SimpleEventTypesResolverContractTests|ConverterOwnedPayloadTests
Result: 40/40 pass on net9.0 and 40/40 pass on net10.0. All four current GitHub checks are SUCCESS.

Tests are necessary but not sufficient; approval also rests on the SEK-G13 packet/review-context contract and intents/sekiban/design/silent-failure-hardening.md alignment. The PR remains within dcb/src, dcb/tests, and required EN/JA docs; Closes #1076; no merge was performed.

PR #1079, opened 2026-07-17
Projection apply-failure integrity: no swallowed folds, faulted projections fail queries

### LINKED ISSUE #1078: SEK-G14 dcb: projection apply failure integrity — no swallowed folds, faulted projections fail queries
# SEK-G14 dcb: projection apply failure integrity — no swallowed folds, faulted projections fail queries

## Goal

A crashing projection fold must never present as an empty successful query. Remove
the InMemory replay swallow (and the ignored failed read) so original exceptions
reach the existing ResultBox/GuardedUnwrap boundaries, and give Orleans
multi-projections a confirmed-fault state that fails normal queries with fault
context until resolved — while ordinary catch-up lag stays allowed.

## Why This Slice Exists Now

Issue #1075 (operator-filed; reproduced on 10.2.2/10.4.0, .NET 10, real
PostgreSQL): a per-event fold exception during in-memory catch-up is swallowed by a
bare catch and the list query returns an empty success. In production this masked
"durable events exist but cannot be projected" as "there is no data" — the most
expensive failure to notice. Investigation confirmed the same class exists in the
Orleans production path: background catch-up records `_lastError` and eventually
stops, while `GetState`/queries keep returning current/empty state.

## Current Observed State

- `InMemoryObjectAccessor.CreateActor`: bare catch around replay (a comment even
  documents the empty-on-failure behavior) + a failed
  `ReadAllSerializableEventsAsync` ResultBox is ignored (only IsSuccess handled).
- The outer `GetActorAsync` already converts exceptions to `ResultBox.Error` — the
  plumbing to surface failures exists; the inner swallow defeats it.
- `GeneralMultiProjectionActor.AddSerializableEventsAsync` has no catch — the
  offending exception is available.
- Orleans: timer catch-up catches fold failures (`_lastError`, retries, stops);
  `ProcessEventBatch`/`ProcessBufferedSerializableEvents` catch/log/return;
  queries can return current/empty state despite the fault. Explicit query errors,
  TagStateGrain, and GeneralQueryExecutor already fail correctly.
- SEK-G13 provides a deterministic malformed-payload fixture for end-to-end tests.

- SEK-G9's GuardedUnwrap boundaries preserve original exceptions and add
  Exception.Data context — carried exceptions must not be re-wrapped in-process.
- SEK-G13 (merged) supplies the poisoned-payload fixture.
- Observational/maintenance catches (safe-promotion, size estimation, GC) are
  legitimately best-effort and stay, documented as non-authoritative.

## Target Repo / Path / Part

Target part: InMemoryObjectAccessor replay path, Orleans multi-projection fault state, Core fault descriptor, tests, docs.

## In Scope

- **InMemory**: remove the bare replay catch; surface failed reads. Original
  exception preserved to `GetActorAsync`'s existing ResultBox.Error boundary.
- **Orleans confirmed-fault lifecycle (precise contract)**:
  - *Creation*: only at the per-event authoritative deserialize/fold boundary,
    with event id/type/projector/position captured BEFORE apply (a batch-level
    catch that cannot name the poison event does not qualify). Partial-batch
    semantics defined: per-event progress or batch atomicity so retry never
    double-applies earlier events.
  - *Durability*: activation/restart can never answer success before the fault is
    re-established — either a persisted grain fault descriptor restored on
    activation, or mandatory synchronous catch-up before the first query (choose
    one, document it). Fresh-activation first-query failure is tested.
  - *Clearing*: only successful replay of the same position after a deterministic
    rebuild/reset, or explicit operator reset/rebuild. Unrelated later success
    never clears a fault.
  - *Query contract*: state, scalar, and list surfaces all fail consistently with
    fault context while unresolved; ordinary lag (no fault) fails nothing.
- **Descriptor/exception separation**: a Core-level serializable
  `ProjectionFaultDescriptor` is distinct from the carried exception. In-process:
  original exception instance rethrown (never wrapped) with descriptor-sourced
  Exception.Data context. Orleans: descriptor stored/transported; boundary
  exception preserves type/message/context (no reference-identity promise).
  No WithoutResult-internal helper dependencies.
- No new result type (use existing failure channels).
- Parity tests incl. fresh-activation, fault-not-cleared-by-unrelated-success,
  partial-batch retry, subscription/publisher best-effort paths, WithResult +
  WithoutResult, fault-vs-lag across all query surfaces, healthy-path invariance.
- EN + JA docs incl. issue #1075 troubleshooting entry.

## Out Of Scope

- Poison-event skip/quarantine policy (future explicit opt-in — never default).
- Observability counters/audit tooling (future packet if evidence warrants).
- Changing observational/maintenance best-effort catches beyond documentation.
- New public result types.

The child PR must remove the in-memory replay swallow and failed-read ignore so
original fold/read failures surface through existing boundaries without
re-wrapping, introduce a confirmed-fault state for Orleans multi-projection
apply/catch-up that fails normal queries with fault context (event id/type/
projector/position) while leaving ordinary catch-up lag unaffected, use a
Core-level fault descriptor with the established Exception.Data context keys,
prove behavior with fixture-driven parity tests across in-memory and Orleans and
both facades, keep healthy paths byte-for-byte unchanged, and update EN + JA docs
with the fault-vs-lag contract and the issue #1075 troubleshooting entry.

## Acceptance Criteria

- InMemory: bare catch removed; failed reads surface; original exception reaches ResultBox.Error boundary un-wrapped.
- Fault creation only at the per-event deserialize/fold boundary with metadata captured before apply; partial-batch semantics (per-event progress or atomicity) prevent double-apply on retry.
- Fault durability: fresh activation/restart never answers success before fault re-establishment (persisted descriptor OR mandatory synchronous catch-up — one chosen and documented); fresh-activation first-query failure tested.
- Fault clearing only via same-position successful replay after deterministic rebuild/reset or explicit operator reset; unrelated later success never clears (tested).
- Query contract: confirmed fault rejects state/scalar/list surfaces consistently with fault context; lag without fault fails nothing (tested across all surfaces).
- Core-level serializable ProjectionFaultDescriptor separate from the carried exception; in-process original-instance rethrow with Exception.Data context; Orleans preserves type/message/context (no reference-identity promise); no WithoutResult-internal dependencies.
- No new result type; poison-skip not implemented (documented future opt-in).
- Fixture-driven tests (shared test-support fixture from SEK-G13): in-memory poisoned event → error (both facades); failed read → error; Orleans fault → query fails with context; fresh-activation; unrelated-success non-clearing; partial-batch retry; subscription/publisher best-effort cannot hide authoritative failure; healthy paths unchanged.
- Docs EN + JA (fault-vs-lag contract + troubleshooting).
- DS-B1 minor-safe: only current false-success flows change behavior.

## Verification

## Related Links

- https://github.com/J-Tech-Japan/Sekiban/issues/1075
- Host design record: intents/sekiban/design/silent-failure-hardening.md
- Depends on SEK-G13 (malformed-payload fixture)

## Knowledge Maintenance

- Intent placement: host design doc silent-failure-hardening.md (slice 2).
- Diagram candidate: none.
- Docs update: 13_common_issues.md EN/JA — in scope.
- Closeout writeback expected: yes — implemented Orleans fault-state shape/recovery path and whether a poison-skip/observability packet is warranted.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1078

## The failure this ends

Issue #1075: a per-event fold that crashed during in-memory catch-up was swallowed by a bare catch, and the list query returned an **empty success**. In production this masked *"durable events exist but cannot be projected"* as *"there is no data"* — the most expensive kind of failure to notice, and it took a multi-round investigation to find. The Orleans path had the same class: background catch-up recorded `_lastError` and eventually stopped, while queries kept answering current/empty state as a success.

## InMemory — the swallow is gone (both facades)

The bare replay catch in `InMemoryObjectAccessor` is removed. A failed read and a failed fold both propagate, so the original exception reaches `GetActorAsync`'s **existing** `ResultBox.Error` boundary un-wrapped (`GetAwaiter().GetResult()` over `Wait()`, so the fold's own exception surfaces, not an `AggregateException`). The plumbing to surface failures already existed — the inner swallow defeated it.

## Actor fault state — the confirmed fault, at the one place it can be named

A per-event fold **or** deserialize failure captures a Core-level `ProjectionFaultDescriptor` (event id / type / projector / position, taken **before** apply, so it points at the poison) and **rethrows the original exception**, annotated with the fault's identifiers on `Exception.Data` under SEK-G9's boundary keys. While faulted:

- `GetStateAsync` — the choke point every **state, scalar and list** query fetches first — fails with the fault instead of handing back the last pre-crash payload;
- applies are **rejected**, so a catch-up retry cannot re-run the pre-poison events and double-count them (the partial-batch semantics: stop at the poison, reject retries);
- the fault clears **only** by a rebuild (a fresh actor, or `ClearFaultForRebuild`) — an unrelated later success never clears it, because a faulted projection has no successes to be had.

## Orleans — the laundering, and the durability

`NativeProjectionQueryExecutor` stopped laundering: a failed state fetch or a failed handler result now returns `ResultBox.Error` instead of an empty `TotalCount=0` success, so the grain's existing `if (!result.IsSuccess) throw` surfaces the fault. **Ordinary catch-up lag returns a successful partial state and is untouched** — only a crash fails a query.

Durability: the fault is persisted in `MultiProjectionGrainState` (`[Id(13..17)]`) and **re-established into the host on activation before the background catch-up starts**, so a freshly-activated grain fails the first query rather than briefly answering empty. A faulted projection makes no safe-checkpoint progress, so the snapshot persist would short-circuit; the fault is written to grain state directly to survive that. A full operator rebuild clears it.

If the descriptor write itself **fails**, the grain does not deactivate (which would discard the only record of the fault and reopen the empty-success window). It **pins** the current activation — whose live in-memory fault already fails every query — and runs a **product-owned retry** (capped exponential backoff with bounded deterministic jitter; every grain-state write in the activation is serialized through a single gate, so the retry can never race a normal snapshot/checkpoint persist on the ETag) that re-attempts the write until it lands. Once the descriptor is durable the pin is released, so a subsequent fresh activation restores it and fails closed.

The former residual restart window — a process loss *while the state store was still down*, which loses the un-persisted descriptor — is now **closed synchronously**. A fresh activation that restored **no** descriptor runs a **first-query barrier**: before the first state/scalar/list answer it reads the **authoritative event-store head**, and if the projection is behind it catches up **in-call** — re-reading the tail, re-folding the poison, and re-establishing the fault **before any success**. All first callers share **one in-flight barrier task**, so concurrent first callers await the same work and all observe its outcome; none bypasses into current/empty state while it runs.

The barrier **never swallows** a head/read/catch-up failure. If the head read fails, a catch-up read fails, or catch-up cannot be shown to have reached the head, the barrier **throws the original exception** and is **not** marked complete — so the query fails closed (the original exception is preserved through `ResultBox.Error`) and the **next** query re-runs it; no query succeeds on empty state because a flag was cleared early. A re-established projection fault is the one non-throwing completion: the actor is left faulted and the query surfaces it as `ResultBox.Error`. Catch-up runs in-call (not on the non-reentrant catch-up timer, which the query turn would deadlock waiting on). Healthy already-caught-up projections are a fast no-op, and an already-active projection's ordinary lag is untouched — the barrier is per-activation, fires once, and only when the activation restored no fault.

Writer safety is enforced by **type ownership, not convention**. The grain's raw `IPersistentState<MultiProjectionGrainState>` is handed to an internal `CoordinatedGrainStateStore` in the constructor and the grain keeps **no direct reference** — and no `MultiProjectionGrainState` field either. Reads escape only as a **true immutable snapshot** (`store.Committed` returns a distinct `MultiProjectionGrainStateSnapshot` record of copied scalar/string values — NOT the persisted payload, so a consumer cannot downcast to a mutable reference). The one field that advances in live runtime ahead of any persist (`LastPosition`) is kept as an ephemeral grain field, not mutated on the persisted payload outside the store.

Every write goes through `ExecuteWriteAsync(kind, mutate)` (or the token-guarded `ExecuteGuardedWriteAsync`), which is **copy-on-write under a single-writer gate**: it clones the last committed state, applies the mutation to the **candidate**, writes the candidate, and publishes it as committed **only after** a successful `WriteStateAsync`. If the write fails, the candidate is discarded and the committed state (and the provider payload) is restored — no uncommitted field is ever visible to a later read or another write, and the exception propagates so the caller's post-await code (e.g. clearing the live actor fault) never runs. The store's responsibility ends there; a **live side effect happens only after the caller awaits a successful write** — so the `OperatorReset` candidate clears only the *candidate* fault fields, and the live actor fault (`_projectionFault` / `ClearFaultForRebuild`, via a non-throwing idempotent primitive) is cleared inline after the durable commit. A failed reset therefore leaves the persisted descriptor **and** the live fault intact, so queries stay faulted and retry stays coherent. Each write carries an explicit **category** (`Checkpoint`, `FaultDescriptor`, `OperatorReset`, `MetadataMaintenance`); a `Checkpoint` is **rejected** (no write, no version bump) when a fault exists on the committed state **or** live on the actor (fault persistence may still be retrying), so a faulted projection makes no checkpoint progress — the G14 rule, enforced in the store. Both the store and the first-query barrier's single-flight core (`FirstQueryCatchUpGate`) are **internal, framework-free components** exposed to the test project via `InternalsVisibleTo` — no public test aid, no Orleans wire method — so they can be driven **directly** by friend tests rather than through the grain.

## Operator reset — the one authorized way to clear a fault

A persisted fault clears only through an **operator-only** admin surface: `IMultiProjectionGrain.ResetProjectionFaultAsync(ResetProjectionFaultRequest)` (same admin plane as `GetStatusAsync`). It is **never invoked automatically** and is **not** exposed through `ISekibanExecutor` — a structural test enforces both. The request carries the **exact current fault identity** (projector name, fault event id, fault stream position) as a **concurrency token**; the operator reads it from the fault context already surfaced on a failed query.

Inside the single-writer gate all **three token fields — projector name, fault event id, fault stream position — are validated as one atomic precondition** against the **current persisted descriptor** (the concurrency authority). Every request AND persisted field must be **non-null and non-empty** before any equality check — a null/empty value is never normalized to `""`, so a missing field is a non-match, not an accidental pass. A mismatch or missing value in **any** field, a stale token, a descriptor changed by a concurrent write, or no fault present is rejected (`ResultBox.Error`) with **no write, no external delete, no fault clear**; a same-token race commits at most once. On a correct token the reset touches two derived stores in order, all under the gate: the derived **external snapshot** (`IMultiProjectionStateStore`) for the projector/version is **invalidated first** (deleted — derived-state only, no authoritative events touched), then the descriptor **and** the derived checkpoint are durably cleared (copy-on-write). This is not one atomic transaction across the two stores, so the **external-first ordering** keeps every partial failure retry-coherent: if the external invalidation fails, the descriptor clear is skipped and nothing changed; if it succeeds but the descriptor clear fails, the snapshot is gone but the descriptor and live fault are retained, so queries stay rejected (a rebuild regenerates the snapshot). No in-flight upsert can race the delete: catch-up runs on an **interleaving** grain timer, so **every** external-store mutation goes through **one activation-local coordinator** (`ExternalStoreCoordinator`) — all three upsert paths (normal persist, streaming persist, version rewrite) **and every delete**, including the public `DeleteExternalStateAsync` admin call and the reset's own invalidation; **no direct-to-store bypass remains**. The delete waits for any parked/in-flight upsert to commit before it deletes, and the coordinator **rejects any upsert while a live or committed fault exists**, so a faulted projection persists no derived state and no stale upsert can recreate what the reset removes. That rejection is an **explicit failure, not a silent success**: a blocked upsert returns a `ResultBox.Error` carrying a stable `ExternalPersistenceBlockedByFaultException`, never a success carrying `false` — so a caller inspecting only `IsSuccess` cannot mistake it for a completed save. Normal and streaming persistence therefore never report "saved" or advance persisted metadata after a rejection, and `OverwritePersistedStateVersionAsync` keeps `updated = false` and performs **no** projector-version write. **Only after** the durable commit is the live actor fault cleared (a non-throwing idempotent primitive) and the projection **rebuilt in-activation** — a fresh host plus a re-armed first-query barrier, so the very next query synchronously catches up from the beginning before it can answer (no early healthy window, no reliance on a deactivation cycle). If the poison remains, the per-event boundary **re-establishes and re-persists** the fault on rebuild (never skip/quarantine).

## The fault-vs-lag contract

| situation | query |
|---|---|
| healthy, caught up | succeeds |
| healthy, catching up (lag) | succeeds (partial state) |
| **faulted** | **fails** with fault context — event id, type, projector, position; never a payload value |

## Descriptor / exception separation

`ProjectionFaultDescriptor` is a plain serializable record. In process the **original exception instance** is rethrown (never wrapped), carrying the descriptor's identifiers on `Exception.Data`. Where the original instance is gone — an Orleans activation that restored a persisted fault, or a fault surfaced across the grain boundary — `SekibanProjectionFaultException` re-raises it, preserving type/message/context (no reference-identity promise). It crosses the Orleans boundary as a ResultBox field via a registered surrogate/converter (`ProjectionFaultExceptionSurrogate`), since the Core exception carries a Core descriptor Orleans has no field codec for. No WithoutResult-internal dependencies; no new result type.

## Tests — fixture-driven, reusing SEK-G13's malformed payload

- **WithResult (8)**: poison faults and surfaces the original exception; `GetState` fails with context; unrelated success does not clear and is rejected; retry rejected not re-applied; rebuild clears; healthy unchanged; InMemory poison → error; failed read → error.
- **Best-effort parity (2)**: the in-memory publisher and subscription swallow their own delivery throw, but the fault still surfaces on the next read — the best-effort layer swallows the throw, not the fault.
- **WithoutResult (2)**: facade parity — the shared Core surfaces the fault through the exception-based facade too.
- **Orleans grain surface**: confirmed fault fails the **state, scalar and list** surfaces through the production query path with context; lag fails nothing; a **fresh activation with poison in the store but no restored descriptor fails the first query synchronously**; the fault is **durable without any manual persist and survives deactivation/reactivation**; a **persist-failure** test — an injected storage rejects the first fault-carrying write, the product-owned retry makes the descriptor durable on its own, and a genuine fresh activation then restores it and fails all first queries closed. First-query barrier surface: a **head-read failure** fails the first query closed with the **original exception type + message** preserved; an **event-read failure** fails **all three** surfaces closed (original exception preserved) and a **later query does not bypass** into empty success; and a **fail-first-then-recover** test — the store recovers and a later query re-runs the barrier, re-folds the poison and fails with the **fault** (never empty), proving the projection is not bricked.
- **Internal component friend tests** (`InternalsVisibleTo`, no Orleans): the **first-query barrier gate** driven directly — multiple callers share one in-flight task, the work runs **exactly once**, all observe the same success or the same failure, a failure is **retryable** and a success is **sticky**. The **external-store coordinator** driven directly (no Orleans): a parked upsert makes a delete **wait and run only after it commits**, and — the mirror — a parked delete makes an upsert **wait and run only after it completes**, pinning the serialization **both directions** (the guarantee the public delete's routing buys); and an upsert while faulted returns an **explicit `ResultBox.Error`** carrying `ExternalPersistenceBlockedByFaultException` (asserted `IsSuccess == false` and exception type) **without invoking the store delegate**, then proceeds once the fault clears. The **coordinated state store** driven directly with a recording persistent-state fake that **clones state on commit** and assigns a monotonic version: (1) a checkpoint and a fault-descriptor write compete, the first parked inside the write — **max one concurrent write**, versions 1 then 2, first commit checkpoint-only, second adds the fault while retaining the checkpoint; (2) a checkpoint write that **fails** rolls back — the uncommitted position/version is not visible to a later read nor carried by a later successful write; (3) with a fault committed first, a checkpoint is **rejected** — no second write/version bump, checkpoint unchanged, fault retained, while a contracted reset still writes; (4) a **live actor fault** (no committed descriptor) also suppresses the checkpoint until it clears.
- **Immutability + rollback tests**: `store.Committed` is a distinct snapshot at **runtime** — `IsNotType<MultiProjectionGrainState>`, `as MultiProjectionGrainState` is null, not assignable — so a downcast cannot reach the payload; a **failed `OperatorReset` write does not advance the committed state** (the descriptor stays), so a caller awaiting it skips its live clear, while a later successful reset clears the committed descriptor.
- **Operator-reset tests** (real TestCluster, driving `ResetProjectionFaultAsync` on a faulted grain): a **wrong projector**, **wrong event id**, and **wrong position** are each independently rejected with no provider write and queries stay faulted; a **same-token race** commits at most once; a **correct token with a now-foldable event** rebuilds **immediately on the same grain** (no manual deactivation) and state/scalar/list recover with **exact** values and the recovered position; a **correct token while the poison remains** rebuilds and **re-faults immediately**, with the re-fault persisted; a **fail-first persisted clear write** leaves the descriptor and the live fault intact (all surfaces rejected) and a retry then succeeds with exact recovery; a **genuine product-persisted pre-poison snapshot** restores healthy, then the poison faults on the production RefreshAsync path (descriptor persisted with no timing assistance), and the reset invalidates the snapshot and rebuilds to the exact state/scalar/list/position; **missing/empty projector, event id, or position** are each rejected with zero effect; and the two **cross-store partial failures** (external-delete-fails; grain-write-fails-after-external-delete) are each retry-coherent. **Fault-rejection semantics (not just "no v2 snapshot")** are proven per path: a **normal persist while faulted** reaches no external upsert (the store's upsert delegate count is unchanged) and stays faulted; a **streaming persist while faulted** (on the streaming-configured cluster) writes no external snapshot and stays faulted; a **version rewrite while faulted** — the one path that genuinely reaches the fault gate (it reads and rewrites an existing snapshot) — is rejected there, so it returns `false`, the store's upsert delegate is **never invoked**, **no** `MetadataMaintenance` projector-version write happens (grain write-count unchanged), no new-version snapshot appears, and the original snapshot is retained. The public **`DeleteExternalStateAsync`** is exercised end-to-end (routed through the coordinator, it removes the derived snapshot); its delete-vs-upsert serialization — masked at the grain level by Orleans non-reentrancy and reachable only via the interleaving timer — is pinned deterministically by the coordinator friend tests above.
- **Persisted-side (not request-side) malformed descriptor tests** (real TestCluster, genuine persist → corrupt-in-storage → reactivate → restore; only test-owned grain storage is touched, no production seam): a persisted **ProjectorName** or **FaultPosition** that is **null or empty** is genuinely restored, the projection stays faulted (the live fault re-establishes from the intact `FaultEventId`), and a **well-formed reset token is rejected** with zero external delete, zero grain/provider write (so no version/ETag change), zero external upsert, no live clear/deactivation, and state/scalar/list all continue to fail closed. **`FaultEventId` null/empty is documented as unreachable rather than asserted as a sixth still-faulted case**, with two evidence tests: with the **poison benign**, a null/empty persisted `FaultEventId` restores **no** live fault (`RestoreProjectionFaultIfPersisted` requires a parseable Guid) so the projection comes up **healthy**; with the **poison still active**, the fresh activation restores no fault, the **first-query barrier re-folds the poison and self-heals to a fresh valid live fault**, and once that valid fault is persisted the original well-formed token resets it — so a malformed `FaultEventId` never becomes a stuck, unresettable descriptor.
- **The non-empty guards are logically redundant against a well-formed token, and the tests reflect that honestly.** A malformed *persisted* field (null/empty) can never equal a well-formed (non-empty) request field, so the ordinal-equality clause already rejects it — removing the three persisted-side `!IsNullOrEmpty(committed.*)` checks leaves all four persisted-side tests **passing** (verified). The presence checks earn their place only against an **empty-identity false match**: three **both-empty false-match tests** corrupt persisted field X to `""` **and** send a token whose field X is also `""` (other fields well-formed and matching), and assert the reset is **rejected** with zero delete/write/live effects — removing the **paired** request+committed presence guards for that field makes `string.Equals("","")` spuriously match and **fails** these tests (verified), while the well-formed-token tests keep passing on the equality backstop. So the presence checks are covered for the one scenario where they are decisive, and no test claims six reachable restored-live-fault cases.
- **Architecture / structural tests** (reflection): `MultiProjectionGrain` holds **no** `IPersistentState` and **no** `MultiProjectionGrainState` field; the store owns the sole `IPersistentState` and exposes **no member returning the mutable state**; `ResetProjectionFaultAsync` is on the grain admin interface and **absent from `ISekibanExecutor`** with no automatic caller. Structural enforcement replaces the earlier source-string audit.
- **Mutation-tested**: ignoring the fault in `GetStateAsync`, allowing applies while faulted, re-swallowing the InMemory read, skipping the fresh-activation synchronous first-query catch-up, swallowing the barrier's head/read failure, losing the barrier's single-flight sharing, marking the barrier satisfied on failure, removing copy-on-write, removing the checkpoint fault-guard, bypassing the single-writer gate, re-introducing an `IPersistentState` field on the grain, **returning the raw mutable payload from `Committed`** (downcast becomes possible), running the `onCommitted`-style live clear **before** the durable reset await (a live clear would fire without a durable commit), removing an individual reset token field from the guard (a wrong projector/event-id/position would clear the fault), **skipping the external-snapshot invalidation** (a fresh activation could restore a pre-poison snapshot), **encoding the fault-block as a success carrying `false` instead of `ResultBox.Error`** (a version rewrite would then set `updated` and issue a `MetadataMaintenance` version write — caught by the coordinator friend test and the version-rewrite rejection test), **removing the fault gate entirely** (a faulted version rewrite would reach the store — caught by the same two tests), and **removing the paired request+committed presence guards for a token field** (an empty token would spuriously match an empty persisted identity — caught by the both-empty false-match tests, while the well-formed-token persisted-malformed tests keep passing on the equality backstop) each fail the corresponding test.

## Scope honesty

- Best-effort observational catches (safe-promotion, size estimation, GC, subscription cleanup) are **unchanged** — documented as non-authoritative, left alone as scoped.
- Poison-event **skip/quarantine is not implemented**: a projection silently skipping events it cannot apply would be a return to the same silence. Documented as a possible future explicit opt-in.

## Evidence

- Additive API only (interface members added with default implementations; no signature removals).
- `Sekiban.slnx` builds clean (`--no-incremental`, 0 errors).
- WithResult **707**, WithoutResult **78**, Orleans **139 / 1 skipped** (net9 and net10), ColdEvents **88**, MaterializedView **21**, MV.MultiProvider **9** — every pre-existing suite green.
- `git diff --check` clean. EN/JA docs add the #1075 fault-vs-lag contract and troubleshooting entry.

## Closeout note

The Orleans fault state is a **persisted descriptor restored on activation**, and the one residual restart window (a process loss while the state store was down) is now closed by a **first-query barrier scoped to fresh activations that restored no descriptor** — it reads the authoritative head, catches up in-call, and fails closed (never swallowing a head/read/catch-up failure) before answering. Healthy, already-caught-up projections stay a fast no-op, so activation latency for the healthy path is unchanged. Auto-recovery after a transient store outage is left to the existing background catch-up (the barrier resets to retry on each query until it can confirm the head), rather than asserted in a timing-dependent test. Whether a poison-skip/quarantine or observability-counter packet is warranted is left open; nothing here needs one, and adding either by default would reintroduce silence.


tomohisa:
Blocking findings for SEK-G14 at head `0582c83848bca6df5aebf9ff98f49d93f65baa39`:

1. **The confirmed Orleans fault is not durably persisted on the authoritative background apply paths.** `ProcessEventBatch` catches the fold failure, calls only synchronous `CaptureProjectionFaultIfAny()`, logs, and returns (`MultiProjectionGrain.cs:3634-3649`). `ProcessBufferedSerializableEvents` and catch-up failure handling have the same capture-without-`WriteStateAsync` shape. The only direct write is later inside `PersistStateAsync`. Consequently a silo loss/deactivation after the swallowed background failure but before a timer/manual persist can lose the descriptor and allow a fresh activation to answer success before re-encountering the poison. The fresh-activation test masks this gap by explicitly calling `grain.PersistStateAsync()` at `ProjectionFaultOrleansTests.cs:100`. Persist the descriptor as part of fault handling, await it before considering the failure handled, and fail closed/deactivate if that state write fails. Add a non-vacuous TestCluster test that does not manually persist, plus an injected persistence-failure test proving no later successful query.

2. **Required acceptance coverage is absent.** The changed tests contain no `InMemoryEventSubscription` / `InMemoryMultiProjectionEventPublisher` parity test required by acceptance (h), so their best-effort paths are not proven unable to hide an authoritative projection failure. Add non-vacuous tests through both production components.

3. **The Orleans query contract is only checked through snapshot JSON.** The packet requires state, scalar, and list query surfaces to all reject a confirmed fault, and ordinary lag to succeed across all three. `ProjectionFaultOrleansTests` has three tests and only exercises `GetSnapshotJsonAsync`; it does not traverse scalar/list query execution or verify `NativeProjectionQueryExecutor` preserves the fault. Add production-path TestCluster coverage for all three surfaces, including lag-vs-fault.

Design alignment checked against `.intent-cli/issues/SEK-G14/packet.yaml`, `review-context.md`, `implementation.md`, `github-body.md`, `intents/sekiban/design/silent-failure-hardening.md`, and the G14 diff. Managed worktree: `.intent-cli/worktrees/review-SEK-G14` (registered). GitHub head/CI reverified: head above, MERGEABLE, dcbTests net9/net10 and both Sonar checks SUCCESS. Independent focused Orleans net9 run passed 3/3 and `git diff --check` passed; those passing tests demonstrate the coverage gap rather than resolving it.

tomohisa:
SEK-G14 rereview at head `bce874d2bc9da8b8313c4c1a9d352ae545014d49`: one durability blocker remains.

The successful-write path is repaired: all three authoritative handlers now await `CaptureAndPersistProjectionFaultIfAnyAsync`, and its normal path awaits `WriteStateAsync`. The new best-effort publisher/subscription tests and state/scalar/list TestCluster coverage are also present and pass independently.

However, **fault-descriptor persistence failure is still not fail-closed across reactivation**. In `CaptureAndPersistProjectionFaultIfAnyAsync`, a failed `WriteStateAsync` is caught, logged, and followed only by `DeactivateOnIdle()`; the exception is not rethrown and the descriptor exists only in the activation being discarded. On the next activation `RestoreProjectionFaultIfPersisted()` has nothing to restore. `OnActivateAsync` then starts `CatchUpFromEventStoreAsync(forceFullCatchUp)` fire-and-forget, and `InitiateCatchUpIfNeeded` merely schedules a zero-due-time grain timer. Query preparation likewise calls `CatchUpFromEventStoreAsync`, which schedules/observes catch-up but does not synchronously re-encounter the poison before answering. Therefore a first state/scalar/list query can still return empty/current success before the timer re-establishes the fault—the exact contract gap this slice must close.

The added durability test covers only successful persistence and uses foreground `RefreshAsync`; it does not inject `WriteStateAsync` failure after poison capture and cannot kill this branch. Please add a deterministic harness that fails descriptor persistence at that exact point and proves both the current activation and the first fresh-activation state/scalar/list queries cannot succeed. The implementation must either retain/pin a fail-closed activation safely, or synchronously re-establish the poison before any fresh-activation query; logging plus deactivation is insufficient.

Resolved and reverified: awaited normal persistence at all three handlers; no-manual-persist successful durability; production-path scalar/list rejection and lag success; non-vacuous `InMemoryMultiProjectionEventPublisher` and `InMemoryEventSubscription` parity; prior per-event descriptor, partial-state query blocking, narrow clear, schema IDs, original in-process exception, docs and PR scope remain intact.

Design alignment checked against `.intent-cli/issues/SEK-G14/packet.yaml`, `review-context.md`, `implementation.md`, `github-body.md`, `intents/sekiban/design/silent-failure-hardening.md`, the full PR diff, and prior durable finding. Managed worktree: `.intent-cli/worktrees/review-SEK-G14` (registered). Verification: `git diff --check` pass; BestEffortPathFaultTests net9 2/2; ProjectionFaultOrleansTests net9 4/4. GitHub head is MERGEABLE and dcbTests net9/net10 plus both Sonar checks are SUCCESS.

tomohisa:
SEK-G14 rereview at head `beee76cc`: the restart-durability blocker remains.

The new storage harness is useful but proves only **same-activation** fail-closed behavior. It deterministically rejects writes containing `FaultEventId`, verifies the live actor's state/scalar/list queries fail, waits 1.5 seconds, and queries that same pinned activation again. It never recovers storage, never demonstrates an automatic retry succeeding, never forces deactivation/restart after durable recovery, and never checks the first queries on a fresh activation.

`DelayDeactivation(TimeSpan.FromDays(3650))` prevents idle reclamation; it does not survive silo/process loss. The descriptor remains unpersisted. The new production comment claims a lost process “re-reads the events on restart and re-faults,” but activation still starts catch-up fire-and-forget via a timer. With no persisted descriptor, a query can therefore answer empty/current success before poison is re-encountered. That violates the packet's activation/restart contract.

There is also no explicit retry lifecycle tied to `_faultPersistFailed` (the field is assigned but never read, as compiler warning CS0414 confirms). A periodic `PersistStateAsync` timer may eventually retry under default configuration, but the new test manually calls `PersistStateAsync`, and neither the product behavior nor a test proves fail-first → automatic retry after recovery → durable descriptor → forced fresh activation → FIRST state/scalar/list queries all fail. Manual persistence or an unrelated query is not sufficient evidence.

Please implement and prove one safe lifecycle:

- automatically retry descriptor persistence while keeping the live activation pinned, then permit/force fresh activation only after the descriptor is durable; or
- synchronously re-read/re-establish the poison before any query on an activation lacking a durable descriptor.

The deterministic test must fail specifically the first descriptor write after poison capture, recover the store, observe the product-owned retry (without manually invoking `PersistStateAsync`), force a genuine fresh activation, and assert the first real state/scalar/list queries cannot succeed. Also make docs/PR comments truthful about process-loss behavior.

Prior fixes remain resolved: all three normal handlers await descriptor writes; best-effort parity and production query surfaces are present; partial-state blocking, narrow clearing, schema compatibility, original in-process exception behavior, EN/JA docs, and diff scope remain intact.

Design alignment checked against the G14 packet, review-context, implementation, GitHub body, `intents/sekiban/design/silent-failure-hardening.md`, prior durable findings, and the full current diff. Managed worktree: `.intent-cli/worktrees/review-SEK-G14` (registered). Verification: `git diff --check` pass; current projection fault Orleans tests net9 5/5 pass, but the new test does not exercise the required recovery/restart branch. GitHub reported checks are all green.

tomohisa:
SEK-G14 rereview at head `431116518e45b8412d2c89a8a616ff84c70aaed1`: automatic transient recovery is now proven, but two blockers remain.

1. **The PR explicitly retains the forbidden restart window.** The PR body now states that if the process is lost while fault-descriptor persistence is failing, the descriptor is lost and background catch-up leaves a brief empty/current-success window after restart. That is a direct deviation from the authoritative packet: activation/restart must never return success before fault re-establishment. A doubly-degraded condition is not an authorized exception. The new fail-first test proves transient recovery when the process survives; it does not simulate loss before the retry lands. Close this with synchronous first-query catch-up/fault establishment when a fresh activation's durable checkpoint is behind the authoritative event-store head (without changing ordinary lag semantics for an already-active healthy projection), or obtain an explicit design/operator contract amendment. Background timer catch-up is not equivalent.

2. **The retry writer is not serialized and its outage policy is unsafe.** `_faultPersistRetryTimer` is registered with `Interleave = true` and calls the same `_state.WriteStateAsync()` used by normal snapshot/checkpoint persistence; no shared gate or single-writer primitive prevents overlap. Other persist timers are also interleavable, so concurrent writes/ETag races are possible. The retry is a fixed 500ms period forever and logs a warning with exception every attempt, yielding unbounded hot-loop pressure and log volume during an outage. Use one serialized grain-state writer across retry and normal persistence, dispose/cancel safely on lifecycle exit, and use a documented capped exponential backoff (with bounded jitter if appropriate). Prove mutation/concurrency relevance and that the pin is released only after the serialized durable write succeeds.

What is resolved: all four entry paths (`RefreshAsync` plus three authoritative background handlers) share capture/persist; the deterministic fail-first storage rejects exactly the first fault-bearing write, the product timer retries without manual persistence, a durable descriptor is read by a fresh activation, and first state/scalar/list calls fail. The focused net9 test passes 1/1 and `git diff --check` passes. Prior best-effort/query parity, partial-state blocking, narrow clear, schema compatibility, and original exception behavior remain intact.

Design alignment checked against the G14 packet, review-context, implementation, GitHub body, `intents/sekiban/design/silent-failure-hardening.md`, PR metadata, prior durable findings, and current full diff. Managed worktree: `.intent-cli/worktrees/review-SEK-G14` (registered). GitHub head is MERGEABLE and all four reported checks are SUCCESS.

tomohisa:
SEK-G14 strict rereview at head `7d602256417888fd1eafc49320a89dc1a039c2e0`: two blocking gaps remain.

1. **The first-query barrier fails open when authoritative catch-up itself fails.** `EnsureFirstQuerySyncCatchUpAsync` clears `_firstQueryNeedsSyncCatchUp` before doing any I/O, then catches every exception from `RefreshAsync`, logs it as though a projection fault was established, calls `CaptureAndPersistProjectionFaultIfAnyAsync`, and returns. If event-store head/read/catch-up fails before a per-event descriptor exists, `_host.CurrentFault` remains null; the state/scalar/list query then continues against empty/current state, and all later queries bypass the barrier because the flag is already false. The contract explicitly requires this path to fail closed with the carried store/catch-up failure. Only swallow the exception when an authoritative projection fault was actually established; otherwise rethrow/preserve it and keep/share the barrier failure semantics. Add deterministic store head/read failure tests for first state/scalar/list queries and concurrent first-query callers.

2. **The single-writer gate has no required concurrency proof.** Static audit shows all seven `MultiProjectionGrain` state writes now route through `WriteGrainStateGatedAsync`, including normal/streaming persistence, reset/rebuild, capture, and retry, with no remaining direct write outside the wrapper. The one-shot capped backoff and lifecycle disposal are materially improved. However, no test injects a blocking storage writer and drives retry plus normal/streaming persistence concurrently; no mutation test proves removing/bypassing the semaphore creates overlap or ETag inconsistency. The round explicitly requires this non-vacuous concurrency evidence. Add a deterministic max-in-flight/order assertion covering competing writer classes and mutation relevance; passing ordinary tests is insufficient.

Additional verification:

- Fresh activation with no restored descriptor now synchronously encounters poison before the first snapshot query; scalar/list then reject the established fault. The current test does not exercise store-read failure or concurrent first calls.
- The retry is one-shot, `Interleave=false`, capped exponential backoff (500ms–30s) with bounded deterministic jitter, debug logging after the initial critical event, and disposal on success/deactivation.
- The exception surrogate preserves descriptor id/type/projector/position/message/timestamp and reconstructs the fault exception without payload fields; additive grain-state IDs remain compatible.
- Prior per-event capture, partial-state blocking, narrow clear, best-effort parity, InMemory identity, EN/JA scope remain intact.

Design alignment checked against the G14 packet, review-context, implementation, GitHub body, `intents/sekiban/design/silent-failure-hardening.md`, PR metadata, prior durable findings, and full diff. Managed worktree: `.intent-cli/worktrees/review-SEK-G14` (registered). Verification: `git diff --check` pass; focused Orleans fault suites pass 6/6 on net9 and 6/6 on net10; GitHub head is MERGEABLE with all four checks SUCCESS. These passing tests do not cover the two branches above.

tomohisa:
SEK-G14 strict rereview at head `c7eec993`: the fail-closed implementation is improved, but the required proof introduces a production API leak and remains incomplete.

1. **Test-only state mutation leaked into the public production grain API.** `IMultiProjectionGrain.ProbeStateWriteConcurrencyAsync()` is now a remotely callable production method that causes two grain-state writes. Any consumer can invoke it; it expands the wire/API surface solely for a test and can create unnecessary ETag/version churn. Remove it from the public interface/runtime contract. Exercise the gate through an internal/friend/test hook or representative real production operations.

2. **The writer-gate test does not satisfy the required coherence proof.** It launches two identical synthetic writes and asserts only `MaxConcurrent == 1`. It does not record stable operation order, ETag/version progression, or the final persisted fault descriptor/checkpoint. It also does not represent normal/streaming persistence competing with fault retry, and the storage stores the same mutable state reference without modelling ETags. Add a deterministic harness where the first real category is parked, the second distinct production category is demonstrably ready, then assert order, monotonic version/ETag, and final descriptor/checkpoint coherence. The mutation bypass must fail those assertions, not merely a public probe's concurrency count.

3. **The first-query tests do not prove the claimed shared/retry lifecycle.** The two “concurrent” client calls target a normal non-reentrant grain method, so Orleans can serialize them before either concurrently awaits `_firstQuerySyncCatchUpInFlight`; the test does not observe a single shared barrier execution or prevent duplicate apply by instrumentation. The suite also never clears `FailReads`/`FailHeadReads` and performs a later successful retry, despite the contract requiring failure to leave the barrier retryable. Add deterministic counters/barriers showing concurrent callers join one execution, then store recovery causes a later query to retry once and reach head/fault correctly. Cover original exception type/context, not only message substrings.

Implementation audit notes:

- Head/read failure now fails closed and leaves `_firstQueryNeedsSyncCatchUp` set; poison re-establishment cannot proceed to successful query because the actor fault gate remains.
- `RefreshAsync(bool forceEvenIfCatchUpActive)` is private and normal public behavior retains its existing guard.
- All seven `MultiProjectionGrain` state writes still route through the semaphore wrapper with no direct nested acquisition found.
- Retry backoff/lifecycle, surrogate fields, schema IDs, partial-state/narrow-clear and prior parity remain intact.

Design alignment checked against the G14 packet, review-context, implementation, GitHub body, `intents/sekiban/design/silent-failure-hardening.md`, PR metadata, prior durable findings, and full diff. Managed worktree: `.intent-cli/worktrees/review-SEK-G14` (registered). Verification: `git diff --check` pass; new barrier/writer/persist tests net9 pass 5/5; GitHub reported checks are green. Passing tests are insufficient because of the vacuity/API issues above.

tomohisa:
SEK-G14 strict rereview at head `19a0f01b613dcac31f045949cfbd3f1de9fb37c8`: public API leak is fixed and the first-query gate is well isolated, but writer proof and current-head checks remain blocking.

1. **The seven-path audit is exactly the brittle string-count proof the delegation rejected.** `All_grain_state_writes_route_through_the_coordinator_seven_call_site_audit` reads `MultiProjectionGrain.cs`, counts the literal `_state.WriteStateAsync()` and `await WriteGrainStateGatedAsync()`, and asserts `1`/`7`. Renaming the wrapper, changing await style, moving a write into another partial/helper, or adding a semantically equivalent bypass can make this theater fail or pass for textual rather than architectural reasons. The production design should make bypass structurally impossible (inject/use the coordinator as the sole state writer boundary), and the test should exercise representative real call categories or inspect a stable seam—not source-string counts.

2. **The coordinator ordering/coherence test snapshots the wrong semantics.** The first write is parked inside `PersistAsync(state)`, then the test mutates the same shared `ModelState` for the second operation before releasing the first. Consequently the first commit already observes both `Checkpoint="cp"` and `FaultDescriptor="fault"` in either ordering. The test records only commit versions `[1,2]`; it does not record operation labels/start order, distinct immutable snapshots, ETags, or prove that checkpoint-first commits checkpoint-only before fault retry (and the inverse ordering correspondingly). `LastCommit` being coherent is therefore largely guaranteed by pre-release shared mutation, not by serialized production write semantics. Capture an immutable per-operation snapshot before queuing, record start/commit labels plus expected version/ETag, prove the second is ready while the first delegate is parked, and assert both intermediate and final commits in both orderings. The mutation bypass must fail these semantic assertions.

3. **Current-head GitHub verification is incomplete.** GitHub currently reports only `dcbTestsNet9=SUCCESS` and `dcbTestsNet10=SUCCESS` for `19a0f01b`; neither SonarCloud check is present. The delegation requires all four current-head checks, so routing cannot be released yet even after code repair.

Resolved/reverified:

- `ProbeStateWriteConcurrencyAsync` is absent from interface, implementation, repository search, and PR body; no wire test API remains.
- `InternalsVisibleTo` is limited to `Sekiban.Dcb.Orleans.Tests`; no signing configuration conflict found.
- `FirstQueryCatchUpGate` directly proves same-task sharing, synchronous success/failure handling, retry after failure, sticky success, and same original exception instance; it does not clear newer generations.
- Orleans fail-first/read-recovery reaches poison and returns projection fault, with no manual reset/persist.
- All current direct grain writes route through the coordinator by manual semantic audit; no nested coordinator acquisition found. Retry lifecycle, surrogate/schema, narrow clear, partial-state and prior parity remain intact.

Design alignment checked against the G14 packet, review-context, implementation, GitHub body, `intents/sekiban/design/silent-failure-hardening.md`, PR metadata, prior durable findings, and full diff. Managed worktree: `.intent-cli/worktrees/review-SEK-G14` (registered). Verification: `git diff --check` pass; focused gate/barrier suites pass 10/10 on net9 and 10/10 on net10; PR is MERGEABLE. The two test-proof defects and missing current-head Sonar checks prevent release.

tomohisa:
SEK-G14 strict rereview at head `b7333931ee2f0722fee0894ec595bf2760d8f8e0`: raw write ownership is improved, but mutable-state bypass and failure semantics remain blocking.

1. **Persisted mutation is still bypassable outside the gate.** `CoordinatedGrainStateStore.State` returns the mutable `MultiProjectionGrainState` instance, and its own documentation explicitly allows field updates. `MultiProjectionGrain` currently assigns `_stateStore.State.LastPosition` directly in three production paths (pending stream processing, normal batch processing, buffered batch processing). Those are persisted fields changed outside `ExecuteWriteAsync`; a later unrelated write can commit them. This violates the required structural rule even though the grain no longer holds raw `IPersistentState`. Expose read-only snapshots/accessors, eliminate mutable payload escape, and route every persisted-field mutation through `ExecuteWriteAsync`. Add an architecture test that detects mutation escape/bypass rather than only raw state ownership.

2. **Failed writes leak uncommitted mutations into live state.** `ExecuteWriteAsync` mutates `_state.State` and then awaits the provider write; on failure it releases the gate without rollback or re-read. Normal checkpoint/version/integrity fields can therefore appear advanced in memory and be consumed by short-circuit/integrity decisions or committed by a later unrelated operation even though their write failed. The fault descriptor intentionally remaining live is a special case, not justification for all fields. Use copy-on-write/rollback or explicit operation-specific retry semantics, and add deterministic tests showing a failed normal checkpoint/version write does not advance visible/persisted progress while fault retry remains live and safe.

3. **Fault-first inverse ordering is absent.** `CoordinatedGrainStateStoreTests` contains only checkpoint-first then fault. There is no deterministic fault-first case proving that once the fault is live/committed, a competing normal checkpoint is rejected or leaves checkpoint unchanged under the G14 no-progress-while-faulted rule. Add the inverse test and production guard; final state must not advance checkpoint beyond poison.

Resolved/reverified:

- `MultiProjectionGrain` retains no `IPersistentState` field; `CoordinatedGrainStateStore` is the sole raw writer by reflection/manual semantic audit.
- Public/wire API is unchanged and no test-only method remains.
- Current checkpoint-first harness parks the first provider write, proves the second mutation waits, clones at commit, observes max concurrency 1 and versions 1→2.
- First-query single-flight/recovery, durability retry/backoff/pin, surrogate/schema, narrow clear, partial-state and prior query/best-effort parity remain intact.
- Exact head reports all four GitHub checks SUCCESS and is MERGEABLE.

Design alignment checked against the G14 packet, review-context, implementation, GitHub body, `intents/sekiban/design/silent-failure-hardening.md`, PR metadata, prior durable findings, and full diff. Managed worktree: `.intent-cli/worktrees/review-SEK-G14` (registered). Verification: `git diff --check` pass; focused coordinator/barrier suites net9 pass 9/9. The three state-integrity gaps above prevent release.

tomohisa:
SEK-G14 strict-final rereview at head `37cdd2a7b9623dcc45ca3f3ca065775d254d749f`: two structural/coherence blockers remain.

1. **`Committed` is not a true immutable view; the mutable object still escapes and is downcastable.** `CoordinatedGrainStateStore.Committed` is declared as `IReadOnlyMultiProjectionGrainState`, but returns `_committed`, whose runtime type is the public mutable `MultiProjectionGrainState`. Any internal/friend consumer can cast it back and mutate persisted fields outside `ExecuteWriteAsync`. The architecture test checks only the declared property return type and therefore passes this exact bypass. Return a distinct immutable snapshot/record (all 18 values copied) whose runtime type is not `MultiProjectionGrainState`, and add a test asserting the returned object cannot be cast to the mutable state. Avoid adding an unnecessary public read-view interface if it is only an internal coordination detail.

2. **Operator reset clears the live actor before the copy-on-write commit succeeds.** `ResetPersistedStateForFullRebuildAsync` invokes `ClearFaultFromState(s)` inside the `ExecuteWriteAsync(OperatorReset, ...)` mutation delegate. That method immediately sets `_projectionFault = null` and calls `_host.ClearFaultForRebuild()` before `_state.WriteStateAsync()` is awaited. If the provider write fails, the coordinated store correctly rolls back the persisted candidate, but the live actor fault has already been cleared. The projection can then answer/apply despite the durable fault descriptor still existing—the reset is not coherent. Keep candidate mutation side-effect-free; clear live fault only after `ExecuteWriteAsync` returns `Committed`, and add a failing operator-reset write test proving both live and persisted fault remain until a successful contracted reset.

Resolved/reverified:

- The 18 persisted fields are cloned completely; ordinary failed checkpoint writes roll back provider buffer and committed state, and a later write does not carry failed position/version.
- Checkpoint-first serialization records immutable commit snapshots/version 1→2. Fault-first suppresses checkpoint without a write/version bump; live fault also suppresses checkpoint.
- The grain retains no raw `IPersistentState` or mutable state field. `_liveLastPosition` replaces the three ephemeral delivery-position assignments; persisted snapshot checkpoints continue to use `LastSortableUniqueId`, while status uses live position as before runtime-only `LastPosition` behavior.
- First-query gate/recovery, retry/backoff/pin, query/best-effort parity, surrogate/schema and narrow-clear intent otherwise remain intact.
- Exact head is MERGEABLE with all four checks SUCCESS.

Design alignment checked against the updated G14 design, packet, review-context, implementation, GitHub body, PR metadata, prior durable findings, and full diff. Managed worktree: `.intent-cli/worktrees/review-SEK-G14` (registered). Verification: `git diff --check` pass; focused coordinator/architecture/fault suites pass 18/18 on net9 and 18/18 on net10. The downcast bypass and reset side-effect ordering prevent release.

tomohisa:
Blocking implementation finding (SEK-G14 final rereview, head `4f01c16e`)

The immutable committed-state issue is resolved: `MultiProjectionGrainStateSnapshot` is a distinct sealed record containing copies of all 18 scalar/string fields, and the runtime downcast test is non-vacuous. The exact head also has all four required checks green.

The operator-reset acceptance evidence is still incomplete at the actual caller boundary. `MultiProjectionGrain.ResetPersistedStateForFullRebuildAsync` is private and `rg` finds no production or test caller; the new coverage in `CoordinatedGrainStateStoreTests.OnCommitted_runs_after_publish_on_success_and_never_on_failure` exercises only a generic store callback. It therefore does not prove the required end-to-end invariant on a faulted grain: a failed reset write retains both the persisted descriptor and live actor fault, state/scalar/list queries remain failed, and a later successful deterministic reset clears both without exposing healthy queries before rebuild state is ready.

Please add the minimal caller-level/internal-friend or real TestCluster test around `ResetPersistedStateForFullRebuildAsync` with an injected fail-first provider, covering failed reset → all three query surfaces remain faulted → successful reset/rebuild → coherent recovery. The test must drive the actual grain reset path, not reproduce its callback logic at store level. If this method is intentionally unreachable, either wire the contracted operator/deterministic rebuild path and test it or remove/adjust the unsupported “full operator rebuild clears it” product claim through the authoritative design route.

Also keep the post-commit continuation non-throwing or explicitly handle callback failure: durable state is published before `onCommitted`, so an exception there cannot safely be reported/retried as though the provider commit failed.

This is an implementation-finding tied to the SEK-G14 narrow-clear/durable fault contract and the final-rereview requirement for caller-level reset failure/success coherence. No merge is authorized.

tomohisa:
Blocking implementation findings (SEK-G14 strict-final rereview, head `bbb169cf`)

The new admin surface is real, Orleans-serializable, absent from `ISekibanExecutor`, and the focused suites pass 16/16 on both net9 and net10. The option-a delta nevertheless misses three required invariants.

1. **The complete token is not validated atomically against persisted authority.** `ResetProjectionFaultAsync` checks `request.ProjectorName` against `GetProjectorName()` before entering the store gate, while `FaultTokenMatchesCommitted` inside `ExecuteGuardedWriteAsync` checks only event id and position. The approved design and delegation require projector name + event id + position to be one precondition against the current persisted descriptor inside the single-writer gate. Move all three comparisons into the guarded predicate (including the committed projector identity), and add isolated missing/wrong projector, id, and position tests proving zero provider write and zero derived/live mutation.

2. **This is not a full rebuild when an external projection snapshot exists.** `ResetDerivedStateForFullRebuild` clears only `MultiProjectionGrainState`. On the fresh activation, initialization still calls `IMultiProjectionStateStore.GetLatestForVersionAsync(projectorName, projectorVersion)` and restores the existing Postgres/Cosmos/in-memory external snapshot for the same version. No reset code deletes, invalidates, or bypasses that snapshot. The new TestCluster starts with a poison as the first event, so no pre-poison external snapshot exists and the tests cannot detect this. This contradicts the approved “projection persistent state full-rebuild reset / rebuild from the beginning” contract. Add a safe, failure-coherent external-derived-state invalidation/bypass strategy and a real TestCluster regression that first persists a pre-poison snapshot, resets, and proves replay from the beginning with exact state values.

3. **The no-early-healthy and failed-reset query proofs are vacuous/incomplete.** The recovery tests call the pre-existing test/admin `RequestDeactivationAsync()` after `ResetProjectionFaultAsync`, so they do not prove that the product's own `DeactivateOnIdle()` is sufficient. They also assert only `IsSuccess`, not the recovered count/list/snapshot value, so an empty/current false success would pass. The fail-first test checks only `GetSnapshotJsonAsync` on failure and only snapshot success after retry, not state/scalar/list. Remove the manual deactivation assist, drive the immediate first real state/scalar/list calls after reset, assert the exact replayed value/position, and prove unchanged poison re-fault persistence plus all three failed-reset surfaces.

These are implementation findings tied directly to design commit `926f02b` §2.5.1 and the final delegation's token-authority/full-rebuild/no-early-success requirements. No authoritative event deletion is requested or authorized; the repair must remain derived-state-only and failure coherent. No merge is authorized.

tomohisa:
Blocking implementation findings (SEK-G14 round-11 rereview, head `9f1e256b`)

The round resolves several prior gaps: the three token values are now read in the guarded predicate, the reset recreates the host and re-arms the first-query barrier without manual deactivation, exact scalar/list/position assertions were added, and the focused suites pass 19/19 on net9 and net10. Three correctness gaps remain.

1. **Missing token values are not deterministically rejected.** `FaultTokenMatchesCommitted` does equality checks but has no non-empty validation. In particular, `(committed.FaultPosition ?? "") == (request.FaultPosition ?? "")` accepts a missing request position when the persisted descriptor position is null/empty; empty projector/id values can likewise become valid if legacy/corrupt state contains the same value. The delegation explicitly requires missing fields to reject. Require non-null/non-empty projector, event id, and position on both the request and persisted descriptor inside the same guarded predicate, and add separate null/empty tests that assert no grain write, no external delete, no checkpoint change, and no live clear.

2. **External snapshot invalidation is not coherent with concurrent snapshot persistence.** `beforeWrite: InvalidateExternalDerivedSnapshotAsync` runs under `CoordinatedGrainStateStore`'s gate, but production external snapshot upserts in `PersistStateAsync`/streaming persistence occur before their later grain-state write and are not protected by that gate. With interleavable timers, an already-started persist can upsert the old derived snapshot after reset's `DeleteAsync` but before/after the descriptor-clear commit; the reset then succeeds while a fresh activation can still restore pre-poison state. A process crash between external delete and grain commit also leaves a failed reset that did mutate derived state, contrary to the current “failed reset changes nothing” docs. Introduce a durable reset epoch/tombstone or a coordinator covering both external writes and reset invalidation so stale upserts cannot become restorable, and test deterministic overlap plus both partial-failure orderings.

3. **The reported background/refresh race is a remaining production durability path, not merely fixture setup.** `RefreshAsync` invokes `ProcessSingleCatchUpBatch()` in a loop and calls `CaptureAndPersistProjectionFaultIfAnyAsync()` only after the loop. If deserialize/fold throws, control exits before that final call. Unlike `ProcessCatchUpBatchAsync` (which routes to `HandleCatchUpBatchFailureAsync`), this in-call authoritative path can leave only `host.CurrentFault` live with no persisted descriptor/retry scheduled. Fix with a catch/finally that awaits capture before propagating/handling, and add a deterministic real-grain test proving a Refresh-triggered poison persists the descriptor without relying on background catch-up, then survives reactivation.

The new external-snapshot test is also not the required persisted pre-poison fixture: it injects four arbitrary bytes directly into `InMemoryMultiProjectionStateStore` after the grain has already faulted, never proves that a valid product-written snapshot could restore, and never combines that snapshot with exact fixed-poison replay. Replace or supplement it with a genuine product persistence → poison → reset sequence and exact rebuild result.

These are implementation findings tied to design commit `926f02b` §2.5.1 and the round-11 durability/atomicity requirements. No authoritative event deletion is authorized. No merge is authorized.

tomohisa:
Blocking implementation finding (SEK-G14 round-12 strict-final rereview, head `d1658e5a`)

The round correctly fixes RefreshAsync with awaited fault capture in `finally`, adds explicit non-empty token checks, genuine product snapshot/rebuild coverage, and both documented cross-store partial failures. Focused tests pass 25/25 on net9 and net10. However, decision A's load-bearing concurrency invariants are not true in the production implementation.

**Decision A assumes a non-reentrant activation with no external upsert outstanding across another turn; this grain explicitly permits interleaving.** `EnsureInitializedAsync` registers persist, fallback and batch timers with `GrainTimerCreationOptions.Interleave = true`; the catch-up and immediate-flush timers are also interleavable. An awaited `UpsertFromStreamAsync` therefore can remain outstanding while another interleavable turn faults or resets the same activation. “Awaited” does not imply “no cross-turn outstanding work” when the timer opts into interleaving.

The required fault gate is also incomplete:

- `PersistStateAsync` detects `host.CurrentFault`, persists the descriptor, but then continues through checkpoint capture and can reach normal or streaming external `UpsertFromStreamAsync`; `TryShortCircuitPersist` has no fault predicate. The new `NoExternalUpsertHappensWhileFaulted` test passes only because that fixture happens to hit the unchanged-checkpoint short circuit.
- `OverwritePersistedStateVersionAsync` is a third grain-owned `UpsertFromStreamAsync` path and has no live/persisted-fault rejection at all.
- All three upserts are awaited, but none is covered by a structural operation gate shared with reset's external delete. A healthy persist can start its upsert, yield, then an interleavable catch-up turn can establish the fault and a reset can delete; the old upsert may subsequently complete and recreate the snapshot after deletion.

This directly violates updated design `f340eac` §2.5.2 invariants 1–3, so the documented claim “no in-flight upsert can race the delete” is currently false. Do not add the superseded epoch/tombstone. Instead satisfy decision A structurally: remove the relevant interleaving or introduce one activation-local async coordinator covering every external upsert plus reset invalidation, add explicit fault rejection before all normal/streaming/version-rewrite upserts, and add a deterministic parked-upsert/reset test proving no overlap or stale recreation. The test must fail if any one of the three upsert paths bypasses the protection.

Evidence gap: the token code rejects null correctly via `IsNullOrEmpty`, but the required matrix currently tests empty strings only. Add explicit null rows for projector/id/position and retain zero delete/write/live-clear assertions.

Refresh durability, external-first partial-failure semantics, genuine snapshot recovery, exact query results, operator-only boundary and earlier G14 invariants otherwise remain accepted. No merge is authorized.

tomohisa:
SEK-G14 round 13 strict rereview — request update

Head reviewed: `2a90749cc9c741da9acbf18e49059675212c2e67`

Design alignment checked against the SEK-G14 packet/review-context, `intents/sekiban/design/silent-failure-hardening.md` at design commit `f340eac` (§2.5.2 decision A), the intent-tree reference, and the PR diff/docs. The new `ExternalStoreCoordinator` correctly serializes its own upsert/delete delegates, but two current-PR acceptance gaps remain:

1. **Implementation finding — external-store mutation still bypasses the coordinator.** `IMultiProjectionGrain.DeleteExternalStateAsync` remains a public/wire “testing aid”, and `MultiProjectionGrain.DeleteExternalStateAsync` directly calls `_multiProjectionStateStore.DeleteAsync(...)` instead of `ExternalStoreCoordinator`. This contradicts the round-13 requirement that *every* `IMultiProjectionStateStore` mutation, including testing aids, use the activation-local coordinator. With `[Reentrant]`/interleaving calls, this delete can overlap an upsert or reset invalidation and bypass the stale-snapshot ordering invariant. Route this mutation through the coordinator (or remove the production testing aid without a compatibility break) and add a non-vacuous path test proving it cannot overlap/reorder with a parked upsert.

2. **Implementation finding — a fault-gated upsert skip is treated as a successful save.** `ExternalStoreCoordinator.UpsertAsync` deliberately returns successful `false` when faulted, but all callers test only `IsSuccess`. Normal and streaming persistence therefore set `externalStoreSaved=true`; more critically, `OverwritePersistedStateVersionAsync` sets `updated=true` and executes a `MetadataMaintenance` state write, which is not fault-suppressed by `CoordinatedGrainStateStore`. Thus a faulted version rewrite can advance the persisted projector version even though the external rewrite was rejected. The existing `VersionRewrite_IssuesNoExternalUpsertWhileFaulted` test asserts only absence of the external v2 snapshot, so it passes while missing this persisted-state mutation. Require `IsSuccess && GetValue()` (or an explicit outcome type), and add path-level assertions that normal, actual streaming, and version rewrite cause zero external write and zero grain-state/checkpoint/version mutation for either live or committed fault.

Evidence:

- Exact GitHub head and mergeability reverified; all four checks are SUCCESS (`dcbTestsNet9`, `dcbTestsNet10`, `SonarCloud`, `SonarCloud Code Analysis`).
- Focused `ExternalStoreCoordinatorTests|ProjectionFaultResetOrleansTests`: 20/20 passed on net9.0 and 20/20 passed on net10.0. These passes are necessary but insufficient because the two paths above are not asserted.
- `git diff --check` passed.
- No merge performed.

Please repair both paths and add mutation-relevant coverage before rereview.

tomohisa:
SEK-G14 d81 strict-final rereview — one remaining evidence blocker

Head reviewed: `d81ade7dd6ee015754e1d9a601c54aef29ceec81`

The two round-13 code findings are resolved. I verified that all three grain upsert paths call `UpsertExternalStateCoordinatedAsync`; reset invalidation and public `DeleteExternalStateAsync` call `ExternalStoreCoordinator.InvalidateAsync`; the only direct grain-level `DeleteAsync`/`UpsertFromStreamAsync` calls are inside those coordinator delegates. Fault rejection is now `ResultBox.Error<ExternalPersistenceBlockedByFaultException>`, normal/streaming keep `externalStoreSaved=false`, and the non-vacuous version-rewrite test proves no upsert, no grain write/version mutation, no v2 snapshot, retained v1 snapshot, and retained fault. Both coordinator orderings are deterministically tested. Lock ordering is acyclic: upserts release the external gate before any later grain-state write; reset holds the grain-state gate while awaiting the external gate, so an in-flight upsert can finish/release rather than waiting back on the grain gate.

One tests-as-contract finding remains:

- **Implementation finding — persisted-token null/empty rows required by the final verification are absent.** `MissingOrEmptyTokenField_IsRejected_NoEffect` contains the six required request-side rows (null and empty for projector/event-id/position), but there is no corresponding setup/test for null and empty values in each of the three **persisted descriptor** fields. The production guard does explicitly check both sides, but the strict-final delegation requires explicit rows for request *and persisted* fields, with zero external delete, zero provider write/ETag mutation, and no live-fault clear. A malformed/backward persisted descriptor is exactly where this branch matters; request-only rows cannot kill a mutation that drops one of the persisted-side presence checks.

Please add the minimal six persisted-side rows using a genuinely persisted malformed descriptor, and assert the reset is rejected with unchanged delete count/snapshot, unchanged grain write count/ETag, and continued state/scalar/list rejection. Include mutation relevance for removing each persisted-field nonempty guard.

Evidence:

- Design alignment checked: SEK-G14 packet, implementation, review-context, github-body, `intents/sekiban/design/silent-failure-hardening.md` at `f340eac` §2.5.2 decision A, intent-tree reference, full delta, EN/JA docs, PR body.
- Exact head and MERGEABLE reverified; all four checks SUCCESS.
- Focused coordinator/reset/actual-streaming suite: 27/27 net9.0 and 27/27 net10.0 passed.
- `git diff --check` passed.
- No merge performed.

tomohisa:
SEK-G14 final rereview — completed, no blocking findings

Approved evidence head: `5b0dcc6bb16ce3a3285ffb2635de3c6b6781ee45`.

Design alignment checked against the SEK-G14 packet/review-context/implementation/github-body, `intents/sekiban/design/silent-failure-hardening.md` at design commit `f340eac` §2.5.2 decision A, the linked intent-tree reference, full PR diff, EN/JA docs, and PR metadata. The accepted A+B evidence substitution is satisfied without adding an artificial production seam:

- Delta from `d81ade7d` is one test file only; `FaultTokenMatchesCommitted` and all production code are unchanged.
- Four genuinely persisted/reactivated ProjectorName/FaultPosition null-or-empty cases reject reset with unchanged provider write/version, external upsert/snapshot, live fault, and state/scalar/list rejection.
- Persisted FaultEventId null/empty is truthfully demonstrated: benign poison restores no live fault and becomes healthy; active poison is synchronously re-folded by the first-query barrier, persists a new valid descriptor, remains query-failing until the valid token resets it.
- Three both-empty false-match rows reject projector/event-id/position empty identities with zero external/grain mutation. The PR correctly claims mutation relevance for removing each paired request+committed presence guard, not for an individually redundant clause.
- Prior coordinator, explicit fault-block error, decision-A compensation, Refresh durability, exact rebuild, operator boundary, parity, and no-bypass conclusions remain unchanged.

Verification:

- Exact GitHub head and MERGEABLE state reverified; all four checks SUCCESS (`dcbTestsNet9`, `dcbTestsNet10`, `SonarCloud`, `SonarCloud Code Analysis`).
- Focused coordinator/reset/actual-streaming suite: 38/38 passed on net9.0 and 38/38 passed on net10.0.
- `git diff --check` passed.
- PR retains `Closes #1078`; no out-of-scope merge/release/migration action performed.

Tests are necessary but not sufficient; approval is based on the packet/design mapping and the non-artificial production-path evidence above. Review is released for the design-owned admin merge fallback. I did not merge.

Issue #1080, opened 2026-07-18
Sekiban.Dcb: IEventStore exposes no optimistic-concurrency / expected-version / unique-append — multi-writer CAS and exactly-once are not expressible


tomohisa:
## Summary

The Sekiban.Dcb event store abstraction (`IEventStore`) exposes **no optimistic-concurrency / expected-version / unique-append** primitive, and the command layer (`ICommandContext`) cannot supply a storage-rejectable caller `EventId`. As a result, an application **cannot express "append this event only if the stream/tag has not advanced"** (compare-and-set) or "append only if this logical key does not already exist" (unique insert). Concurrent writers can both append conflicting/duplicate events, and the application has no way to have the store reject one of them at write time.

This forces application-level convergence patterns (e.g. allow duplicate writes and pick a deterministic winner by durable `SortableUniqueId` ordering) wherever a true exactly-once / mutual-exclusion guarantee is needed.

## Where it bites (concrete)

Building a startup migration in a multi-host deployment, we needed a durable compare-and-set so that of N racing hosts, exactly one performs a one-time migration and writes a single completion marker. We found:

- `IEventStore` write APIs (`WriteEventsAsync` / `WriteSerializableEventsAsync`) accept events unconditionally; there is no `expectedVersion` / `ifMatch` / unique-key parameter and no storage-level rejection path.
- `ICommandContext` cannot pass a caller-chosen `EventId` that the store would enforce as unique, so "insert-if-absent" is not expressible either.
- The in-process executor path uses process-local reservations, which do **not** coordinate across hosts sharing the same durable store.

We resolved it at the application layer with a **first-durable-event-wins** design (allow duplicate idempotent marker writes; authority = the first `SortableUniqueId` in durable order; all hosts re-read and converge on the same winner). That is correct for our idempotent case, but it only works because our operation is idempotent — it is not a general substitute for storage CAS.

## Why worth adding

Multi-writer coordination (single-leader election, exactly-once side effects, unique aggregate creation, idempotency keys) is a common need. Without a storage-level optimistic-concurrency or unique-append primitive, every such case must be hand-rolled with idempotency + total-order resolution, which is easy to get subtly wrong (we spent multiple review rounds proving convergence).

## Suggested direction (either)

- **Optimistic concurrency on append**: an `expectedVersion` (per tag/stream) on the write API; the store rejects the append if the current version differs, surfaced as a typed concurrency error the command layer can catch.
[...]

PR #1082, opened 2026-07-18
SEK-G15: conditional-append contract — optional interface, capability descriptor, typed errors, in-memory reference

### LINKED ISSUE #1081: SEK-G15 dcb: conditional-append contract — optional interface, capability descriptor, typed errors, in-memory reference
# SEK-G15 dcb: conditional-append contract — optional interface, capability descriptor, typed errors, in-memory reference

## Goal

Make storage-level conditional append expressible without touching `IEventStore`:
a separate optional interface for unique-append (idempotency-key) semantics, a
runtime-resolved write-condition capability descriptor, typed concurrency errors,
opt-in executor/serialized seams, and a deterministic in-memory reference
implementation — proven fully additive.

## Why This Slice Exists Now

Issue #1080 (operator-filed, from Sekiban Cloud): an N-host one-time migration
needed "exactly one host performs this" but nothing in the stack can express it —
tag reservations only cover commands through one Orleans cluster/executor, and
`IEventStore` writes are unconditional with server-generated EventIds. The team
hand-rolled first-durable-event-wins convergence, which only works for idempotent
operations.

## Current Observed State

- `IEventStore` exposes unconditional `WriteSerializableEventsAsync` only;
  EventId/SortableUniqueId are generated server-side immediately before write;
  `ICommandContext.AppendEvent` accepts payload/tags only.
- Tag reservations are cluster-wide only within one Orleans cluster/executor;
  process-local for the in-memory executor; the store never validates them.
- Downstream implements `IEventStore` directly (SekibanAsAService production
  `ServicePartitionedEventStore`; WasmRuntime embedded/test implementors) — adding
  abstract members would break them.
- Providers already have collision material (PG/SQLite PK, Cosmos 409, Dynamo
  attribute_not_exists) but no caller-facing contract; provider work is the NEXT
  slice (SEK-G16).

- G10 capability-descriptor pattern (runtime-resolved, decorator/factory
  propagation) is established and must be reused for write-condition capability.
- G9/G14 exception contract: typed intentional failures; original provider errors
  preserved; WithResult=ResultBox.Error, WithoutResult=guarded throw.
- Frozen compatibility conditions (operator gate): separate optional interfaces,
  opt-in only, fail-closed unsupported, existing IEventStore/DTO/signatures
  untouched.

## Target Repo / Path / Part

Target part: optional conditional-append interface, capability descriptor, typed errors, executor/serialized opt-in seams, in-memory reference, compatibility proof, docs.

## In Scope

- **Optional interface** (e.g. `IConditionalEventStore`): **single-event**
  conditional append with a caller-supplied idempotency key (batch claims are a
  future capability kind, not this contract).
- **Outcome state machine** (one observable contract): `Appended` /
  `AlreadyCommittedSameOperation` (carries the durable receipt: winner EventId,
  SortableUniqueId, fingerprint) / `KeyReuseConflict` (identity exists, operation
  fingerprint differs) / `ConditionNotSupported`. Provider exceptions, when
  present, are preserved as diagnostic causes; a duplicate discovered by read has
  no provider exception and none is fabricated.
- **Operation fingerprint + durable receipt**: canonical fingerprint (derivation
  version + domain separator + serviceId + normalized key + event type +
  canonical payload + ordered tags) persisted with the claim. Identity match +
  fingerprint match = same operation; fingerprint mismatch = KeyReuseConflict.
  Raw keys never logged/persisted unnecessarily; key normalization, UTF-8
  encoding/domain separator, max size, and derivation version specified;
  ServiceId is part of the identity (no cross-service claims).
- **Write-condition capability descriptor** (G10 pattern): distinguishes
  capability kinds (single-event unique now; batch / expected-position as future
  kinds); actually-resolved store evaluated; decorators/factories/composites
  propagate (Hybrid = hot store; composite = all-underlying-support only); never
  type-name-based.
- **Fail-closed**: requesting an unsupported capability kind fails with the typed
  capability error BEFORE handler invocation, EventId allocation, serialization,
  or any store call (ordering pinned by test); no silent degradation
  (architecture test: no default pass-through implementation).
- **Executor seams**: existing `ExecuteAsync` overloads unchanged; new opt-in
  options/overload on both facades; `ICommandContext` unchanged; NEW versioned
  serialized-commit DTO/method for the WASM path (existing positional DTO
  untouched).
- **InMemory reference implementation** (testing stack per the misuse-prevention
  taxonomy): deterministic semantics for the full outcome machine — first wins
  with receipt; same-operation retry → AlreadyCommittedSameOperation (same
  receipt); different-operation same-key → KeyReuseConflict.
- **Compatibility proof tests (strengthened)**: frozen public API baselines for
  IEventStore/ICommandContext/positional DTO (baseline test fails on change);
  separate consumer fixture compiled against new assemblies; legacy paths never
  capability-cast and serialized bytes/events preserved; existing suites pass
  unmodified; fail-closed ordering proof (before handler / EventId / serialization
  / store).
- **Docs (EN + JA)**: contract, capability discovery, typed errors, and the
  boundary that storage guarantees one durable claim (external side-effect
  exactly-once still needs outbox/idempotency).

## Out Of Scope

- Production provider implementations (SEK-G16).
- Expected-position/CAS semantics (P2 — Cosmos variant needs separate approval).
- Changes to IEventStore, ICommandContext, existing DTOs, or default behavior.
- Orleans reservation-layer changes.

The child PR must add the separate optional conditional-append interface with
typed outcomes, the runtime-resolved write-condition capability descriptor with
decorator/composite propagation, fail-closed rejection before handler/write on
unsupported stores, opt-in executor overloads and a new versioned serialized
DTO/method (all existing signatures untouched), a deterministic in-memory
reference implementation placed per the testing taxonomy, compatibility-proof
tests (external implementor unchanged, unconditional path unchanged), and EN + JA
docs — additive only, zero default behavior change.

## Acceptance Criteria

- Separate optional interface defining SINGLE-EVENT conditional append; IEventStore and all existing signatures untouched.
- One outcome state machine: Appended / AlreadyCommittedSameOperation (durable receipt: winner EventId + SortableUniqueId + fingerprint) / KeyReuseConflict / ConditionNotSupported; executor and serialized surfaces' exposure defined; provider exceptions preserved as causes only when present (no fabricated InnerException on read-discovered duplicates).
- Canonical operation fingerprint (derivation version, domain separator, serviceId, normalized key, event type, canonical payload, ordered tags) persisted with the claim; identity+fingerprint classification rules; raw keys never logged/persisted unnecessarily; key normalization/encoding/max-size/derivation-version specified; ServiceId in identity.
- Capability descriptor distinguishes capability kinds (single-event unique / future batch / expected-position); runtime-resolved; decorator/factory/composite propagation; never type-name detection.
- Fail-closed before handler invocation, EventId allocation, serialization, and store calls (ordering pinned); architecture test proves no silent pass-through exists.
- Opt-in executor overloads on both facades; ICommandContext unchanged; new versioned serialized DTO (existing DTO untouched).
- InMemory reference implements the full outcome machine deterministically; end-to-end contract tests incl. same-operation retry and key-reuse conflict.
- Compatibility proof: frozen API baselines (IEventStore/ICommandContext/positional DTO); separate consumer fixture; legacy paths never capability-cast; serialized bytes/events preserved; existing suites unmodified.
- Docs EN + JA (contract, outcome machine, fingerprint/receipt, capability kinds, one-durable-claim boundary, provider support arriving in SEK-G16).
- Additive-only public API; zero default behavior change.

## Verification

## Related Links

- https://github.com/J-Tech-Japan/Sekiban/issues/1080
- Host design record: intents/sekiban/design/storage-concurrency-primitives.md

## Knowledge Maintenance

- Intent placement: host design doc storage-concurrency-primitives.md (P0).
- ADR candidate: none (decision recorded host-side).
- Diagram candidate: none.
- Docs update: 11_storage_providers.md EN/JA — in scope.
- Closeout writeback expected: yes — seam friction findings feeding SEK-G16/P3.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1081

## What this adds

A storage-level **conditional (unique-key) append** contract so callers can express "exactly one host performs this" (issue #1080's N-host one-time migration), which nothing in the stack could express before. **Strictly additive**: `IEventStore`, `ICommandContext`, and the existing serialized DTOs are untouched; there is zero default behavior change; unsupported stores **fail closed**.

## The contract (`Sekiban.Dcb.Core`)

- **Optional interface** `IConditionalEventStore.AppendIfUniqueAsync(ConditionalAppendRequest, …)` — single-event append under a caller-supplied idempotency key, feature-detected with `is` exactly like `IStreamingSerializableEventStore`. Never a member of `IEventStore`, so external implementors keep compiling.
- **One outcome machine**: `Appended` / `AlreadyCommittedSameOperation` (both carry a durable **receipt** — winner `EventId` + `SortableUniqueId` + operation fingerprint) / `KeyReuseConflict` / `ConditionNotSupported`.
- **Typed errors** per the G9/G14 contract: `KeyReuseConflictException`, `ConditionNotSupportedException` (WithResult → `ResultBox.Error`; WithoutResult → guarded throw). A provider exception is preserved as the inner cause **only when one actually occurred**; a conflict discovered by read has none and none is fabricated.
- **Canonical operation fingerprint** (`OperationFingerprint`): length-prefixed, domain-separated SHA-256 over **derivation version + canonicalization version** + domain separator + ServiceId + normalized key (NFC, trimmed, ≤ 512 UTF-8 bytes) + **authoritative event-type identity** (registered CLR `FullName`) + **canonical payload** (deserialize → re-serialize → recursively key-sorted, whitespace-free JSON; array order preserved; numbers/strings as emitted) + ordered tags. Currently **derivation v2 / canonicalization v1**, pinned by golden vectors with literal digests (any drift fails a test). Server-generated ids are **excluded** so a genuine retry is still recognized; ServiceId is part of identity (no cross-service claims); raw keys never logged/returned.
- **Conservative supported-shape boundary (enforced, not documented)**: the payload's effective `JsonTypeInfo` graph is validated (cycle-guarded, no mutation, reflection or source-gen) inside the store's `AppendIfUniqueAsync`. Only a JSON-object root, ordered collections, and allowlisted primitive leaves whose **effective converter is authoritatively built-in** are admitted; a custom converter at the **options, type, or property level — even on an otherwise-allowed leaf (e.g. a custom `string`/enum converter)** — plus converter-owned/`None`/set/dictionary/unknown-leaf shapes, is rejected with a typed `OperationCanonicalizationException`. **Ordering (truthful):** this gate runs AFTER the handler has produced the event and after the executor serialized the candidate — so the handler and the payload's converter do run — but BEFORE the store computes any fingerprint or performs any write. So no fingerprint is ever produced for an unsupported shape and no durable claim is made, which is what makes two differing fingerprints for one key impossible. This is proven on the reflection path (an alternating options-level `string` converter — the test captures the two emitted candidates and asserts they genuinely differ) AND on a genuine source-gen/AOT path (a `[JsonConverter]` on an enum leaf, rejected through the production `ComputeCanonical`). The in-memory reference's `WriteCalls` counts only SUCCESSFUL durable writes (incremented after the base append commits; a failed base write is not counted, pinned by a test), so the absence of a durable claim is asserted directly, not inferred from empty state.
- **Secret-safe canonicalization failure**: a converter/deserializer exception (which can embed the raw payload/key) is **discarded**, never chained into the `ResultBox.Error` graph; the typed exception carries only sanitized metadata (the registered type name) and has no inner-exception constructor.

## Capability descriptor (G10 pattern, runtime-resolved, fail-closed)

`WriteConditionKind` (`SingleEventUniqueKey` now; `BatchUniqueKey` / `ExpectedPosition` reserved) + `WriteConditionCapabilityDescriptor` + `IWriteConditionCapabilityProvider` + a resolver switch. Resolved from the **live** store the container built — never a type-name check. `HybridEventStore` reports what its hot store enforces (writes land there; never upgrades on its own authority) and forwards the append; composites intersect. Requesting an unsupported kind raises `ConditionNotSupportedException` **before** the handler runs, before EventId allocation, serialization, or any store call (ordering pinned by a test). An architecture test proves there is **no default pass-through** — any `IConditionalEventStore` must also declare `IWriteConditionCapabilityProvider`.

## Executor seams (opt-in only)

- New `ExecuteAsync(command, handler, CommandExecutionOptions, …)` overloads on both facades via a separate `IConditionalCommandExecutor`; the existing `ExecuteAsync` overloads and `ICommandContext` are unchanged.
- New **versioned** `SerializedConditionalCommitRequest`/`Result` + `ISerializedConditionalSekibanDcbExecutor` for the WASM boundary; the existing positional `SerializedCommitRequest` is untouched.

## Reference implementation + tests

- `Sekiban.Dcb.Testing.InMemoryConditionalEventStore` (in `Sekiban.Dcb.Core.Testing`, never referenced from a runtime project) implements the full outcome machine deterministically, per-ServiceId.
- Tests: fingerprint determinism/sensitivity/normalization; outcome machine (Appended → AlreadyCommittedSameOperation same-receipt no-second-write → KeyReuseConflict no-inner-exception); two distinct fail-closed gates with truthful ordering — the **capability preflight** rejects an unsupported store BEFORE the handler runs (handler not invoked, nothing written), while the **shape/canonicalization boundary** runs after the handler but before any fingerprint/write (asserted via `WriteCalls` == 0, no event, no receipt); capability resolution + Hybrid propagation; **frozen API baselines** for `IEventStore` / `ICoreCommandContext` / positional `SerializedCommitRequest` (fail on change); external consumer fixture; serialized conditional path; WithoutResult guarded-throw parity.
- **Docs EN + JA** (`11_storage_providers.md`): contract, capability discovery, typed errors, fingerprint/receipt, the one-durable-claim boundary, and that provider implementations arrive in **SEK-G16**.

## Out of scope (per packet)

Production provider implementations (SEK-G16); expected-position/CAS (later slice). This slice is the contract + in-memory reference + compatibility proof only.

## Evidence

- `Sekiban.slnx` builds clean (0 errors). `git diff --check` clean.
- WithResult **731** (was 707, +24), WithoutResult **81** (+3), ColdEvents **90** (+2) — green on **net9 and net10**; existing suites unchanged.
- Additive public API only; no signature removals or modifications (frozen-baseline tests enforce this).

> Note on line endings: `CoreGeneralSekibanExecutor.cs` is stored **CRLF** in `main` (it predates the repo's LF convention). This PR **preserves that CRLF blob** — the conditional support is added as CRLF lines, so the raw diff for that file is **additions-only (~221 add / 0 removals)** with no whole-file churn. A narrowly-scoped `.gitattributes` entry (`… CoreGeneralSekibanExecutor.cs whitespace=cr-at-eol`) keeps `git diff --check` clean while the diff stays textual; no other file is affected.


tomohisa:
SEK-G15 strict review — request update

Head reviewed: `c6f4574e978fcd33e1380ea7a45ab2afa91c429b`

Design alignment checked against the SEK-G15 packet, implementation, review-context, GitHub body, `intents/sekiban/design/storage-concurrency-primitives.md`, intent-tree reference, full diff, and EN/JA docs. The optional-interface direction is correct, but the current implementation does not yet satisfy the frozen P0 contract:

1. **Implementation finding — fingerprint is serialized-byte identity, not the required canonical logical append.** `OperationFingerprint.Compute` hashes `SerializableEvent.Payload` verbatim and the supplied `EventPayloadName`. It defines no canonical JSON/payload normalization (object-property order, whitespace, equivalent numeric/string encodings, dictionary ordering, serializer/runtime differences) and no authoritative exact event-type identity validation. The current tests only pass identical bytes (`"payload"u8`) and change one raw component. Consequently a semantically identical WASM/provider/runtime retry can conflict solely because its JSON bytes differ, while a spoofed/simple-name type identity can be classified incorrectly. Define and version the provider-independent canonical payload/type/tag contract, use it in both typed and serialized seams, and add cross-shape/runtime vectors (equivalent JSON property order/whitespace, dictionary/tag ordering/duplicates, exact registered type identity) with frozen expected hashes.

2. **Implementation finding — zero-event conditional commands silently succeed.** `CoreGeneralSekibanExecutor.ExecuteAsync` returns the legacy empty `ExecutionResult` when `collectedEvents.Count == 0` before entering `ExecuteConditionalAppendAsync`. The documented/packet contract says conditional append requires exactly one event; zero and multi-event shapes must fail closed. Move the exact-one validation before the zero-event legacy return when conditional options are present, and test zero and >1 on both facades with no store call.

3. **Implementation finding — versioned WASM request ignores its version.** `CommitSerializableEventConditionallyAsync` never validates `SerializedConditionalCommitRequest.Version`; any future/invalid version executes as v1. Validate `CurrentVersion` before EventId allocation, serialization, publisher/store calls and add unsupported-version/zero-version fail-closed tests. Also validate the result/version contract at the serialized boundary.

4. **Implementation finding — Hybrid can advertise a lying hot-store capability and fail only after handler execution.** `HybridEventStore.DescribeWriteConditions` forwards the hot descriptor even when `_hotStore` does not implement `IConditionalEventStore`; Hybrid itself implements the interface, so the executor's early descriptor+cast check passes, invokes the handler/allocates/serializes, and only `Hybrid.AppendIfUniqueAsync` rejects. This violates fail-before-handler for the actual selected store. Hybrid must report support only when the hot instance both advertises the kind and implements the matching interface; add a descriptor-only hot fake proving handler/EventId/serialization/store counters remain zero.

5. **Implementation finding — compatibility proof is not signature-complete or genuinely external.** The `IEventStore` baseline compares method names only, `ICoreCommandContext` only counts `AppendEvent`, and the “external” store is a nested class compiled in the same test project against current sources. These tests miss return/parameter/generic/default/nullability/signature changes and are not an independently compiled downstream fixture. Add frozen full API signatures (including inherited members/constructors/positional DTO shape) and a separate consumer project/fixture referencing the built packages/assemblies through legacy surfaces. Prove legacy executor and serialized bytes/events unchanged and no capability cast.

6. **Implementation finding — normalize the obscuring Core executor EOL churn.** Raw diff is `935+ / 746-`, while `--ignore-space-at-eol` is only `189+ / 0-`; the semantic change is additive but the checked-in CRLF normalization obscures it. Restore the file's base EOL convention so the raw diff contains only the conditional additions.

Evidence:

- Focused conditional suite: 24/24 passed on net9.0 and 24/24 on net10.0; tests are necessary but insufficient because the missing cases above are unrepresented.
- `git diff --check` passed.
- Exact head and MERGEABLE reverified. Required check conclusions are recorded separately in the review result.
- No G16 provider or expected-position implementation found; no merge performed.

Please repair these contract gaps and add mutation-relevant evidence before rereview.

tomohisa:
SEK-G15 strict rereview — request update

Head reviewed: `90f775cbe6e216bb2b863fcd4e37d67124fa3b3b`

The prior findings for zero/multi-event fail-closed behavior, serialized-version ordering, Hybrid runtime-store capability resolution, compatibility coverage, and CRLF-only churn are resolved. The conditional focused suite passes on net9.0 and net10.0 (29/29 each), and all four current-head checks are successful. Two implementation findings still block the packet contract:

1. **The public canonical fingerprint contract is not frozen by non-vacuous golden evidence.** `OperationFingerprint.DerivationVersion = 2` and `CanonicalizationVersion = 1` are public derivation inputs, but the tests assert neither version and contain no literal expected digest. The current “golden” test only compares two serializations of a flat SimpleEventTypes record, so coordinated changes to the implementation and expected computation—or a derivation-version change—can pass silently. The packet/design requires a stable cross-runtime/provider contract, not merely self-consistency of the current implementation. Add frozen expected vectors that explicitly pin both versions and cover nested objects/dictionaries/arrays, numeric and string/Unicode escaping behavior, reflection and AOT metadata, configured naming and deterministic converters, plus the documented tag order/duplicate semantics. Shapes for which authoritative deterministic canonicalization cannot be established (for example converter-owned nondeterministic output) must fail with the typed canonicalization error before any store call. Do not claim the derivation algorithm is externally stable until these vectors pass independently of the implementation under test.

2. **`OperationCanonicalizationException` can expose payload content through an arbitrary inner exception.** The safe outer message omits the key and payload, but the implementation preserves exceptions produced by caller-configured converters/deserialization. Such an exception may include raw payload values in its message/data/inner graph, and the resulting exception is carried in `ResultBox.Error`. This does not meet the packet requirement that the typed fail-closed error be secret-free. Sanitize the carried failure boundary (or expose only a safe typed/categorized cause) and add a hostile converter test whose thrown exception contains a payload secret; assert that the full returned/loggable exception graph exposes neither that secret nor the idempotency key. The key is not currently passed to the converter, but that alone does not make the arbitrary inner exception safe.

Checked against `.intent-cli/issues/SEK-G15/{packet.yaml,review-context.md,implementation.md,github-body.md}`, `intents/sekiban/design/storage-concurrency-primitives.md`, the full repair delta, compatibility fixture/baseline, conditional executor paths, Hybrid forwarding, canonicalizer and tests. No G16 provider or expected-position scope was introduced.

tomohisa:
SEK-G15 strict-final rereview — request update

Head reviewed: `45b9f0de33ff734a28d9519e797cc28e6599f980`

The literal production-path digests, recursive object sorting, array/numeric/Unicode/NFC/tag/type vectors, and the secret-safe `OperationCanonicalizationException` repair are real and the 16 focused tests pass on both net9.0 and net10.0. All four current-head checks are successful. One contract blocker and two evidence/metadata gaps remain:

1. **Unsupported/non-deterministic converter shapes are documented but not detectably rejected in production.** `ComputeCanonical` accepts whatever valid JSON `IEventTypes.SerializeEventPayload` emits, canonicalizes it, and hashes it. A converter that alternates an array order/value or otherwise emits different valid JSON for the same object therefore returns successful but different fingerprints; no supported-shape boundary or deterministic metadata check rejects it. This violates the packet's cross-runtime/provider stable-operation identity and the explicit final-review requirement that unsupported shapes fail closed before store access. Add an authoritative production boundary (for example, constrain canonicalization to metadata-backed supported shapes or otherwise prove/validate determinism) and a non-vacuous alternating-converter test that invokes the real conditional executor path twice for the same semantic operation and proves typed failure with zero capability/store calls. Documentation saying such converters are unsupported is not enforcement.

2. **The requested Simple/AOT production-path parity vector is absent.** Every end-to-end test in `ConditionalAppendCanonicalContractTests` constructs `SimpleEventTypes`; none registers source-generated/AOT event metadata or pins that path to the same literal contract. Add an AOT/source-generated vector that traverses `ComputeCanonical` and freezes the expected digest (plus the relevant configured naming/converter behavior), rather than testing only `CanonicalizeJson` or the pure hash helper.

3. **PR metadata still contradicts the repaired diff.** The PR body says `CoreGeneralSekibanExecutor.cs` “is normalized to LF” and uses that to explain the inflated diff. The current branch instead preserves the base CRLF and adds the scoped `.gitattributes` rule; the raw semantic diff is additions-only. Update the body/evidence so the review record states the actual repair and includes the canonicalization v1 input in its ordered fingerprint summary.

The hostile failure test does recursively inspect `Message`, `Data`, `InnerException`, `StackTrace`, and `ToString()` at the externally returned `ResultBox.Error` boundary, with the same sentinel in payload/key/tag; the production exception graph has no cause. The five previously repaired fail-closed/Hybrid/compatibility/CRLF findings remain resolved, no G16 provider or expected-position scope was introduced, and `git diff --check` is clean.

Checked: `.intent-cli/issues/SEK-G15/{packet.yaml,review-context.md,implementation.md,github-body.md}`, `intents/sekiban/design/storage-concurrency-primitives.md`, full head and repair delta, production fingerprint/executor/capability paths, compatibility fixture, EN/JA docs, PR metadata, and exact-head GitHub checks.

tomohisa:
SEK-G15 strict-final rereview — request update

Head reviewed: `5ecb3b243c448916fb9c180ce30a1c164d85047f`

The optional metadata provider, genuine AOT/source-generated literal vector and Simple parity, root/nested converter rejection, executor no-write proof, PR metadata, and CRLF additions-only repair are present. The focused new suites pass on net9/net10 and all four exact-head checks are successful. One production contract blocker remains:

**`CanonicalShapeValidator` admits non-deterministic type/global converters on every allowlisted leaf.** `IsSupportedType` returns `true` for `string`, numeric types, enums, `byte[]`, date/time types, etc. *before* resolving their effective `JsonTypeInfo`. Consequently a normal Object root such as `record E(int Value)` passes even when its `JsonSerializerOptions.Converters` contains a custom `JsonConverter<int>` that alternates valid numeric output on each serialization. The property has no `JsonPropertyInfo.CustomConverter`, so the existing check does not see the converter; `ComputeCanonical` then deserializes/re-serializes through that global converter and successfully hashes whichever output it produced. The same hole applies to nullable/enum/other allowlisted leaves and a custom converter for an allowed leaf/collection such as `byte[]`. This contradicts the packet's stable cross-runtime/provider fingerprint contract and the PR/docs claim that “no custom converters are admitted.”

Required unblock: validate the *effective* metadata/converter contract for allowlisted leaves as well as structural nodes, without mutating caller/source-gen metadata. Built-in converters may be admitted through a defensible authoritative distinction; user/global/type-level converter ownership must fail closed. Add a non-vacuous real-executor test with an ordinary Object payload whose primitive property uses an alternating options-level converter. Execute twice and prove both return the typed secret-safe failure and that no conditional claim/event/store write occurs. Include Simple and AOT coverage where their effective converter metadata can differ. A root-level `[JsonConverter]` test alone does not kill this bypass.

Additional evidence note: the current hostile executor test proves the root converter-owned shape ultimately writes nothing, but its class summary says the converter is “never invoked” while the inline assertion commentary acknowledges executor serialization invokes it; the test does not assert `HostileConverter.WriteCalls`. Align the description and assert the allowed ordering/call count so this evidence cannot pass vacuously or overclaim pre-serialization rejection.

Verified as resolved/preserved: `IEventTypeJsonMetadataProvider` is additive and feature-detected (no `IEventTypes` member change); missing metadata affects only conditional canonicalization; AOT `ComputeCanonical` uses registered source-generated metadata and a literal digest; Simple/AOT parity is real; prior zero/multi/version/Hybrid/compatibility/secret-safe fixes remain; raw `CoreGeneralSekibanExecutor` diff is 221 additions/0 removals with CRLF preserved; no G16 provider or expected-position scope; `git diff --check` clean.

Checked against `.intent-cli/issues/SEK-G15/{packet.yaml,review-context.md,implementation.md,github-body.md}`, `intents/sekiban/design/storage-concurrency-primitives.md`, full repair delta, effective Simple/AOT metadata paths, production executor/store path, EN/JA docs, PR body, and exact-head GitHub facts.

tomohisa:
SEK-G15 strict rereview — request update

Head reviewed: `6e029d7928e6d78a89e214d3b68eb0664beeaa31`

The production leaf-validation order is repaired for reflection metadata: it resolves effective metadata before allowlist acceptance and checks root/type/options/property converter ownership. Comparing the converter type's actual `Assembly` object to the loaded `System.Text.Json` assembly is materially stronger than comparing a name and external code cannot place a type into that already-loaded framework assembly; structural source-gen Object nodes are separately validated by Kind and recursion. The focused tests pass on net9/net10 and all four exact-head checks are successful. Two required evidence gaps and one inaccurate test seam remain:

1. **No AOT/source-generated hostile converter rejection vector.** The existing AOT tests prove only a supported source-generated Object, frozen digest, and Simple parity. They do not construct source-generated/effective metadata with a property/type converter and prove the validator rejects it. Because the packet and PR promise the same Simple/AOT supported-shape semantics, add an AOT test covering converter-owned `Kind=None` and, where the source-gen options allow it, a property/leaf converter. It must traverse production `ComputeCanonical` (or the real conditional executor), return the typed secret-safe error, and perform zero durable writes.

2. **The alternating converter test does not prove its outputs differed.** It asserts only `WriteCalls >= 2`; a mutation that makes `Write` emit the same value each time still passes. Capture the two emitted values/serialized candidates through a test-only converter recorder and assert they are distinct, along with exact handler/converter counts, two typed secret-safe failures, no receipt, zero conditional store writes/events, and no sentinel in the externally observable exception graph. This is required to make the nondeterministic-shape regression non-vacuous.

3. **`InMemoryConditionalEventStore.WriteCalls` is described as successful durable writes but increments before `base.WriteSerializableEventsAsync` succeeds.** A failed base write is therefore counted as durable. Move the increment after the successful result (inside the same lock) or rename/document it as attempts; the current value is being used as authoritative no-side-effect evidence and must have truthful semantics.

The legacy unconditional path, optional metadata-provider compatibility, literal fingerprint vectors, secret-safe exception boundary, zero/multi/version/Hybrid/compatibility fixes, EN/JA and PR ordering text, CRLF additions-only diff, and G15 scope guard remain intact. `git diff --check` is clean and no G16 provider/expected-position work was introduced.

Checked: `.intent-cli/issues/SEK-G15/{packet.yaml,review-context.md,implementation.md,github-body.md}`, `intents/sekiban/design/storage-concurrency-primitives.md`, full R4 delta, Simple/AOT metadata and converter resolution, real executor/store tests, docs/PR body, and exact-head GitHub facts.

tomohisa:
SEK-G15 minimal-final rereview — request update (assertion-only remainder)

Head reviewed: `cce1923b0ef23daa5a7afbb2ed2257158bdac0bc`

The production fixes are coherent: AOT registered source-generated metadata reaches `AotEventTypes -> InMemoryConditionalEventStore -> ComputeCanonical`; alternating Simple candidates are captured and distinct with exact counts; `WriteCalls` increments only after base success and `AppendAttempts` has a separate truthful meaning. The focused set passes 14/14 on net9 and net10 and all four exact-head checks are successful. The strict delegation's required proof is still missing a few non-vacuous assertions:

1. **AOT hostile test does not assert zero attempts/receipt or recursively sanitized boundary.** Add `AppendAttempts == 0`, explicitly establish that no receipt/value can be obtained, and run the same secret sentinel through payload/key/tag where possible; recursively inspect `Message`, `Data`, `InnerException`, and `ToString()` on the returned error. The test currently checks only `InnerException == null`, `WriteCalls == 0`, and empty events.

2. **Alternating real-executor test does not verify both exception graphs are secret-free.** It now proves distinct emitted candidates and exact handler/converter counts, but uses no payload/key sentinels and checks only `InnerException == null`. Add distinct payload and idempotency-key sentinels and recursively assert neither appears in either externally observable exception graph. Keep the existing exact counts, `AppendAttempts == 0`, `WriteCalls == 0`, and empty events.

3. **Fail-first durable-write test does not pin the intermediate and final state completely.** Immediately after the injected failure assert empty events/no durable receipt or claim; after retry assert `AppendAttempts == 2`, `WriteCalls == 1`, exactly one event, and a valid `Appended` receipt. This makes “failure leaves no claim/receipt and retry succeeds without double count” mutation-relevant rather than inferred only from the retry status.

These are minimal test assertions; I found no remaining production semantic defect in the R5 delta. All prior accepted findings, legacy unconditional behavior, PR/docs truthfulness, CRLF additions-only diff, secret-safe production exception construction, and no-G16/expected-position scope guard remain intact. `git diff --check` is clean.

Checked against the SEK-G15 packet/review-context/implementation/github-body, `intents/sekiban/design/storage-concurrency-primitives.md`, full R5 delta, AOT/Simple metadata and executor/store paths, tests, PR metadata, and exact-head GitHub facts.

tomohisa:
SEK-G15 final rereview — completed, no blocking findings

Reviewed exact head `3d41be027a402df62febdb6793389a2f945cb41e` against the SEK-G15 packet/review-context and `intents/sekiban/design/storage-concurrency-primitives.md`.

The final test-only delta closes the remaining evidence gaps:

- AOT hostile converter uses registered source-generated `JsonTypeInfo` through `AotEventTypes -> conditional store -> ComputeCanonical`, returns the sanitized typed failure, and pins zero attempts/writes/events/claim/receipt plus recursive key-sentinel safety.
- The Simple alternating converter records two genuinely distinct serialized candidates with exact handler/converter counts; both external error graphs exclude payload/key sentinels recursively and all store effects remain zero.
- Fail-first durable append pins the empty/no-claim intermediate state, then exactly `AppendAttempts=2`, `WriteCalls=1`, one stored winner and a valid receipt; the third same-operation retry returns the same receipt without increments.

The delta is confined to four test files. All prior semantic, compatibility, canonicalization, secret-safety, legacy-path, Hybrid, version-ordering and CRLF findings remain resolved; no G16 provider or expected-position scope was introduced. Focused tests pass 13/13 on net9 and 13/13 on net10, `git diff --check` is clean, and all four exact-head GitHub checks are successful. Tests were treated as necessary, not sufficient; packet and design alignment were checked independently.

Review is released for the design-owned admin merge fallback. The review thread did not merge.

PR #1084, opened 2026-07-18
feat(dcb): SEK-G16 — conditional unique-append in Postgres, SQLite, Cosmos, DynamoDB (#1083)

### LINKED ISSUE #1083: SEK-G16 dcb: uniform unique-append across Postgres, Cosmos, SQLite, DynamoDB
# SEK-G16 dcb: uniform unique-append across Postgres, Cosmos, SQLite, DynamoDB

## Goal

Implement the SEK-G15 single-event conditional unique-append contract in all four
production providers with one uniform guarantee: for a given idempotency key,
exactly one append durably succeeds (Appended + durable receipt); a retry of the
same operation converges as AlreadyCommittedSameOperation via fingerprint match;
a same-key different-operation attempt is rejected as KeyReuseConflict — with
safe crash/retry convergence and capability propagation through
wrappers/factories.

## Why This Slice Exists Now

Issue #1080's concrete case — N racing hosts, exactly one may perform a one-time
migration — is directly answered by unique-append. Investigation showed every
provider already has the physical collision material (PG/SQLite primary keys,
Cosmos per-partition 409, Dynamo attribute_not_exists); what is missing is a
deterministic caller-derived identity, the fingerprint/receipt classification,
and a uniform contract. This is the lowest-risk cross-provider guarantee and does not depend on
the harder expected-position work (P2).

## Current Observed State

- SEK-G15 (merged) ships the optional interface, capability descriptor, the
  outcome state machine with fingerprint/receipt contract, fail-closed seams, and
  the InMemory reference semantics.
- Providers reject duplicate EventIds physically but surface generic exceptions;
  callers cannot supply a deterministic identity through the normal path.
- SQLite's existing write uses INSERT OR REPLACE (overwrites duplicates) — its
  conditional path must be new; the existing path must stay unchanged.

- The SEK-G15 contract shapes (interface, typed errors, capability descriptor,
  executor seams) are fixed; this slice implements providers against them.
- Key derivation must be identical across providers (same key = same storage
  identity everywhere), documented.
- Storage guarantees one durable claim; external side-effect exactly-once remains
  an application/outbox concern.

## Target Repo / Path / Part

Target part: single-event conditional unique-append in the four provider event stores + fingerprint/receipt outcome classification + capability reporting + wrapper/factory propagation + tests + docs.

## In Scope

- **Postgres**: transactional conditional insert; SQLSTATE 23505 → fingerprint
  comparison against the durable receipt → AlreadyCommittedSameOperation or
  KeyReuseConflict (provider exception preserved as diagnostic cause).
- **Cosmos**: deterministic event identity per key; a CreateItem 409 triggers
  fingerprint comparison AND repair/verification — AlreadyCommittedSameOperation
  only once the committed state is verified (409 alone proves nothing);
  interaction with the 10.3 roll-forward/repair machinery covered by
  crash-window tests. Single-event contract; multi-event requests fail closed.
- **SQLite**: NEW conditional path (plain INSERT / ON CONFLICT DO NOTHING with
  rejection detection) — existing INSERT OR REPLACE path untouched and
  regression-pinned.
- **DynamoDB**: conditional put + attribute_not_exists; ConditionalCheckFailed /
  cancellation → fingerprint comparison → amended outcome classification.
  Single-event contract; multi-event requests fail closed.
- **Deterministic key→identity derivation** identical across providers,
  documented per provider.
- **Crash/retry convergence tests** per provider (fault seams where available):
  retry after crash/ambiguous failure finds the durable claim or completes; no
  double effect.
- **Single-event contract enforced everywhere**: multi-event conditional
  requests fail closed with the typed capability error on every provider (no
  partial application anywhere; batch claims are a future capability kind).
- **Propagation**: Hybrid/hot-store capability forwarding; composite stores
  report supported only when all underlying stores support it (tested).
- **End-to-end race tests**: (a) N concurrent writers, same key + same
  operation — one Appended, rest AlreadyCommittedSameOperation with the
  identical receipt, all converge on the durable winner (real provider
  containers where available); (b) same key + different operation —
  KeyReuseConflict.
- **Docs (EN + JA)**: per-provider support, derivation/fingerprint, the outcome
  machine, the N-host migration recipe (claimed/completed or outbox for external
  work), one-durable-claim boundary.

## Out Of Scope

- Expected-position/CAS (P2; the Cosmos variant additionally requires a
  separately approved protocol design).
- Changes to unconditional write paths, IEventStore, or default behavior.
- Command-ergonomics expansion beyond the SEK-G15 seams (P3).

The child PR must implement the SEK-G15 conditional unique-append contract in
Postgres, Cosmos, SQLite (new path only), and DynamoDB with an identical
key→identity derivation, the amended outcome classification (fingerprint-verified
AlreadyCommittedSameOperation / KeyReuseConflict) preserving provider errors as
diagnostic causes when present, single-event enforcement that never partially
applies, safe
crash/retry convergence proven per provider, capability reporting with
wrapper/composite propagation, an N-writer race test converging on one durable
winner, and EN + JA docs including the migration recipe — with unconditional
paths and defaults byte-for-byte unchanged.

## Acceptance Criteria

- Four providers implement the SINGLE-EVENT unique-append contract with the uniform G15 outcome machine (Appended + receipt / AlreadyCommittedSameOperation via fingerprint match / KeyReuseConflict / fail-closed unsupported); capability descriptors report the single-event kind.
- Identical documented key→identity derivation and fingerprint computation across providers (same key + same operation = same identity + fingerprint everywhere).
- SQLite conditional path is new; existing INSERT OR REPLACE path regression-pinned unchanged.
- Provider rejections (23505 / 409 / ConditionalCheckFailed / constraint) preserved as diagnostic causes; read-discovered duplicates carry no fabricated exception.
- Crash/ambiguous-failure retry converges via receipt + fingerprint per provider (tested, fault seams where available); cancellation/timeout after durable commit resolves through the same receipt path.
- Cosmos: AlreadyCommittedSameOperation only after repair/verification reaches committed state; 409 alone is insufficient (crash-window tested).
- Multi-event conditional requests fail closed everywhere (single-event contract; no partial application).
- Hybrid/composite/factory propagation tested incl. capability=false fail-closed case.
- Race tests: N writers same key + same operation → one Appended, rest AlreadyCommittedSameOperation with identical receipt; same key + different operation → KeyReuseConflict.
- Docs EN + JA (per-provider support, derivation/fingerprint, recipe with claimed/completed-or-outbox guidance, one-durable-claim boundary).
- No unconditional-path, schema-migration, or default-behavior changes (additive only).

## Verification

## Related Links

- https://github.com/J-Tech-Japan/Sekiban/issues/1080
- Host design record: intents/sekiban/design/storage-concurrency-primitives.md
- Depends on SEK-G15 (contract/capability/errors)

## Knowledge Maintenance

- Intent placement: host design doc storage-concurrency-primitives.md (P1).
- Diagram candidate: none.
- Docs update: 11_storage_providers.md + 13_common_issues.md EN/JA — in scope.
- Closeout writeback expected: yes — per-provider RU/latency notes and P2 (expected-position) go/no-go inputs incl. Cosmos protocol-design material.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1083.

Implements SEK-G16: the SEK-G15 single-event conditional unique-append contract in all four production providers (PostgreSQL, SQLite, Cosmos DB, DynamoDB) with a crash-safe, secret-safe, uniform outcome machine. Additive only — the unconditional path, `IEventStore`, and every default are unchanged; no schema migration; no expected-position/CAS.

## Shared orchestrator (Core)
`ConditionalAppendExecution` defines the outcome once. A provider supplies THREE primitives: durably write the claim under a deterministic id; read the committed winner back; and (only where event/index rows are NOT written atomically) bring the winner's committed state to convergence. Deterministic identity: `EventId = SHA-256(domain-sep ‖ version ‖ ServiceId ‖ normalized-key)` (`ConditionalAppendIdentity`) — the existing primary key IS the uniqueness primitive, so **no schema migration / no new column** on any provider.

## Outcome taxonomy
- **Appended** / **AlreadyCommittedSameOperation** (durable receipt: winner EventId, SortableUniqueId, fingerprint).
- **KeyReuseConflict** — same key, different operation (typed; real provider exception preserved as cause only when one occurred).
- **Typed retryable in-doubt** (`ConditionalAppendInDoubtException`, `IsRetryable=true`) — closed `ConditionalAppendInDoubtReason` enum (`WinnerUnreadableAfterConflict` / `AmbiguousAfterWrite` / `CommittedStateUnverified`), stable `ReasonCode`, provider/ServiceId/derived-EventId, preserved cause; never the raw key/payload. Not a fifth status; a retry converges.
- **Non-retryable committed-state corruption** (`ConditionalAppendCommittedStateCorruptionException`, `IsRetryable=false`) — an existing committed index/tag row DISAGREES with the event. The row is never overwritten; the unsafe provider integrity exception is NOT chained (only sanitized safe context + derived row-id hash).
- **ConditionNotSupported** — fails closed before any store call.
- **DynamoConditionalAppendLimitException** — TransactWriteItems limit (>100 items or duplicate item key) enforced before the network.

On WithResult these are `ResultBox.Error`; on WithoutResult the guarded boundary rethrows the exact typed exception; the versioned serialized boundary preserves it verbatim (unknown version rejected first).

## Per-provider mapping (only the intended collision is a claim conflict)
| Provider | Uniqueness / write shape | Mapped conflict | Winner read-back | Committed-state gate |
|---|---|---|---|---|
| PostgreSQL | `(ServiceId,Id)` PK, plain txn, event+tags in ONE transaction | 23505 **with ConstraintName `PK_dcb_events`** (other 23505 = provider failure) | `AsNoTracking` point read | none (atomic) |
| SQLite | `(ServiceId,Id)` PK, plain INSERT (legacy INSERT OR REPLACE unchanged), event+tags in ONE transaction | `SqliteExtendedErrorCode 1555` PRIMARYKEY on isolated event insert | point read | none (atomic) |
| Cosmos DB | item id in partition `{svc}\|{id}`, event doc then tag rows in **separate phases** | 409 Conflict | consistent point read | **required** — idempotently repair/verify every missing tag row before AlreadyCommitted; disagreeing row → corruption |
| DynamoDB | `attribute_not_exists(pk)`, ONE TransactWriteItems (event Put index 0 + one Put/tag), no ClientRequestToken | `ConditionalCheckFailed` at cancellation-reason **index 0** (other index/reason = provider failure) | `GetItem` ConsistentRead | none (atomic) |

Atomicity is **not uniform**: PG/SQLite/Dynamo commit event+tags together; Cosmos does not, so its committed-state gate closes the event/tag window on the next same-op attempt. DI, the per-service factories, and `HybridEventStore` propagate the capability (fail-closed composite intersection).

## Test evidence (net9.0 + net10.0)
- **Uniform contract** per provider: Appended, same-op identical receipt, different-op KeyReuseConflict.
- **Post-write failure taxonomy** (distinct cases, each proven):
  - **(a) rollback before commit (known pre-commit failure)**: SQLite/Postgres cancelled-before-commit → nothing durable; surfaces as the EXACT original OperationCanceledException/token, NEVER AmbiguousAfterWrite (phase-aware — only a provider-declared post-commit marker enters verification); retry converges.
  - **(b) post-commit response loss (transport OR cancellation)**: PG/SQLite/Cosmos commit durably, then lose the response. The provider raises an INTERNAL post-commit ambiguity marker (never raw transport), and the shared orchestrator resolves it authoritatively ON THE SAME CALL under a BOUNDED, caller-independent verification budget → AlreadyCommittedSameOperation with a receipt IDENTICAL to the independently-read stored winner (deterministic EventId, SortableUniqueId, canonical fingerprint), no second event/tag. Cosmos uses a multi-tag event proving every ordered tag row exists/content-matches with no duplicates.
  - **(c) bounded verification cannot complete**: an orchestrator unit test proves readback-within-budget → AlreadyCommitted and readback-past-budget → prompt typed `AmbiguousAfterWrite` preserving the original OperationCanceledException and its exact CancellationToken (caller cancellation never gates the verification).
  - **(d) conflict with unresolvable winner**: Cosmos/Dynamo bare-conflict-no-winner and Cosmos repair-failure → typed retryable in-doubt (closed reason).
- **Concurrency**: real cross-transaction N-writer race on Postgres (Testcontainers) and DynamoDB (thread-safe fake); concurrent writers + different-op race on Cosmos (lock-serialized in-memory double) and SQLite (independent store instances/connections against the same real DB — not one-store semaphore).
- **Precise mapping**: Postgres unrelated-unique-index and SQLite unrelated-unique-constraint stay provider failures; Dynamo absent/malformed/wrong-index cancellation reasons and a request-too-large rejection stay provider failures; Dynamo 99(ok)/100(fail)/duplicate limit boundaries.
- **Cosmos crash-window/repair**: crash after event create → retry repairs tags → AlreadyCommitted with exact tag-scoped visibility; repair-still-failing → in-doubt; bare-409-no-winner → in-doubt then recovery converges.
- **Corruption**: multi-tag mismatched-first + missing-later row → non-retryable, no overwrite, no later repair, full before/after byte equality, repeat stability, recursive exception-graph secret-safety (sentinels in row content).
- **Facade/serialized**: typed in-doubt and corruption through WithResult (Error), WithoutResult (throw), and the versioned serialized boundary assert `Assert.Same` instance + cause identity, closed reason/retryability/safe context, recursive key/payload secret-safety, and version-first rejection with zero store/capability side effects.

Real-provider vs double: Postgres is a real container (Testcontainers); SQLite is a real temp-file DB; Cosmos uses the production fault seams over an in-memory faithful double; DynamoDB uses a thread-safe fake modeling atomic conditional put, cancellation-reason indices, ambiguous commit, and limits.

## Docs
EN + JA `docs/dcb_llm{,_ja}/11_storage_providers.md` (per-provider mapping, Cosmos repair gate, Dynamo transaction/limits, typed in-doubt vs corruption, non-uniform atomicity) and `13_common_issues.md` (conditional-append troubleshooting).


tomohisa:
SEK-G16 strict review — request update

Head reviewed: `399b4815f485dab998a92763af0508dad75b9b49`

The G15 public outcome/fingerprint/canonicalization surfaces are unchanged, deterministic storage identity is shared, and the basic provider outcome suites pass. However, the packet's crash-safe uniform contract is not implemented or evidenced. Blocking implementation findings:

1. **Cosmos returns `AlreadyCommittedSameOperation` from an event-only 409 without restoring/verifying the committed tag state.** `TryWriteConditionalClaimAsync` creates the event and then writes tags. If the tag write fails/crashes, the event remains. The next call receives 409, `ReadConditionalWinnerAsync` reads only the event, and shared `ConditionalAppendExecution` immediately returns `AlreadyCommittedSameOperation` on fingerprint match. It never invokes `WriteSerializableTagsWithBatchAsync`, the repair service, or any tag verification. This directly violates the packet/G15 design requirement that the event/tag roll-forward window be closed before the same-operation committed receipt is returned. Add a Cosmos-specific convergence protocol: on conflict, authoritative event fingerprint first, then idempotently repair/verify every required tag row to the contracted committed state before returning AlreadyCommitted. Cover failures after event create and before/mid/after tag batches, retry exhaustion/cancellation, fresh store activation, corrupt rows, and a retry that proves exact tag visibility—not merely one event document. A bare 409 test with no winner is necessary but does not cover the real partial-commit window.

2. **“Retryable in-doubt” is an undocumented generic error and loses real conflict context.** When a conflict is followed by a missing/unreadable winner, `ConditionalAppendExecution` creates a plain `InvalidOperationException` and discards `outcome.ConflictException` (for example the real Cosmos 409). There is no typed or machine-readable retryability contract, while docs tell callers to retry. This is not a fifth success/conflict outcome, but it still needs a stable typed failure or an original-exception-preserving G9-style boundary (`Exception.Data`/cause as appropriate) that distinguishes in-doubt from permanent failure without exposing key/payload. Define and test it consistently for conflict-read gaps, timeout/cancellation after commit, and ambiguous provider responses; do not generically wrap or fabricate an outcome.

3. **Provider collision mapping is too broad or unbounded.** PostgreSQL treats every SQLSTATE 23505 from event/tag `SaveChanges` as the deterministic `(ServiceId, Id)` claim collision; it does not verify the intended constraint, so an unrelated tag/unique constraint is misrouted to winner classification. SQLite similarly catches base result code 19 on the event INSERT without pinning the PK/extended constraint and has no rollback/fault proof. Dynamo treats any transaction cancellation reason containing `ConditionalCheckFailed` as the event claim condition, does not pin the reason index, and builds `1 + tag-count` transact items without enforcing Dynamo's 100-item/size limits or duplicate-item rules before the call. Tighten provider-specific classification; unrelated validation/constraint failures must preserve their original failure and never become claim conflicts. Add 99/100-tag boundary (and size/duplicate-key where applicable) with fail-before-network behavior.

4. **The required crash/ambiguous/race matrix is largely absent.** Shared tests cover happy first/retry/different operation. There are no per-provider fault tests for pre-commit, post-commit-before-return, cancellation/timeout, unreadable winner then recovery, or retry convergence. Cosmos uses a permissive in-memory container and only sequential retries; Dynamo's fake has no ambiguous-success/cancellation-reason-index/limit behavior. SQLite's “N writers” all call one store instance and are serialized by its instance `_writeLock`, so it does not prove independent store/process writers against the real DB; there is no transaction rollback test. PostgreSQL has a real same-operation race but no different-operation race, constraint-name discrimination, isolation/read timing, or ambiguous commit test. Add non-vacuous same-key same-op and mixed/different-op races plus provider fault seams; use independent SQLite stores/connections, Postgres Testcontainers, a Cosmos emulator or production-faithful transactional/fault harness, and a Dynamo fake that models exact cancellation reasons and ambiguous commit/readback.

5. **Required docs/evidence are incomplete and overclaim uniformity.** Packet knowledge maintenance requires EN/JA `13_common_issues.md`, but the PR changes only `11_storage_providers.md`. The provider table says Cosmos point-read verification is sufficient and Dynamo is a “single-item” transaction although it includes one event plus every tag. Docs must describe the actual Cosmos repair/verification gate, typed in-doubt representation, transaction limits/RU/latency, and provider-specific crash boundaries without implying green shared tests prove atomicity. Update EN/JA 11 and 13 equivalently.

Additional verified facts: deterministic EventId includes ServiceId and normalized key; no expected-position/CAS or schema migration was introduced; legacy SQLite `INSERT OR REPLACE` remains on the unconditional path; G15 derivation v2/canonicalization v1 and four-outcome state machine did not drift; capability/factory wiring uses the selected store. Focused Cosmos/Dynamo/SQLite/propagation tests pass 18/18 on net9 and net10, and all four exact-head GitHub checks are successful, but tests are necessary—not sufficient—for packet acceptance.

Checked against `.intent-cli/issues/SEK-G16/{packet.yaml,review-context.md,implementation.md,github-body.md}`, `intents/sekiban/design/storage-concurrency-primitives.md` at host contract `ab4bfb5`, all four provider implementations, shared coordinator/identity, factories/DI/Hybrid, provider tests, docs, full diff, and exact-head GitHub facts.

tomohisa:
SEK-G16 strict rereview — request update

I rechecked head `bc63e34983c5cf076fab3f52546ee3f84776bb32` against the packet/review-context and `intents/sekiban/design/storage-concurrency-primitives.md`. The four current-head checks are green, and the Postgres constraint-name narrowing, SQLite event-insert boundary, Dynamo item-count/index checks, and Cosmos missing-tag repair are meaningful improvements. The following current-PR contract gaps remain:

1. **Cosmos corruption is incorrectly classified as retryable in-doubt.** `CosmosTagWriteStage.EnsureRowAsync` correctly throws `CosmosTagIndexCorruptionException` when an existing deterministic row fails strict `ContentEquals`, but `ConditionalAppendExecution.ResolveByReadbackAsync` catches every `ensureCommittedAsync` exception and replaces it with `ConditionalAppendInDoubtException(IsRetryable=true)`. That tells callers to retry a known non-retryable corruption indefinitely and contradicts the packet requirement to handle mismatched rows and the established Cosmos corruption contract. Preserve the typed non-retryable corruption result (while transient/unresolved repair failures remain typed in-doubt), and add a service-level collision/readback test with a partially present mismatched tag row proving: no overwrite, no `AlreadyCommitted`, and the correct non-retryable classification.

2. **`ReasonCode` is not a closed machine-readable contract.** `ConditionalAppendInDoubtException` exposes a public constructor accepting any `string reasonCode`, despite the contract and docs promising stable reason codes that cannot drift. The three constants do not prevent arbitrary values. Use a closed additive representation (or validate the constructor against the supported set / expose controlled factories) and pin invalid-value rejection plus serialization/facade behavior.

3. **The claimed facade/provider failure matrix is still not evidenced.** `ConditionalAppendParityTests` covers Appended, AlreadyCommitted, unsupported, and KeyReuseConflict only; it never drives a real typed in-doubt through WithoutResult, and no serialized-boundary in-doubt test exists. Provider tests also remain uneven: SQLite has real concurrent writers but no injected transaction rollback/ambiguous-after-commit case; Postgres has races and unrelated-constraint discrimination but no crash/ambiguous readback case; Cosmos uses the faithful in-memory seam but has no mismatched-row or cancellation/readback case; Dynamo uses a fake and does not exercise absent/malformed cancellation-reason collections or request-size/provider rejection. Add the packet-required non-vacuous boundary and per-provider cases, and describe the real-provider/emulator/fake evidence without overclaiming.

These are implementation findings tied to the packet's typed uniform outcome, Cosmos complete committed-state verification, and per-provider crash/ambiguous-path acceptance criteria. No G16 merge is authorized by this review.

Evidence checked: packet.yaml, review-context.md, implementation.md, github-body.md, storage-concurrency-primitives design, EN/JA storage-provider/common-issues docs, full repair delta from `399b4815`, production provider paths and focused conditional tests. `git diff --check` is clean; GitHub reports dcbTestsNet9, dcbTestsNet10, SonarCloud, and SonarCloud Code Analysis all SUCCESS on the exact head.

tomohisa:
SEK-G16 strict rereview of `38385fb6` — request update

The previous three findings are only partially resolved. The closed enum, Cosmos corruption branch, and added facade/provider tests are directionally correct, but these current-PR blockers remain against the packet's uniform crash-safe and secret-safe contract:

1. **The new Cosmos corruption boundary is not secret-safe and the no-later-repair proof is incomplete.** `ConditionalAppendCommittedStateCorruptionException` preserves `CosmosTagIndexCorruptionException` as `InnerException`. That inner exception's public message/properties include the raw tag, partition key, and full expected/actual row descriptions. Consequently the externally observable recursive exception graph can disclose tag content, while the new facade test checks only payload/key sentinels on synthetic safe causes and the Cosmos test performs no recursive secret assertion. The Cosmos test also seeds only one required row and checks one `eventType` value; it does not prove exact full-row bytes are preserved or that a mismatch stops later missing rows from being created. Sanitize the externally preserved cause (without losing a useful typed/non-retryable diagnostic), and add a multi-tag partial fixture with a mismatched first row plus a missing later row proving full before/after row equality, zero later repair/overwrite, repeat stability, and recursive secret safety.

2. **True post-durable-commit ambiguity is still missing for three providers.** The new Postgres and SQLite tests cancel an already-cancelled token and explicitly prove rollback-before-commit. That is useful, but it is not the packet-required seam where the transaction commits successfully and only the response is lost. Neither provider has an apply/commit-then-throw seam followed by authoritative read/fingerprint convergence. Cosmos likewise has event-only crash/repair and bare 409 cases, but no post-durable-write cancellation/timeout seam proving readback plus complete tag convergence, nor an unreadable-winner variant. Dynamo has the only genuine apply-then-throw test (in a faithful fake). Add non-vacuous provider-specific post-commit response-loss seams/tests for Postgres, SQLite, and Cosmos; keep rollback-known tests separate and label evidence accurately.

3. **Facade evidence does not yet pin the identity/token/secret contract it claims.** The WithResult, WithoutResult, and serialized tests assert only exception type/properties; none asserts `Assert.Same` for the injected original exception or its cause, and no cancellation cause verifies original `CancellationToken`. `AssertSecretSafe` checks only `ToString()` rather than Message/Data/InnerException recursively, despite a shared recursive helper already existing. The serialized unknown-version test also does not assert zero handler/capability/store effects. Strengthen these tests through the real boundaries to prove exact instance/cause/token preservation as applicable, recursive graph safety, version-first rejection, and zero side effects.

4. **PR metadata remains materially stale and overbroad.** The current body still describes the old two-primitive orchestrator, broad Postgres `23505` / SQLite code 19 mappings, omits the committed-state gate and typed corruption/in-doubt distinctions, and reports only an earlier provider/test matrix. Update it to the exact repaired head and clearly distinguish rollback-known, ambiguous-after-commit, in-doubt, corruption, and fake/emulator/real-provider evidence.

Checked: SEK-G16 packet/review-context/implementation/github-body, `storage-concurrency-primitives.md`, full `bc63e349..38385fb6` delta, provider production paths/tests, and EN/JA docs. `git diff --check` is clean. Focused conditional tests exit 0. GitHub reports all four exact-head checks SUCCESS. No G17/P2 scope was accepted and no merge is authorized.

tomohisa:
SEK-G16 strict rereview of `974cd039` — request update

R3 resolves the prior secret-bearing Cosmos inner exception and adds genuine commit-before-hook seams. The full packet still has these implementation/evidence gaps:

1. **The post-commit recovery tests do not yet prove the required fresh-instance + identical-winner contract across providers.** SQLite and Cosmos retry with a fresh store, but Postgres retries through the same fixture `PostgresEventStore`. None of the three transport-loss tests pins the original durable winner's complete receipt (EventId, SortableUniqueId, canonical fingerprint) and proves the retry receipt is identical; they assert only status/count. Cosmos uses a single-tag event, so this path does not prove complete ordered tag-set visibility after response loss. Use a fresh Postgres store/context against the same real database, capture/read the durable winner independently, assert the exact retry receipt against it on every provider, and use a multi-tag Cosmos case proving all rows/content before `AlreadyCommitted`.

2. **The test-only production seams are not structurally guarded as required.** Three mutable `AfterConditionalCommitHook` members were added to production store classes, plus new IVT grants. They are internal and not registered today, but there is no architecture/API test proving they remain absent from public surface, production DI/factory configuration, and non-test assembly usage. Add a non-vacuous structural guard (including the narrowly scoped IVT expectations) so these response-loss hooks cannot silently become a reachable runtime mutation surface.

3. **Version-first zero-side-effect evidence remains narrower than the delegated contract.** `SerializedBoundary_UnknownVersion...` counts only capability description and store append. It does not independently prove zero handler invocation, EventId allocation, payload serialization/canonicalization, or underlying store reads/writes. Add counters/seams at the real public serialized executor boundary for each preflight stage named by the packet, and keep the exact typed instance/cause/token assertions already added for the successful boundary traversal.

4. **EN/JA docs do not yet document the new response-loss taxonomy claimed by the PR body.** R3 changed production/tests and PR metadata but no docs. The provider/common-issues pages still lack an explicit distinction between: known rollback-before-commit; non-cancellation transport failure after durable commit (first call surfaces the original provider/transport failure and explicit retry converges); cancellation/timeout after durable commit (same-call independent readback may resolve); and typed in-doubt when authoritative verification cannot resolve. Update EN/JA text so operators do not infer that every post-commit response loss is same-call typed in-doubt or same-call success.

Verified improvements: Cosmos corruption graph is sanitized and the multi-row mismatch test proves exact before/after JSON, no later repair, repeat stability; SQLite's committed guard avoids rollback after commit; Postgres/SQLite hooks run after actual commit; Cosmos hook runs after event plus tags; facade exact-instance/cause assertions and PR metadata are materially improved. Checked packet/review-context/implementation/github-body, `storage-concurrency-primitives.md`, full R3 delta, provider paths/tests, and EN/JA docs. `git diff --check` is clean, focused conditional tests exit 0, and all four exact-head checks are SUCCESS. No merge or P2 scope is authorized.

tomohisa:
SEK-G16 strict rereview of `16d795df` — request update

R4 materially improves fresh-store receipt reconstruction, multi-tag Cosmos proof, seam guards, and EN/JA troubleshooting text. Two semantic blockers and one structural-evidence blocker remain:

1. **Post-commit ambiguity still has two incompatible observable contracts.** The authoritative packet/review-context requires a uniform typed, machine-readable, crash/ambiguous-failure contract and G9/G14 typed failure preservation. Yet `ConditionalAppendExecution` handles only `OperationCanceledException`/`TimeoutException` via authoritative readback; every other exception is returned raw. The new Postgres/SQLite/Cosmos hooks deliberately throw `InvalidOperationException` after durable commit, and tests/docs now freeze the first result as that raw transport exception. A caller cannot know it is retryable/in-doubt and receives neither the derived identity nor stable reason code. This contradicts the typed `ConditionalAppendInDoubtException(AmbiguousAfterWrite)` contract introduced by the prior repair and makes observable semantics depend on exception class rather than commit ambiguity. Route provider-declared post-commit response loss through the shared authoritative receipt resolution; if it cannot resolve, return typed `AmbiguousAfterWrite` preserving the real transport cause. Do not broadly convert known permanent pre-commit provider failures—use an explicit internal provider outcome/seam that distinguishes the ambiguous-after-write condition.

2. **Cancellation verification is independent but unbounded.** The shared catch calls `ResolveByReadbackAsync(..., CancellationToken.None)`. This avoids the already-cancelled caller token, but can hang indefinitely on a failed provider and is not the delegated bounded independent verification semantic. Add an internal bounded verification timeout/token policy, prove same-call success when readback completes, prove typed in-doubt when the bounded read/committed-state gate cannot complete, and preserve the original cancellation exception/token as the diagnostic cause.

3. **Version-first and test-seam structural proofs remain incomplete.** The unsupported-version test still infers zero handler/EventId/SortableUniqueId/serialization/canonicalization solely from `DescribeCalls == 0` plus store counters; it has no direct probe for the named stages, so moving allocation or serialization ahead of both version and capability could pass. Add direct non-vacuous counters/hooks or an architecture/control-flow mutation guard for each packet-required stage. Likewise, the seam architecture tests prove reflection visibility/default instance state but do not guard the exact IVT allowlist or absence of production assignments/DI/options exposure; `rg` currently finds assignments only in tests, but that fact is not regression-pinned.

Verified as resolved: fresh Postgres store over the shared real DB; independently reconstructed exact receipts for Postgres/SQLite/Cosmos; Cosmos three-tag exact visibility/no duplicates; sanitized multi-row corruption; post-commit hook placement; refreshed PR body and EN/JA taxonomy; prior provider mapping/limits/capability/legacy/G15/no-P2 constraints. Checked packet.yaml, review-context, implementation, host design `storage-concurrency-primitives.md`, full R4 delta, provider tests/docs. `git diff --check` is clean, focused conditional tests exit 0, and exact-head GitHub checks are being treated as necessary but not sufficient. No merge is authorized.

tomohisa:
SEK-G16 strict rereview of `1c1fc0f4` — request update

R5 adds the right internal post-commit marker concept and a nominal verification budget, but the phase and budget contracts are not yet correct end to end:

1. **Known pre-commit rollback is still mislabeled as ambiguous.** `ConditionalAppendExecution` continues to catch every `OperationCanceledException` or `TimeoutException` from `tryWriteClaim` and route it through `ResolveAmbiguousAfterWriteAsync`. The Postgres/SQLite cancelled-before-commit tests explicitly prove rollback/no durable event, yet still expect typed `AmbiguousAfterWrite`. The delegated contract requires only `PostCommitResponseLostException` (or a provider outcome explicitly saying commit status is unknown) to enter ambiguity verification; a provider-known rollback must preserve the original failure. Add a phase-aware provider result/marker for unknown commit status if needed, keep known rollback cancellation/timeout raw and identity-preserved, and mutation-pin all three phases: known pre-commit failure, unknown commit status, and provider-confirmed post-commit response loss.

2. **The 30-second budget is not enforced by the Cosmos production path.** `ResolveAmbiguousAfterWriteAsync` passes `cts.Token`, but `CosmosDbEventStore.ReadConditionalWinnerAsync` calls `ReadItemAsync` without that token, and `EnsureConditionalCommittedAsync` invokes `WriteSerializableTagsWithBatchAsync` without propagating the budget token. A stuck Cosmos read or tag convergence can therefore exceed the advertised bound indefinitely. The focused bounded test uses a cooperative fake delegate, so it cannot detect this production gap. Thread the independent budget token through Cosmos point read and every repair/retry/delay operation, add deterministic non-cooperative/provider-seam coverage for both winner read and ensureCommitted timeout, and prove no retry/background work remains after the typed in-doubt result. Audit the other providers similarly.

3. **The serialized version test still lacks the required direct handler/serialization/canonicalization probes.** R5 directly probes EventId and SortableUniqueId allocation, which is good, but continues to infer serialization/canonicalization from `AppendAttempts == 0`; there is no handler counter at all. The packet explicitly requires handler invocation, allocation, serialization, and store call to be individually fixed behind the version gate. Add a direct mutation-relevant serializer/canonicalizer probe, and either a direct handler probe on the applicable public conditional executor path or a structural test proving the versioned serialized method has no handler path and that the ordinary handler path cannot consume an unsupported serialized request.

4. **The seam assignment/IVT guard is incomplete and the Core diff has avoidable EOL churn.** Reflection tests pin visibility/defaults and exact IVT names, but do not regression-pin that production source/DI contains no assignments to `AfterConditionalCommitHook` or allocator factories (current source search shows test assignments only). Add a structural source/IL guard. Also restore the existing CRLF form of `CoreGeneralSekibanExecutor.cs`; raw diff is 121+/112- while `--ignore-space-at-eol` shows only 11+/2-, obscuring review of a sensitive executor.

Resolved and retained: internal secret-safe `PostCommitResponseLostException`; provider hooks after durable commit; exact fresh-store receipts; Cosmos corruption/multi-tag proofs; closed reason codes; facade identity; provider mappings/limits; current docs/PR taxonomy. Checked packet/review-context/implementation, `storage-concurrency-primitives.md`, full R5 delta, production providers/tests, EN/JA docs. `git diff --check` is clean, focused tests exit 0, and all four exact-head checks are SUCCESS, treated as necessary but not sufficient. No merge/P2 scope authorized.

tomohisa:
SEK-G16 strict rereview on `cb29aed38e311db0b6fb756a755ece507b840fe9`: request update.

Two blocking findings remain:

1. DynamoDB pre-dispatch cancellation is still misclassified as post-commit ambiguity. In `DynamoDbEventStore.TryWriteConditionalClaimAsync`, the already-cancelled caller token is passed directly to `TransactWriteItemsAsync` (lines 118-122), and every resulting `OperationCanceledException` is wrapped as `PostCommitResponseLostException` (lines 139-144). An SDK can reject an already-cancelled token without dispatching any request; that is known no-commit and must preserve the original cancellation/token. Add an explicit pre-dispatch cancellation boundary and non-vacuous tests proving: already-cancelled before invocation => zero SDK dispatch/durable state + exact raw OCE token; cancellation/timeout after dispatch => ambiguity marker/readback/typed fallback as applicable.

2. The production seam assignment architecture test does not cover the required mutation surface. `AssertNoSetterCallSites` scans only `call/callvirt set_*` (lines 157-180). Assignments within a declaring type compile as `stfld` (including auto-property backing fields), and reflection can assign through `PropertyInfo.SetValue`; both evade this test. The test's own comment acknowledges `stfld` but excludes it. Strengthen the structural guard to cover direct backing-field stores outside the narrowly allowed default initialization, init-only/delegate field assignments, and reflection-based production assignment paths, while retaining the exact IVT allowlist.

Verified on exact head: GitHub OPEN/MERGEABLE; dcbTestsNet9, dcbTestsNet10, SonarCloud, and SonarCloud Code Analysis all SUCCESS. Cosmos bounded verification now threads the budget token through winner read and tag create/read/batch paths, and the blocking-read test demonstrates prompt completion/no later mutation. Serialized version validation is visibly the first statement before capability/allocation/candidate construction/append; this boundary has no handler or serialization/deserialization before append. CoreGeneralSekibanExecutor diff versus base is 14+/5- both raw and ignore-EOL; diff-check is clean.

Design alignment checked against `.intent-cli/issues/SEK-G16/{packet.yaml,review-context.md,implementation.md,github-body.md}` and `intents/sekiban/design/storage-concurrency-primitives.md`.

tomohisa:
SEK-G16 minimal final rereview on `34ef44211b2dfc8a6031caba6ea9b8160bbfe6d7`: one blocking evidence gap remains.

The DynamoDB phase-classification finding is resolved: `ThrowIfCancellationRequested` is immediately before the SDK call and outside the unknown-commit catch; the new test proves exact-token raw cancellation, zero dispatch, and zero durable state, while the existing commit-then-cancel test preserves the post-dispatch ambiguity/readback path.

Blocking: the seam-write scanner still does not cover the claimed assignment surface and its control is not mode-specific.

- `FindSeamWrites` reports reflection `SetValue` only when `seamTypes.Contains(type)`, where `type` is the type containing the calling method. A different production type can obtain the seam property and call `PropertyInfo.SetValue`; that caller is not a seam-declaring type, so the write is silently ignored. This contradicts the required “no production reflection assignment escapes scan” guard.
- `SeamWriteScanner_IsNonVacuous_DetectsRealWrites_InTheTestAssembly` asserts only that the aggregate hit list is nonempty. A setter-call hit alone makes it pass even if `stfld` and reflection detection are broken. The current test assembly’s reflection assignments occur in non-seam-declaring test types, so this control does not prove reflection detection at all. Add independent, asserted controls for setter call, direct backing-field `stfld`/`stsfld`, and reflection assignment from a non-declaring type; each control must fail if its corresponding scanner branch is removed. The production scan must reject a reflection assignment from any production caller (or implement reliable target-property dataflow), while preserving only the narrowly allowed constructor/default and own-setter stores.

Exact-head facts: GitHub OPEN/MERGEABLE; dcbTestsNet9, dcbTestsNet10, SonarCloud, and SonarCloud Code Analysis all SUCCESS. Focused net9 Dynamo/seam tests passed. `git diff --check` is clean. CoreGeneralSekibanExecutor remains 14+/5- both raw and ignore-EOL.

Design alignment remains checked against `.intent-cli/issues/SEK-G16/{packet.yaml,review-context.md,implementation.md,github-body.md}` and `intents/sekiban/design/storage-concurrency-primitives.md`; all other previously resolved G15/provider/Cosmos/facade/docs/legacy/no-P2 findings remain accepted.

tomohisa:
SEK-G16 minimal final rereview on `14958a43c706ff99b3937df18c2af9dd93998d44`: one narrowly scoped scanner blocker remains.

Resolved in this test-only delta:

- Reflection `PropertyInfo`/`FieldInfo.SetValue` calls are now scanned production-wide in each relevant assembly, with no production reflection assignments silently allowlisted.
- Setter and reflection controls are independently asserted from non-declaring fixtures.
- Production code is unchanged; the prior Dynamo/provider/Cosmos/facade/docs/G15/legacy/no-P2 conclusions remain accepted.

Blocking: backing-field matching still uses only the normalized field **name**, not the identity of the actual seam property/backing field. `FindSeamWrites` normalizes every resolved field and matches `seamNameSet.Contains(normalized)`. Therefore an unrelated field named `AfterConditionalCommitHook` in any type is treated as a seam write. The new `ExternalFieldStorer` control is exactly such an unrelated plain field and is asserted as a positive seam hit, so the test currently codifies the false-positive behavior that the delegation explicitly requires avoiding.

Resolve the actual target fields from the seam-declaring production properties/types (or otherwise bind field identity to the real seam member) and compare the resolved `FieldInfo` identity/declaring type, not a global name. Add two independent controls: (1) a real seam backing-field/direct-store fixture that proves `stfld`/`stsfld` detection, and (2) an unrelated same-named field fixture that is explicitly **not** reported. Keep own-setter/declaring-constructor allowances tied to that exact target field and declaring type.

Exact-head facts: GitHub OPEN/MERGEABLE; dcbTestsNet9, dcbTestsNet10, SonarCloud, and SonarCloud Code Analysis all SUCCESS. Focused architecture tests pass, but currently pass with the false-positive control above. Delta is exactly one test file; `git diff --check` is clean; CoreGeneralSekibanExecutor remains 14+/5- raw and ignore-EOL.

Design alignment remains checked against `.intent-cli/issues/SEK-G16/{packet.yaml,review-context.md,implementation.md,github-body.md}` and `intents/sekiban/design/storage-concurrency-primitives.md`.

tomohisa:
SEK-G16 minimal final rereview on `4dc2d28d78e62bb3681f8d0b24ac0e71bda37b63`: one coverage blocker remains.

The identity-matching repair itself is correct: configured properties resolve their actual getter `ldfld/ldsfld` target, stores compare exact `(Module, MetadataToken)` identities, missing targets fail closed, setter identity is exact, and the positive exact-field plus negative unrelated-same-name controls address the prior finding.

Blocking: the production scan does not target every actual settable seam, despite its own comment saying Postgres/SQLite/Cosmos declare `AfterConditionalCommitHook`.

- `PostgresEventStore` has an internal settable `AfterConditionalCommitHook` (`dcb/src/Sekiban.Dcb.Postgres/PostgresEventStore.cs`, used at lines 41 and 97-101), but `ProductionAssemblies_ContainNoSeamWrite_ByAnyPath`, `SeamStoreTypes`, constructor/options/public-surface guards, and the provider IVT checks cover only SQLite and Cosmos. The Postgres provider tests verify basic visibility/default state but do not run the new setter/stfld/reflection production scanner. Therefore one real production seam is outside the structural guarantee.
- The scanner treats production-wide reflection `SetValue` rejection as part of `FindSeamWrites`, but that method is never run over the Postgres or DynamoDB assemblies. Dynamo has no settable seam and should be excluded from target-field resolution, but a separate production-wide reflection scan must still cover every relevant production assembly if the claimed reflection ban is to be complete.

Add the Postgres hook to exact seam target scanning and the same API/constructor/IVT guards (in its provider test project if dependency boundaries require it). Decouple the assembly-wide reflection-assignment scan from seam target resolution so Postgres, SQLite, Cosmos, DynamoDB, and Core are all covered without inventing a missing Dynamo seam. Include explicit assertions enumerating the expected production seam inventory so a future seam cannot silently escape.

Exact-head facts: GitHub OPEN/MERGEABLE; dcbTestsNet9, dcbTestsNet10, SonarCloud, and SonarCloud Code Analysis all SUCCESS. Delta is one test file, `git diff --check` is clean, and CoreGeneralSekibanExecutor remains 14+/5- raw and ignore-EOL. All previously accepted production behavior remains unchanged.

Design alignment remains checked against `.intent-cli/issues/SEK-G16/{packet.yaml,review-context.md,implementation.md,github-body.md}` and `intents/sekiban/design/storage-concurrency-primitives.md`.

tomohisa:
SEK-G16 minimal final rereview on `298acbad287fb921b187ccd33be4c746db4071d0`: one inventory-exhaustiveness blocker remains.

Resolved in this test-only delta:

- Postgres now has equivalent exact-field/API/constructor/IVT/zero-write guards in `PostgresSeamArchitectureTests`.
- Reflection scanning is decoupled from seam resolution and currently executes over Core, Postgres, SQLite, Cosmos, and DynamoDB.
- Exact field identity, fail-closed resolution, independent setter/stfld/reflection controls, and the same-name negative control are preserved.

Blocking: `SeamInventory` is still a manually pinned six-row snapshot, not an exhaustive reverse-checked inventory as required.

- The only reverse discovery checks inspect `Func<Task>` properties on Postgres/SQLite/Cosmos provider store types. They do not inspect Core at all, even though three of the six current seams are `Func<Guid>`, `Func<string>`, and `TimeSpan?`. A newly added Core probe, a provider seam with a different delegate/value shape, a seam on another production type, or a Dynamo seam can remain absent from `SeamInventory` while the six-row snapshot and all current reverse guards continue to pass.
- `SeamInventory.ReflectionScannedAssemblies` is declared but unused by the tests. The actual scan coverage is split between a separate four-entry array and the Postgres test. Removing an assembly from either execution path does not compare against the authoritative five-assembly list, so coverage can silently drift.

Add a true reverse inventory assertion over the relevant production types/assemblies that discovers every supported seam shape (including all current Core allocator/budget probes, not only `Func<Task>` provider hooks) and proves each discovered `(assembly,type,property)` tuple is listed. Avoid merely asserting the current count. Also make reflection coverage report/compare the union of assemblies actually scanned against `SeamInventory.ReflectionScannedAssemblies`, so all five must execute exactly once or at least be evidenced as executed. Add a negative/mutation control showing an unlisted non-`Func<Task>` seam shape makes the reverse assertion fail.

Exact-head facts: GitHub OPEN/MERGEABLE; dcbTestsNet9, dcbTestsNet10, SonarCloud, and SonarCloud Code Analysis all SUCCESS. Delta is four test/test-support files, `git diff --check` is clean, and CoreGeneralSekibanExecutor remains 14+/5- raw and ignore-EOL. All previously accepted production/provider/facade/docs/G15/legacy/no-P2 behavior remains unchanged.

Design alignment remains checked against `.intent-cli/issues/SEK-G16/{packet.yaml,review-context.md,implementation.md,github-body.md}` and `intents/sekiban/design/storage-concurrency-primitives.md`.

tomohisa:
SEK-G16 strict final rereview completed on `a473c3318ca3fd9e4500b197fcac50b046482590`: no blocking findings remain.

Verified final repair:

- `ProductionSeamRule` reverse-discovers every non-public writable `Func<Task>`, `Func<Guid>`, `Func<string>`, and `TimeSpan?` property assembly-wide, and separately rejects non-backing mutable fields of those shapes.
- Discovered seam identities compare bidirectionally and exactly with `SeamInventory` for Core, Postgres, SQLite, Cosmos, and DynamoDB ownership. Four independent shape/new-declaring-type controls, an unlisted non-`Func<Task>` equality-failure control, and a hidden-field control make discovery non-vacuous.
- Exact getter-field identity, setter identity, direct `stfld/stsfld`, external reflection, fail-closed missing target, and unrelated-same-name negative controls remain intact.
- Split project ownership is pinned: WithResult scans Core/SQLite/Cosmos/Dynamo and delegates exactly Postgres; Postgres scans itself and delegates exactly the other four. Each side asserts scanned + delegated equals the authoritative five-assembly list. Reflection `SetValue` scanning therefore executes across all five, including Dynamo despite no seam.
- Postgres has equivalent exact field/API/constructor/IVT/production-zero guards; all six actual seams are covered.

Independent verification: focused net9 architecture suites passed (WithResult 23/23; Postgres 9/9). GitHub exact head is OPEN/MERGEABLE; dcbTestsNet9, dcbTestsNet10, SonarCloud, and SonarCloud Code Analysis are all SUCCESS. Delta is three test/test-support files; `git diff --check` is clean; CoreGeneralSekibanExecutor remains 14+/5- raw and ignore-EOL.

All prior G16 provider/ambiguity/corruption/facade/version-first/capability/docs findings remain resolved. G15 identity/fingerprint/receipt contract, legacy paths, and no-P2 scope remain unchanged.

Design alignment checked against `.intent-cli/issues/SEK-G16/{packet.yaml,review-context.md,implementation.md,github-body.md}` and `intents/sekiban/design/storage-concurrency-primitives.md`.

Review result: completed — no blocking findings. GitHub self-approval may be unavailable because the authenticated reviewer is the PR author; route to the standing design-owned admin merge fallback. Review does not merge.

PR #1118, opened 2026-08-10
feat(dcb): gate materialized view activation on authoritative eligibility

### LINKED ISSUE #1113: SEK-G27 dcb materialized view: gate active-version cutover on authoritative eligibility
# SEK-G27 dcb materialized view: gate active-version cutover on authoritative eligibility

## Goal

Allow an MV version to become active only after authoritative evidence proves
that it reached a captured target checkpoint and is safe. Initialization and
absence of an existing active pointer are never eligibility.

## Why This Slice Exists Now

SekibanCloud requires safe generation activation. This upstream request did not
come from an existing GitHub issue or SekibanWasmRuntime work. It follows
SEK-G26 because activation cannot be safe without explicit Known/Unknown truth.

## Current Observed State

All four classic MV executors call `SetActiveAsync` during initialization when
no active row exists. `IMvRegistryStore.SetActiveAsync` is unconditional and
has no expected-current version, target capture, or eligibility result. G24's
`IsCaughtUp` is best-effort monitoring and cannot authorize cutover.

- Capture an authoritative target as explicit Known truth before judging.
- Eligibility requires matching identity, Known target and at/through position,
  safe lifecycle, and current generation/provenance.
- Remove initialization-time auto-activation from all four executors.
- Switch only when expected active/generation and eligibility still match in
  the provider transaction; conflicts leave the former pointer unchanged.
- Initial activation uses the same eligibility path.

## Target Repo / Path / Part

Target paths: `dcb/src dcb/tests docs/dcb_llm docs/dcb_llm_ja`

Target part: eligibility contract, target capture, registry active-pointer
operation, four-provider transactions, race tests, and docs.

## In Scope

- Eligibility contract/result and authoritative target-capture seam.
- `IMvRegistryStore` active-pointer contract.
- Postgres, MySQL, SQL Server, and SQLite stores/executors.
- Atomicity/race tests and additive cutover operation API.
- EN/JA documentation.

## Out Of Scope

- Parallel-generation scheduling and active-version reader routing.
- Rollback workflow orchestration and G24 heartbeat publication.
- Cloud management APIs.

Consume SEK-G26 checkpoint truth to capture an authoritative target, evaluate a
candidate fail-closed, remove initialization auto-activation, and atomically
compare-and-switch the active pointer in all four providers with typed conflict
and ineligibility results.

## Acceptance Criteria

- Behind, Unknown, unsafe, faulted, wrong-identity, stale-target, and
  stale-expected-active candidates return typed failure with zero mutation.
- An eligible candidate atomically switches in all four providers; bypassing
  eligibility fails the common provider suite.
- Concurrent attempts against one expected generation cannot both commit.
- Initial activation with no pointer still requires Known target/position and
  safe state; restoring initialization auto-activation fails a test.
- Provider failure preserves the former pointer and retries are deterministic.
- A killing test proves G24 best-effort status is not authoritative evidence.
- Compatibility, affected tests, EN/JA docs, and `git diff --check` pass.

## Verification

Run the four-provider eligibility/atomicity suite, race and failure tests,
initial-activation mutation, G24-substitution killing test, compatibility checks,
affected DCB tests, docs checks, and `git diff --check`.

## Related Links

- Dependency: [SEK-G26 #1112](https://github.com/J-Tech-Japan/Sekiban/issues/1112).
- Source handoff: `reports/sekiban-matview-upstream-handoff-20260808.md`

## Knowledge Maintenance

- Intent/state documentation: required.
- ADR candidate: none unless provider transaction semantics require one.
- Docs update and closeout writeback: required.

## Guide Reachability

No intent-cli/operator role-facing command is added.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1113. Adds authoritative target capture, typed fail-closed eligibility, four-provider expected active/generation CAS, no initialization auto-activation, initial activation through the same boundary, compatibility surfaces, race/zero-mutation tests, and EN/JA docs. Verification: full dcb solution net9/net10, MaterializedView unit/integration/Postgres suites, git diff --check.

### DIFF
--- a/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvExecutor.cs
+++ b/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvExecutor.cs
@@ -53,11 +53,16 @@ public sealed class MySqlMvExecutor : MvExecutorBase<MySqlConnection>, IMvExecut
         return await CatchUpFromStoreAsync(
                 host,
                 exactServiceId,
-                _eventStoreFactory is null ? _legacyEventStore : _eventStoreFactory!.CreateForService(exactServiceId),
+                SelectEventStoreForService(exactServiceId),
                 _eventStoreFactory is not null,
                 cancellationToken)
             .ConfigureAwait(false);
     }
+
+    protected override IEventStore SelectEventStoreForService(string exactServiceId) =>
+        _eventStoreFactory is not null
+            ? _eventStoreFactory.CreateForService(exactServiceId)
+            : _legacyEventStore!;
 
     protected override async Task<MySqlConnection> OpenConnectionAsync(CancellationToken cancellationToken)
     {
--- a/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvRegistryStore.cs
+++ b/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvRegistryStore.cs
@@ -49,6 +49,7 @@ public sealed class MySqlMvRegistryStore : IMvRegistryStore
                 service_id VARCHAR(200) NOT NULL,
                 view_name VARCHAR(200) NOT NULL,
                 active_version INT NOT NULL,
+                active_generation BIGINT NOT NULL DEFAULT 0,
                 activated_at DATETIME(6) NOT NULL,
                 PRIMARY KEY (service_id, view_name)
             );
@@ -85,6 +86,25 @@ public sealed class MySqlMvRegistryStore : IMvRegistryStore
         }
 
         await connection.ExecuteAsync(new CommandDefinition(activeSql, cancellationToken: cancellationToken)).ConfigureAwait(false);
+        var activeGenerationColumns = (await connection.QueryAsync<string>(
+                new CommandDefinition(
+                    """
+                    SELECT column_name
+                    FROM information_schema.columns
+                    WHERE table_schema = DATABASE()
+                      AND table_name = 'sekiban_mv_active'
+                      AND column_name = 'active_generation';
+                    """,
+                    cancellationToken: cancellationToken)))
+            .ToHashSet(StringComparer.OrdinalIgnoreCase);
+        if (!activeGenerationColumns.Contains("active_generation"))
+        {
+            await connection.ExecuteAsync(
+                    new CommandDefinition(
+                        "ALTER TABLE sekiban_mv_active ADD COLUMN active_generation BIGINT NOT NULL DEFAULT 0;",
+                        cancellationToken: cancellationToken))
+                .ConfigureAwait(false);
+        }
     }
 
     public async Task RegisterAsync(MvRegistryEntry entry, IDbTransaction? transaction = null, CancellationToken cancellationToken = default)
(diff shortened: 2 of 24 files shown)


tomohisa:
Request update — SEK-G27 semantic review at full head `c3f370aea61adf4c041f514701f06dd3e8ae8a3e`.

1. **The provider CAS does not fence the candidate snapshot through the complete transaction.** The packet requires eligible compare-and-switch to be atomic in all four providers, concurrent or stale attempts to leave the former pointer unchanged, and provider failure to preserve the former pointer. In Postgres `TryActivateInTransactionAsync`, the first statement validates the candidate snapshot while changing `sekiban_mv_active` (lines 454-530), then `markActiveSql` updates the registry rows (lines 532-544), but its affected-row count is ignored and the method always returns success. MySQL lines 491-582, SQL Server lines 485-592, and SQLite lines 519-640 have the same two-step/result-discard shape. On providers where a registry row can change after the first statement read and before the second update, the second update can affect zero rows while the active pointer still commits. Lock/fence the candidate rows for the whole operation or otherwise make a candidate mismatch roll back every pointer mutation, including the overload that receives a caller transaction. Assert that exactly `CandidateCount` rows transition before returning success.

2. **The required four-provider race, failure, and zero-mutation proof is absent.** The only new real store CAS test is `SqliteRegistry_UsesExpectedGenerationAndLeavesPointerUnchangedForInvalidOrStaleRequests`, and it performs two sequential calls against SQLite. The common multi-provider diff only asserts that initialization leaves the active pointer null. There is no concurrent same-generation test with exactly one winner, candidate-change interleaving test, all-four-provider invalid/bypass zero-mutation suite, or provider-failure plus deterministic-retry proof. Also, `BehindCandidate_IsRejectedEvenWhenASeparateBestEffortStatusWouldSayCaughtUp` assigns an unrelated local boolean; no production G24 observation reaches the activation path, so this is not the packet-required killing test. Add substantive common-provider tests covering Unknown, unsafe/faulted, wrong identity, stale target/expected pointer, actual concurrent winners, rollback/pointer preservation, initialization mutation, and a wired favorable G24 observation that still cannot authorize activation without authoritative truth.

Sonar triage: the 35.1% duplication is confined to the waived four-provider same-shape store/executor class (provider-neutral activation/contracts/base and Orleans grain report 0 duplicated new lines). No open vulnerability exists and security/reliability/maintainability ratings are A. The S8970 findings are non-semantic null-forgiving cleanup at constructor/transaction/host invariants. S3776 on `MvActivationEligibility.Evaluate` is a real maintainability signal for this authorization boundary (complexity 30); the straight fail-closed guards are readable, but the current branch coverage is far too sparse. Extracting per-entry checks and testing every typed rejection is the natural repair and should accompany finding 2.

Confirmed: initialization itself no longer calls `SetActiveAsync`; authoritative target capture and the typed evaluator exist; expected active version plus generation are carried; G24 status is absent from production authorization; EN/JA docs are present; no cloud/coordinator/rollback/heartbeat scope widening was found; and the PR body correctly closes #1113. Design alignment checked against SEK-G27 `packet.yaml`, `implementation.md`, `review-context.md`, `github-body.md`, issue #1113 In/Out of Scope and AC, `intents/sekiban/intent-tree/00-map.md`, and `intents/sekiban/design/materialized-view-service-scope.md`. Focused net9/net10 tests passed 6/6 each, but tests are necessary and not sufficient.

tomohisa:
SEK-G27 rereview at exact head `a7ea9ac987c79b78d179e4462a796c6c8845c2e8`: one implementation finding remains.

**Packet AC: A killing test proves G24 best-effort status is not authoritative evidence.** `FavorableWiredG24Observation_CannotAuthorizeBehindCandidate` now creates a real production `ProjectionStatusReader` observation with `IsCaughtUp == true`, but that observation is discarded before authorization: the test then calls `MvActivationEligibility.Evaluate` directly with a behind registry row. No executor or Orleans activation entry point receives or can observe the G24 result. Consequently, a mutation that makes `MvExecutorBase` or `MaterializedViewGrain` authorize cutover from favorable G24 status would not fail this test; only the independent static evaluator remains tested. This is improved setup, but it is still not the required production-path killing proof.

Required unblock: compose a real activation path (provider executor or actual grain/catch-up completion) in the same production-style container as the favorable G24 status, invoke that path with Unknown/behind authoritative truth, and assert the typed rejection plus zero active-pointer mutation. An architecture/dependency assertion may supplement this, but the test must fail if production activation starts treating G24 observation as authorization.

The original provider-CAS blocker is resolved: all four stores lock/fence candidate rows, compare the exact snapshot, require `marked == CandidateCount`, and roll local or caller-owned operations back through provider-native savepoints. The common suite is substantive and locally passed 24/24 with zero skips on both net9.0 and net10.0, covering same-generation one-winner concurrency, candidate interleaving, invalid/bypass zero mutation, provider failure/retry, caller savepoint rollback, and trigger-forced mark-count mismatch rollback. `MvActivationEligibility.Evaluate` is decomposed into per-entry/active-pointer checks; every evaluator rejection is covered, S3776 is gone, and only S8970 minor cleanup remains. Sonar reports vulnerabilities 0 and A security/reliability/maintainability; 29.5% duplication remains confined to the standing-waiver four-provider store/executor shape.

All other packet AC and prior findings remain satisfied by semantic inspection; no scheduling/routing/rollback/G24-publication/cloud scope widening was found, EN/JA docs remain aligned, `git diff --check` passes, and `Closes #1113` is correct. design_alignment_checked=true: SEK-G27 `packet.yaml`, `implementation.md`, `review-context.md`, `github-body.md`, issue #1113 In/Out of Scope and AC, prior review comment 5245149411, `intents/sekiban/intent-tree/00-map.md`, and `intents/sekiban/design/materialized-view-service-scope.md`. Tests are necessary but not sufficient.

PR #1127, opened 2026-08-14
feat(dcb): add verify-only materialized view initialization

### LINKED ISSUE #1125: SEK-G32 dcb materialized view: add verify-only initialization and host-owned SQL statement policy
# SEK-G32 dcb materialized view: add verify-only initialization and host-owned SQL statement policy

## Origin

This is a consumer-side upstream request from a SekibanCloud design pass. The
receiving Sekiban repository owns the API shape, implementation, and release
timing. The capability is requested on its own merits for self-hosted Sekiban
users: the user's database is the user's property.

## Goal

Let a host start a materialized view against a pre-provisioned target without
creating or altering schema, and let the host approve or reject every
projector-supplied SQL statement before a provider executes it.

## Why This Slice Exists Now

Sekiban's materialized-view execution model is now service-scoped, observable,
and generation-aware, but initialization still assumes the runtime may ensure
infrastructure and execute initialization SQL. Self-hosted and bring-your-own-
database deployments often deliberately deny DDL privileges and require schema
ownership to stay with the database owner. That is a general framework need,
not a cloud-specific exception.

## Current Observed State

Grounded against `J-Tech-Japan/Sekiban` `origin/main` at
`1ffdf7d5fad9065e60d13d87b69a017dbd4373bd`:

- `MvExecutorBase.InitializeCoreAsync` always calls
  `IMvRegistryStore.EnsureInfrastructureAsync`, opens a transaction, invokes
  `IMvApplyHost.InitializeAsync`, and executes every returned statement.
- Postgres, MySQL, SQL Server, and SQLite registry stores create framework
  tables from `EnsureInfrastructureAsync`; provider initializers also emit DDL.
- `IMvApplyHost.InitializeAsync` and `ApplyEventAsync` return SQL statement
  batches that reach provider execution without a host-selectable policy verdict.
- There is no explicit verify-only mode that proves a pre-provisioned schema is
  compatible while guaranteeing zero DDL.

No duplicate open or closed Sekiban issue was found for this capability on
2026-08-12.

- The target database and its schema belong to the self-hosted user. The
  framework may mutate schema only under an explicit caller-selected mode.
- Preserve current create/ensure behavior as an explicitly named compatibility
  mode; do not silently make it verify-only and do not let verify-only fall back.
- Verify-only may use provider metadata/capability probes, but it performs no
  CREATE, ALTER, DROP, migration, registry auto-create, or projector DDL.
- A missing or incompatible required object is a typed closed failure before
  event read, view write, registry mutation, catch-up, or activation.
- The statement-policy contract is provider-neutral and host-selectable. It
  judges projector-supplied initialization and apply statements before provider
  execution with exact phase and service/view/version/table context.
- This request defines the capability and safety boundary, not a particular SQL
  parser or public type name.

## Target Repo / Path / Part

Target paths: `dcb/src/Sekiban.Dcb.MaterializedView*`,
`dcb/tests/Sekiban.Dcb.MaterializedView*`, `docs/dcb_llm`, `docs/dcb_llm_ja`

Target part: provider-neutral initialization modes, schema/capability
verification, pre-execution SQL policy, provider parity, tests, and docs.

## In Scope

- Explicit create/ensure and verify-only initialization modes.
- Verification of required framework registry objects, view bindings, and
  provider capabilities without schema mutation.
- A stable pre-execution statement-policy seam for initialization/apply batches.
- Typed, secret-safe failure reasons and zero-side-effect rejection.
- Postgres, MySQL, SQL Server, and SQLite behavioral parity.
- Additive compatibility, EN/JA documentation, and decisive mutations.

## Out Of Scope

- WASM ABI, artifact digest, manifest/metadata matching, or WASM query callbacks.
- SekibanCloud management APIs, UI, secret custody, schema provisioning, or
  migrations.
- Removing the explicit create/ensure compatibility mode.
- Prescribing one SQL parser, policy language, or database ownership model.

Add a provider-neutral verify-only materialized-view initialization path that
can validate a compatible pre-provisioned target while issuing zero DDL, and a
host-owned policy gate that sees and decides every projector-supplied
initialization/apply SQL statement before provider execution. Keep create/ensure
as an explicit compatible mode, fail closed before side effects, prove parity
across all four relational providers, and document the ownership boundary.

## Acceptance Criteria

- A compatible pre-provisioned target initializes in verify-only mode while the
  runtime principal has DDL denied; recorded CREATE/ALTER/DROP/migration counts
  are zero.
- Missing framework registry objects, missing view objects, incompatible
  columns/types/keys, or unsupported provider capabilities return typed failure
  before event read, write, registry mutation, catch-up, or activation.
- Verify-only never enters `EnsureInfrastructureAsync`, executes projector
  initialization DDL, or falls back to create/ensure. Mutations restoring any of
  those paths fail decisive tests.
- Create/ensure remains explicitly selectable and retains compatible behavior.
- Every projector-supplied statement from initialization and apply batches is
  presented to the configured policy with phase plus exact service, view,
  version, logical/physical table bindings, SQL, and parameter metadata before
  provider execution.
- A rejected statement produces a typed safe result and zero provider command
  executions; bypassing the policy, mislabelling the phase, or checking after
  execution drives tests red.
- The common contract runs for Postgres, MySQL, SQL Server, and SQLite.
- Public API changes are additive or carry an explicit compatibility plan;
  focused/affected suites, compatibility checks, docs, and `git diff --check`
  pass.

## Verification

Run the common four-provider initialization/policy suite, a real or faithful
DDL-denied compatible-schema test, missing/drifted-schema tests, policy rejection
zero-I/O tests, the create/ensure compatibility suite, public API checks,
affected DCB tests, EN/JA documentation checks, and `git diff --check`.

## Related Links

- Consumer materialized-view intent:
  `intents/sekiban-as-a-service/intent-tree/means/07-projection-and-view-strategy.md`
- Consumer execution dependency:
  `intents/sekiban-as-a-service/execution/08-bring-your-own-domain-goal.md`
- Companion WASM-host request:
  [SWR-G079 #267](https://github.com/J-Tech-Japan/SekibanWasmRuntime/issues/267).

## Knowledge Maintenance

- Intent placement: `intents/sekiban/intent-tree/00-map.md`.
- ADR candidate: only if implementation discovers a public compatibility decision.
- Diagram candidate: none required.
- Docs update: required in `docs/dcb_llm` and `docs/dcb_llm_ja`.
- Closeout writeback expected: yes; record final seams and provider parity.

## Guide Reachability

No `intent-cli` or operator role-facing command is added by this library slice.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1125

## Implementation

- Completed the single VerifyOnly gate before projector initialization. VerifyOnly derives bindings only from the declarative contract, never calls `host.InitializeAsync`, and never enters ensure, normal open, registration, transaction, commit, target capture persistence, status mutation, activation, stream subscription, refresh timers, or event-apply writes.
- Closed the remaining implicit paths: target capture, empty-history catch-up/checkpoint capture, status/activation, stream event apply, Orleans grain orchestration, and generation-coordinator forced reverse all fail closed before registry mutation. VerifyOnly event apply is an inspection no-op after the schema gate and cannot open a normal target connection or transaction.
- Made each provider's `IMvReadOnlyMvInspector` implementation own the read-only route: SQLite `Mode=ReadOnly` and table-valued PRAGMA catalog functions; PostgreSQL `default_transaction_read_only=on`; MySQL read-only transaction session; SQL Server an explicit least-privilege inspection connection.
- SQL Server no longer treats `ApplicationIntent=ReadOnly` as enforcement. `MvOptions.SqlServerInspectionConnectionString` is required for VerifyOnly. The inspector opens that connection, verifies a distinct principal with database `SELECT` and `VIEW DEFINITION`, rejects any database/object write permission, and returns typed `UnsupportedProviderCapability` before catalog inspection when the capability cannot be established. Legacy writable paths remain unchanged.
- Extended `MvSchemaContract`/requirements and deterministic verification to defaults, required indexes (ordered columns and uniqueness), generated semantics/expressions, size, and precision/scale with typed fail-closed mismatch diagnostics.
- Removed the SQLite S2077 dynamic SQL hotspot with static catalog SQL, parameterized table-valued PRAGMAs, strict contract-derived identifier validation, and faithful derivation of declared lengths, precision/scale, and generated expressions. Required dimensions that SQLite cannot faithfully inspect return precise typed `UnsupportedProviderCapability` failures.
- Preserved the enforced SQL policy boundary and enriched its context/failure with exact origin, provider, rule id, phase, batch/statement index, and SQL fingerprint. Query rows/single/scalar, raw bypass, mutating CTE/comment/multi-statement shapes, complete batch preflight, fault, and cancellation remain covered.
- Added the EN/JA documentation updates and retained the binary consumer fixture restored against the published `Sekiban.Dcb.MaterializedView` `10.13.0` package.

## Validation

Production-path real-store VerifyOnly/policy and schema matrix:

- `dotnet test dcb/tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests.csproj --no-build --no-restore -f net9.0 --filter 'FullyQualifiedName~MvVerifyOnly|FullyQualifiedName~SchemaMatrix'` — **31/31 passed**.
- Same focused production-path command with `-f net10.0` — **31/31 passed**.
- The four-provider real-store proof starts schema creation and ordinary writes with the normal write-capable fixture connection. VerifyOnly uses only the production inspector route; catalog inspection commands are recorded as reads, projector initialization and SQL policy are not invoked, and registry entries before/after are equal. The SQL Server test uses a separate inspector principal with `db_datareader` plus non-writing `VIEW DEFINITION`; an actual INSERT through that connection is rejected with SQL error 229 and the real table row count is unchanged.
- The production-inspector mismatch matrix runs against real PostgreSQL, MySQL, SQL Server, and SQLite stores. For every provider it proves compatible pass, isolated default mismatch (`DefaultMismatch`), ordered/unique index mismatch (`RequiredIndexMissing`), generated semantics mismatch (`GeneratedSemanticsMismatch`), width mismatch (`SizeMismatch`), precision mismatch (`PrecisionMismatch`), and scale mismatch (`PrecisionMismatch`). SQLite additionally proves an unbounded declared type requested with a bounded contract returns `UnsupportedProviderCapability` with the precise unsupported-dimension message.
- SQLite real-store VerifyOnly separately exercises successful initialization, target capture, catch-up with a non-authoritative target, activation eligibility, status/checkpoint observation, and event apply. Rejecting registry counters fail on every mutating API; all counters remain zero and the target remains unchanged after catch-up. The VerifyOnly host initialization count is zero.

Affected MV and compatibility suites:

- `dotnet test dcb/tests/Sekiban.Dcb.MaterializedView.Tests/Sekiban.Dcb.MaterializedView.Tests.csproj --no-build --no-restore -f net9.0` — **78/78 passed**.
- Same core MV command with `-f net10.0` — **78/78 passed**.
- `dotnet test dcb/tests/Sekiban.Dcb.MaterializedView.Postgres.Tests/Sekiban.Dcb.MaterializedView.Postgres.Tests.csproj --no-build --no-restore -f net9.0 --filter FullyQualifiedName~MaterializedViewPostgresTests` — **7/7 passed**.
- Same Postgres real-store command with `-f net10.0` — **7/7 passed**.
- Binary consumer: compiled against the actual published `10.13.0` package and ran without source recompilation: `binary-consumer-ok:binary-consumer:Sekiban.Dcb.MaterializedView.IMvExecutor`.
- `git diff --check` and EN/JA documentation contract checks passed.

## Mutation-kill evidence

Each decisive mutation was applied to the production path, the named test was run to failure, and the original implementation was restored before the final runs:

- Removing VerifyOnly target-capture and catch-up checkpoint guards failed at the registry update (`SetTargetCheckpointAsync`). Removing activation or event-apply guards failed the zero-mutation counters, proving normal target open/transaction/apply was reintroduced.
- Reintroducing SQL Server's writable-connection `ApplicationIntent=ReadOnly` fallback made all three SQL Server production tests fail at the restricted capability check; this kills the fake-read-only route.
- Removing SQLite generation-expression mapping failed with `SQLite pragma metadata identifies a generated column, but its generation expression could not be derived from sqlite_master.` Removing width mapping failed with `SQLite cannot faithfully inspect the required declared character or binary length.` Removing precision/scale mapping failed with `SQLite cannot faithfully inspect the required declared numeric precision and scale.`
- Removing SQLite default mapping failed with the expected `DefaultMismatch`. Reversing PostgreSQL catalog index ordinality failed the real-store matrix with `RequiredIndexMissing`.
- Bypassing enforced raw/query policy, phase/origin labeling, and first-statement-only/post-execution batch checking fails the corresponding policy tests. All mutations were restored.

The unfiltered Orleans-hosted Postgres run was attempted separately; its environment-only Orleans `TestClusterPortAllocator` path hit `System.OverflowException` in `BsdIPGlobalProperties.GetTcpConnections` while allocating TCP listener pairs. The affected production-path real-store filters above pass independently.

Exact head: `7c5cb72f6dd15d15eedcc44da164962accdfb1a7`

## Final rereview round 2 matrix repair

- Added a real non-unique index to every provider schema fixture and added it to the declarative contract as IsUnique=false. The matching contract passes; an isolated IsUnique=true requirement returns exactly RequiredIndexMissing. The existing ordered-column mismatch remains a separate exact RequiredIndexMissing case.
- Replaced the constant generated expression with a non-constant expression derived from base_value for PostgreSQL, MySQL, SQLite, and SQL Server (ISNULL(base_value + 1, 0) preserves SQL Server computed-column nullability). The matching contract passes; changing only the expression to base_value + 2 returns exactly GeneratedSemanticsMismatch.
- Both new cases use the production IMvReadOnlyMvInspector against real PostgreSQL, MySQL, SQL Server, and SQLite stores. The shared assertion checks the complete deterministic code list in both result.Mismatches and result.Failure.Mismatches. SQL Server expected default/expression values now come from the real system catalog; SQLite keeps faithful expression derivation.
- Final matrix result: SchemaContractMatrix_UsesTheProductionInspector 4/4 passed on both net9.0 and net10.0; focused MV/policy/schema suite 31/31 passed on both frameworks.

## Final matrix mutation-kill evidence

Each mutation was applied to the provider production inspector, the real-store matrix test was run to red, and the mutation was restored before final validation:

- PostgreSQL, MySQL, SQL Server, and SQLite uniqueness observation forced to true: each failed with the exact RequiredIndexMissing for ix_schema_matrix_nonunique.
- PostgreSQL, MySQL, SQL Server, and SQLite generation-expression observation forced to literal 7: each failed with the exact incompatible generation-expression / GeneratedSemanticsMismatch result.

Final validation after restoration: core MV 78/78 on net9.0 and net10.0; Postgres real-store affected suite 7/7 on net9.0 and net10.0; git diff --check passed. The final pushed head is 4cbf1f6801e42743878167a1e2cf19cbdf374d34.

### DIFF
--- a/dcb/internalUsages/Dcb.Domain.WithoutResult/MaterializedViews/OrderSummaryMvV1.cs
+++ b/dcb/internalUsages/Dcb.Domain.WithoutResult/MaterializedViews/OrderSummaryMvV1.cs
@@ -4,13 +4,44 @@ using Sekiban.Dcb.MaterializedView;
 
 namespace Dcb.Domain.WithoutResult.MaterializedViews;
 
-public sealed class OrderSummaryMvV1 : IMaterializedViewProjector
+public sealed class OrderSummaryMvV1 : IMaterializedViewProjector, IMvSchemaRequirementsProvider
 {
     public string ViewName => "OrderSummary";
     public int ViewVersion => 1;
 
     public MvTable Orders { get; private set; } = default!;
     public MvTable Items { get; private set; } = default!;
+
+    public IReadOnlyList<MvSchemaTableRequirement> GetSchemaRequirements(
+        MvDbType databaseType,
+        IMvTableBindings tables) =>
+    [
+        new MvSchemaTableRequirement(
+            "orders",
+            tables.GetPhysicalName("orders"),
+            [
+                new("id", MvSchemaTypeFamily.Guid, false),
+                new("status", MvSchemaTypeFamily.String, false),
+                new("total", MvSchemaTypeFamily.Decimal, false),
+                new("created_at", MvSchemaTypeFamily.DateTime, false),
+                new("_last_sortable_unique_id", MvSchemaTypeFamily.String, false),
+                new("_last_applied_at", MvSchemaTypeFamily.DateTime, false)
+            ],
+            ["id"]),
+        new MvSchemaTableRequirement(
+            "items",
+            tables.GetPhysicalName("items"),
+            [
+                new("id", MvSchemaTypeFamily.Guid, false),
+                new("order_id", MvSchemaTypeFamily.Guid, false),
+                new("product_name", MvSchemaTypeFamily.String, false),
+                new("quantity", MvSchemaTypeFamily.Integer, false),
+                new("unit_price", MvSchemaTypeFamily.Decimal, false),
+                new("_last_sortable_unique_id", MvSchemaTypeFamily.String, false),
+                new("_last_applied_at", MvSchemaTypeFamily.DateTime, false)
+            ],
+            ["id"])
+    ];
 
     public async Task InitializeAsync(IMvInitContext ctx, CancellationToken cancellationToken = default)
     {
--- a/dcb/internalUsages/Dcb.Domain.WithoutResult/MaterializedViews/WeatherForecastMvV1.cs
+++ b/dcb/internalUsages/Dcb.Domain.WithoutResult/MaterializedViews/WeatherForecastMvV1.cs
@@ -4,12 +4,32 @@ using Sekiban.Dcb.MaterializedView;
 
 namespace Dcb.Domain.WithoutResult.MaterializedViews;
 
-public sealed class WeatherForecastMvV1 : IMaterializedViewProjector
+public sealed class WeatherForecastMvV1 : IMaterializedViewProjector, IMvSchemaRequirementsProvider
 {
     public string ViewName => "WeatherForecast";
     public int ViewVersion => 1;
 
     public MvTable Forecasts { get; private set; } = default!;
+
+    public IReadOnlyList<MvSchemaTableRequirement> GetSchemaRequirements(
+        MvDbType databaseType,
+        IMvTableBindings tables) =>
+    [
+        new MvSchemaTableRequirement(
+            "forecasts",
+            tables.GetPhysicalName("forecasts"),
+            [
+                new("forecast_id", MvSchemaTypeFamily.Guid, false),
+                new("location", MvSchemaTypeFamily.String, false),
+                new("forecast_date", MvSchemaTypeFamily.DateTime, false),
+                new("temperature_c", MvSchemaTypeFamily.Integer, false),
+                new("summary", MvSchemaTypeFamily.String, true),
+                new("is_deleted", MvSchemaTypeFamily.Boolean, false),
+                new("_last_sortable_unique_id", MvSchemaTypeFamily.String, false),
+                new("_last_applied_at", MvSchemaTypeFamily.DateTime, false)
+            ],
+            ["forecast_id"])
+    ];
 
     public async Task InitializeAsync(IMvInitContext ctx, CancellationToken cancellationToken = default)
     {
--- a/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvExecutor.cs
+++ b/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvExecutor.cs
@@ -17,6 +17,8 @@ public sealed class MySqlMvExecutor : MvExecutorBase<MySqlConnection>, IMvExecut
     private readonly IEventStore? _legacyEventStore;
     private readonly IServiceIdProvider? _legacyServiceIdProvider;
 
+    protected override MvDbType DatabaseType => MvDbType.MySql;
+
     /// <summary>
     /// Creates the legacy single-service compatibility executor over an aggregate event store.
     /// Multi-service hosts should use the <see cref="IEventStoreFactory"/> constructor.
--- a/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvRegistryStore.cs
+++ b/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvRegistryStore.cs
@@ -8,6 +8,7 @@ public sealed partial class MySqlMvRegistryStore : MvForcedReverseRegistryStoreB
 {
     private readonly string _connectionString;
     private readonly Action<string>? _catalogCommandRecorder;
+    private readonly Action<string>? _readOnlyConnectionRecorder;
 
     public MySqlMvRegistryStore(string connectionString)
         : this(connectionString, null)
@@ -15,10 +16,28 @@ public sealed partial class MySqlMvRegistryStore : MvForcedReverseRegistryStoreB
     }
 
     public MySqlMvRegistryStore(string connectionString, Action<string>? catalogCommandRecorder)
+        : this(connectionString, catalogCommandRecorder, null)
+    {
+    }
+
+    public MySqlMvRegistryStore(
+        string connectionString,
+        Action<string>? catalogCommandRecorder,
+        Action<string>? readOnlyConnectionRecorder)
     {
         _connectionString = connectionString;
         _catalogCommandRecorder = catalogCommandRecorder;
+        _readOnlyConnectionRecorder = readOnlyConnectionRecorder;
     }
+
+    private string ReadOnlyConnectionString => _connectionString;
+
+    private void RecordReadOnlyConnection() => _readOnlyConnectionRecorder?.Invoke("mysql:transaction_read_only=on");
+
+    private static async Task SetReadOnlySessionAsync(MySqlConnection connection, CancellationToken cancellationToken) =>
+        await connection.ExecuteAsync(
+                new CommandDefinition("SET SESSION TRANSACTION READ ONLY;", cancellationToken: cancellationToken))
+            .ConfigureAwait(false);
 
     private CommandDefinition CatalogCommand(
         string sql,
@@ -391,11 +410,20 @@ public sealed partial class MySqlMvRegistryStore : MvForcedReverseRegistryStoreB
             cancellationToken).ConfigureAwait(false);
     }
 
-    public async Task<IReadOnlyList<MvRegistryEntry>> GetEntriesAsync(
+    public Task<IReadOnlyList<MvRegistryEntry>> GetEntriesAsync(
         string serviceId,
         string viewName,
         int viewVersion,
-        CancellationToken cancellationToken = default)
+        CancellationToken cancellationToken = default) =>
+        ReadEntriesWithConnectionAsync(_connectionString, readOnly: false, serviceId, viewName, viewVersion, cancellationToken);
+
+    private async Task<IReadOnlyList<MvRegistryEntry>> ReadEntriesWithConnectionAsync(
+        string connectionString,
+        bool readOnly,
+        string serviceId,
+        string viewName,
+        int viewVersion,
+        CancellationToken cancellationToken)
     {
         const string sql = """
             SELECT service_id AS ServiceId,
@@ -425,8 +453,12 @@ public sealed partial class MySqlMvRegistryStore : MvForcedReverseRegistryStoreB
             ORDER BY logical_table;
             """;
 
-        await using var connection = new MySqlConnection(_connectionString);
+        await using var connection = new MySqlConnection(connectionString);
         await connection.OpenAsync(cancellationToken).ConfigureAwait(false);
+        if (readOnly)
+        {
+            await SetReadOnlySessionAsync(connection, cancellationToken).ConfigureAwait(false);
+        }
         var rows = await connection.QueryAsync(
             CatalogCommand(sql, new { ServiceId = serviceId, ViewName = viewName, ViewVersion = viewVersion }, cancellationToken))
             .ConfigureAwait(false);
@@ -437,13 +469,24 @@ public sealed partial class MySqlMvRegistryStore : MvForcedReverseRegistryStoreB
         string serviceId,
         string viewName,
         int viewVersion,
-        CancellationToken cancellationToken = default) =>
-        GetEntriesAsync(serviceId, viewName, viewVersion, cancellationToken);
+        CancellationToken cancellationToken = default)
+    {
+        RecordReadOnlyConnection();
+        return ReadEntriesWithConnectionAsync(ReadOnlyConnectionString, readOnly: true, serviceId, viewName, viewVersion, cancellationToken);
+    }
 
-    public async Task<MvActiveEntry?> GetActiveAsync(
+    public Task<MvActiveEntry?> GetActiveAsync(
         string serviceId,
         string viewName,
-        CancellationToken cancellationToken = default)
+        CancellationToken cancellationToken = default) =>
+        ReadActiveWithConnectionAsync(_connectionString, readOnly: false, serviceId, viewName, cancellationToken);
+
+    private async Task<MvActiveEntry?> ReadActiveWithConnectionAsync(
+        string connectionString,
+        bool readOnly,
+        string serviceId,
+        string viewName,
+        CancellationToken cancellationToken)
     {
         const string sql = """
             SELECT service_id AS ServiceId,
@@ -459,12 +502,25 @@ public sealed partial class MySqlMvRegistryStore : MvForcedReverseRegistryStoreB
               AND view_name = @ViewName;
             """;
 
-        await using var connection = new MySqlConnection(_connectionString);
+        await using var connection = new MySqlConnection(connectionString);
         await connection.OpenAsync(cancellationToken).ConfigureAwait(false);
+        if (readOnly)
+        {
+            await SetReadOnlySessionAsync(connection, cancellationToken).ConfigureAwait(false);
+        }
         var row = await connection.QuerySingleOrDefaultAsync(
             new CommandDefinition(sql, new { ServiceId = serviceId, ViewName = viewName }, cancellationToken: cancellationToken))
             .ConfigureAwait(false);
         return row is null ? null : MapActiveEntry(ToDictionary(row));
+    }
+
+    public Task<MvActiveEntry?> ReadActiveAsync(
+        string serviceId,
+        string viewName,
+        CancellationToken cancellationToken = default)
+    {
+        RecordReadOnlyConnection();
+        return ReadActiveWithConnectionAsync(ReadOnlyConnectionString, readOnly: true, serviceId, viewName, cancellationToken);
     }
 
     public Task SetActiveAsync(
--- a/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvRegistryStore.cs
+++ b/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvRegistryStore.cs
@@ -4,13 +4,29 @@ using MySqlConnector;
 
 namespace Sekiban.Dcb.MaterializedView.MySql;
 
-public sealed partial class MySqlMvRegistryStore : MvForcedReverseRegistryStoreBase<MySqlConnection>, IMvRegistryStore, IMvSchemaVerifier
+public sealed partial class MySqlMvRegistryStore : MvForcedReverseRegistryStoreBase<MySqlConnection>, IMvRegistryStore, IMvReadOnlyMvInspector
 {
     private readonly string _connectionString;
+    private readonly Action<string>? _catalogCommandRecorder;
 
     public MySqlMvRegistryStore(string connectionString)
+        : this(connectionString, null)
+    {
+    }
+
+    public MySqlMvRegistryStore(string connectionString, Action<string>? catalogCommandRecorder)
     {
         _connectionString = connectionString;
+        _catalogCommandRecorder = catalogCommandRecorder;
+    }
+
+    private CommandDefinition CatalogCommand(
+        string sql,
+        object? parameters = null,
+        CancellationToken cancellationToken = default)
+    {
+        _catalogCommandRecorder?.Invoke(sql);
+        return new CommandDefinition(sql, parameters, cancellationToken: cancellationToken);
     }
 
     public async Task EnsureInfrastructureAsync(CancellationToken cancellationToken = default)
@@ -412,10 +428,17 @@ public sealed partial class MySqlMvRegistryStore : MvForcedReverseRegistryStoreB
         await using var connection = new MySqlConnection(_connectionString);
         await connection.OpenAsync(cancellationToken).ConfigureAwait(false);
         var rows = await connection.QueryAsync(
-            new CommandDefinition(sql, new { ServiceId = serviceId, ViewName = viewName, ViewVersion = viewVersion }, cancellationToken: cancellationToken))
+            CatalogCommand(sql, new { ServiceId = serviceId, ViewName = viewName, ViewVersion = viewVersion }, cancellationToken))
             .ConfigureAwait(false);
         return rows.Select(row => (MvRegistryEntry)MapEntry(ToDictionary(row))).ToList();
     }
+
+    public Task<IReadOnlyList<MvRegistryEntry>> ReadRegistryEntriesAsync(
+        string serviceId,
+        string viewName,
+        int viewVersion,
+        CancellationToken cancellationToken = default) =>
+        GetEntriesAsync(serviceId, viewName, viewVersion, cancellationToken);
 
     public async Task<MvActiveEntry?> GetActiveAsync(
         string serviceId,
--- a/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvRegistryStore.cs
+++ b/dcb/src/Sekiban.Dcb.MaterializedView.MySql/MySqlMvRegistryStore.cs
@@ -4,7 +4,7 @@ using MySqlConnector;
 
 namespace Sekiban.Dcb.MaterializedView.MySql;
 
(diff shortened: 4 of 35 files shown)


tomohisa:
## Blocker — SEK-G32 privilege-boundary contract is not met

At reviewed head `8a10156d71944bac133045bb68aa37351dc183a7`, the packet acceptance criteria **FOUR-PROVIDER ZERO-WRITE PROOF**, **HOST SIDE-EFFECT PROBE**, **WHOLE-BATCH PREFLIGHT PROOF**, **ALL-SQL-SURFACES PROOF**, and **RAW-CONNECTION BYPASS PROOF** and the review-context require VerifyOnly to make no projector or registry mutation, and an enforced policy to govern every projector SQL execution path.

Current production paths violate that contract:

- `MvExecutorBase.InitializeCoreAsync` calls `host.InitializeAsync` and `AuthorizeStatementsAsync` before its VerifyOnly branch (`MvExecutorBase.cs:231-240`). That invokes projector code and records its `CREATE` statements, so it cannot satisfy the required zero-side-effect host probe.
- After successful schema verification, a missing registry row falls through to `OpenConnectionAsync`, `RegisterAsync`, and `CommitAsync` (`MvExecutorBase.cs:290-313`). For SQLite, `OpenConnectionAsync` also executes `PRAGMA synchronous=NORMAL` (`SqliteMvExecutor.cs:67-74`). This is an explicit VerifyOnly mutation/fallback and violates the no-registry-write, dedicated-read-only-inspector boundary.
- Policy is applied only after `host.ApplyEventAsync` returns its batch (`MvExecutorBase.cs:797-815`). The live query ports directly execute SQL, for example `SqliteMvContexts.cs:48-91`; every provider port implements `IMvApplyDbConnectionPort`, and the native adapter exposes `Connection` and `Transaction` (`MvApplyHostSupport.cs:299-306`). A mutating CTE, commented or multi-statement query, or raw connection can therefore execute before or without a verdict. There is no Enforced mode with fail-closed policy behavior.

Required unblock:

1. Put all explicit and implicit initialization paths behind one mode gate that derives bindings from the declarative contract without invoking projector initialization, uses only a dedicated read-only catalog inspector in VerifyOnly, and never calls `EnsureInfrastructureAsync`, normal SQLite open, registration, or commit in that mode.
2. Add an enforced, fail-closed policy wrapper around initialization, apply, rows, single, and scalar execution before any provider command; materialize and preflight each returned batch completely; remove raw connection exposure in enforced mode while preserving legacy mode.
3. Add decisive production-path tests: four real stores with a command allowlist recorder and read-only principals, a throwing/counting host, a missing-registry zero-write case, [allow, deny] zero-execution and checkpoint proof, mutating-CTE/commented/multi-statement query-port coverage, raw-bypass coverage, and fault/cancellation semantics. Complete the remaining packet proofs for deterministic full mismatch reporting, binary consumer compatibility, hosted-worker gating, and release documentation.

The currently green checks and the new tests are necessary but insufficient because their VerifyOnly test explicitly accepts a recorded `CREATE TABLE` and does not exercise these bypasses.

tomohisa:
## Blocker — SEK-G32 focused rereview remains request-update

Reviewed immutable head: `fd632274b0f64fe38e4d700c2326c57c88c3264a`.

The new Enforced wrapper materially improves the returned-batch and query-port paths, but the frozen packet and #1125 acceptance criteria are still not fully met. Green CI is necessary, not sufficient.

### 1. VerifyOnly isolation is still not closed

The packet's **FOUR-PROVIDER ZERO-WRITE PROOF**, **NO-FALLBACK PROOF**, and **IMPLICIT-PATH ARCHITECTURE PROOF** require VerifyOnly to use a dedicated read-only inspector and never reach registry mutation.

- After the VerifyOnly boundary succeeds, `CaptureTargetCheckpointAsync` still calls `_registryStore.SetTargetCheckpointAsync` (`MvExecutorBase.cs:152-172`), and the catch-up checkpoint path does the same (`MvExecutorBase.cs:547-573`). In SQLite this is an `UPDATE sekiban_mv_registry` (`SqliteMvRegistryStore.cs:382-413`) through a normal `SqliteConnection` (`:748-759`). A compatible pre-provisioned VerifyOnly route can therefore mutate a registry row.
- `IMvReadOnlyMvInspector` is only an interface boundary: `ReadRegistryEntriesAsync` delegates to `GetEntriesAsync`, which opens `new SqliteConnection(_connectionString)` (`SqliteMvRegistryStore.cs:450-463`), and `VerifySchemaAsync` does the same (`SqliteMvSchemaVerifier.cs:18-19`). The Postgres/MySQL/SQL Server counterparts use the same ordinary connection construction. No production inspector forces a read-only connection/principal.
- The SQLite test passes a `Mode=ReadOnly` connection string from the test itself (`MvVerifyOnlyAndPolicyTests.cs:33-45`); it does not prove that the inspector selects a read-only path with an ordinary configured store. The three server-store tests likewise use ordinary fixture credentials.

Required unblock: make each VerifyOnly inspector independently read-only (including SQLite `Mode=ReadOnly` plus read-only metadata commands), prevent every VerifyOnly implicit operation from calling any mutating registry API, and add decisive real-store tests with a compatible pre-provisioned schema, writable normal configuration, read-only principal/connection enforcement, and counters that fail if normal open, ensure, register, target/checkpoint update, transaction, or commit is reintroduced. Exercise target capture, status lookup, checkpoint capture, and event apply independently after the gate succeeds.

### 2. Required non-vacuous packet proof is incomplete

The packet's **MISMATCH MUTATION MATRIX**, **BINARY EVIDENCE**, and exact SQL-policy contract remain incomplete:

- `MvSchemaColumnRequirement` and `MvObservedSchemaColumn` model only name/type/nullability, while `MvSchemaTableRequirement` and `MvObservedTableSchema` model only columns/primary key (`MvInitialization.cs:77-129`). `Validate` can never produce `DefaultMismatch` or `RequiredIndexMissing` (`:292-371`), nor validate generated semantics, size/precision, unique keys, or correctness-critical indexes. The new deterministic test only covers table/type/nullability/PK (`MvInitializationCompatibilityTests.cs:75-116`), not the required all-provider mutation matrix.
- The PR adds reflection pins but no fixture compiled against the `Sekiban.Dcb.MaterializedView` 10.13.0 package and then executed against the new assemblies. A local API diff being additive is useful evidence, but it is not the required binary-consumer run.
- `MvSqlStatementContext` supplies only `Initialization`/`Apply` phase and an unset nullable `DatabaseType` (`MvSqlStatementPolicy.cs:29-42`); it has no required `ProjectorInitialize`/`ProjectorApply`/`ProjectorQuery` origin. `MvSqlPolicyDecision` cannot carry a rule id, so the failure's `RuleId` is never populated. This does not meet the packet's required policy context and typed-denial payload even though the wrapper blocks the tested query ports.
- The current implicit-path test lists target-capture/catch-up/apply only. Its catch-up failure occurs at checkpoint capture before the subsequent status lookup, so it is not a non-vacuous production-path proof for all four named routes.

Required unblock: complete the declarative, versioned contract and all four provider inspectors for defaults/generated semantics/type width/precision/keys/indexes/bindings with deterministic full-list typed diagnostics; add the independent v10.13.0 package-built binary fixture; carry provider, exact origin, statement metadata, fingerprint, and rule id through policy decisions/failures; and add real-store killing tests that fail for each named regression. Preserve the demonstrated Legacy defaults and existing behavior while doing so.

The resolved/outdated S2077 thread was re-triaged separately: its static parameterized SQLite catalog query is appropriate and is not this blocker.

tomohisa:
## Blocker — SEK-G32 final rereview still requires update

Reviewed immutable head: `00e096325d8d4cf72247653124e768d35128028c`.

This review is limited to the two residuals from [the prior blocker](https://github.com/J-Tech-Japan/Sekiban/pull/1127#issuecomment-5292312496). The central executor/worker guards, mutation counters, v10.13.0 binary-consumer run, and focused suites materially improve the PR, but both residual contracts are not yet closed.

### 1. VerifyOnly does not have an actual read-only SQL Server boundary

The frozen packet's **FOUR-PROVIDER ZERO-WRITE PROOF** and review context require production inspector-enforced read-only access from ordinary writable configuration, plus a real-store proof that mutation is impossible.

- `SqlServerMvRegistryStore` turns the ordinary connection string into one with only `ApplicationIntent=ReadOnly` ([lines 33–45](https://github.com/J-Tech-Japan/Sekiban/blob/00e096325d8d4cf72247653124e768d35128028c/dcb/src/Sekiban.Dcb.MaterializedView.SqlServer/SqlServerMvRegistryStore.cs#L33-L45)); the schema verifier and inspector reads use that string ([verifier lines 15–17](https://github.com/J-Tech-Japan/Sekiban/blob/00e096325d8d4cf72247653124e768d35128028c/dcb/src/Sekiban.Dcb.MaterializedView.SqlServer/SqlServerMvSchemaVerifier.cs#L15-L17), [registry lines 447–455](https://github.com/J-Tech-Japan/Sekiban/blob/00e096325d8d4cf72247653124e768d35128028c/dcb/src/Sekiban.Dcb.MaterializedView.SqlServer/SqlServerMvRegistryStore.cs#L447-L455)). It neither selects a restricted principal nor otherwise makes a standalone SQL Server session read-only.
- The all-provider test passes the ordinary fixture connection string into the production store and asserts only the recorder marker `sqlserver:ApplicationIntent=ReadOnly` ([test lines 1128–1227](https://github.com/J-Tech-Japan/Sekiban/blob/00e096325d8d4cf72247653124e768d35128028c/dcb/tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests/MvVerifyOnlyAndPolicyTests.cs#L1128-L1227)). Its SQL Server fixture is a standalone `MsSqlBuilder` instance, not an availability group or restricted-user setup ([fixture lines 602–615](https://github.com/J-Tech-Japan/Sekiban/blob/00e096325d8d4cf72247653124e768d35128028c/dcb/tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests/MultiProviderFixtures.cs#L602-L615)).
- I verified this against a disposable SQL Server 2022 real store: a `sqlcmd -K ReadOnly` connection successfully executed `INSERT INTO readonly_probe ...` and returned `rows_written = 1`. This matches Microsoft’s documentation: `ApplicationIntent` is an availability-group/read-routing signal and, when no special read target exists, the regular database is used ([Microsoft Learn](https://learn.microsoft.com/en-us/sql/connect/odbc/linux-mac/odbc-driver-on-linux-support-for-high-availability-disaster-recovery?view=sql-server-ver17)).

Required unblock: make the SQL Server inspector use a genuinely enforced read-only capability (for example, an explicit least-privilege inspection principal or a verified readable-replica route), fail typed before inspection when that capability cannot be established, and add a production-inspector real-store killing test beginning with the normal writable configuration that attempts DML and proves rejection, catalog-only access, and zero registry/view mutation. A string marker is not proof.

### 2. Schema-contract completion is not provider-complete or non-vacuously proven

The residual **MISMATCH MUTATION MATRIX** requires all four provider inspectors to validate defaults, ordered/unique indexes, generated semantics/expressions, width, and precision/scale with deterministic complete diagnostics.

- The SQLite inspector obtains `type`, default, and generated flag from `pragma_table_xinfo`, but constructs observed columns with only `DefaultSql` and `IsGenerated` ([lines 9–54](https://github.com/J-Tech-Japan/Sekiban/blob/00e096325d8d4cf72247653124e768d35128028c/dcb/src/Sekiban.Dcb.MaterializedView.Sqlite/SqliteMvSchemaVerifier.cs#L9-L54)). It never derives `GenerationExpression`, `MaxLength`, `Precision`, or `Scale`, even though the common validator compares each of those required fields ([validator lines 439–475](https://github.com/J-Tech-Japan/Sekiban/blob/00e096325d8d4cf72247653124e768d35128028c/dcb/src/Sekiban.Dcb.MaterializedView/MvInitialization.cs#L439-L475)). A compatible SQLite contract requiring an expression or `DECIMAL(p,s)` therefore cannot be recognized as compatible; the inspector cannot distinguish a matching from a mismatching value.
- The new mismatch test constructs synthetic `MvObservedSchemaColumn` instances directly ([test lines 119–180](https://github.com/J-Tech-Japan/Sekiban/blob/00e096325d8d4cf72247653124e768d35128028c/dcb/tests/Sekiban.Dcb.MaterializedView.Tests/MvInitializationCompatibilityTests.cs#L119-L180)). The real four-provider projector contract contains only basic type/null/key requirements ([fixture lines 37–55](https://github.com/J-Tech-Japan/Sekiban/blob/00e096325d8d4cf72247653124e768d35128028c/dcb/tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests/MultiProviderFixtures.cs#L37-L55)), so it cannot kill an inspector that drops these metadata fields.

Required unblock: complete the SQLite metadata path (or return a precise unsupported-capability failure for contract dimensions SQLite cannot faithfully inspect) and add real-store, production-inspector mutation tests for every required schema dimension across all four providers. Each test must show a compatible contract passes, each isolated mismatch returns its exact typed code in the deterministic full list, and a mutation removing that provider check drives the test red.

`Closes #1125`, `main` base, the public API/binary fixture, and the previously accepted policy/default paths were rechecked but are not the basis of this blocker. The documented host queue linkage remains a separate `host-metadata-blocked` gap and is not a PR finding.

tomohisa:
## Blocker — SEK-G32 final rereview round 2 requires update

Reviewed immutable head: `7c5cb72f6dd15d15eedcc44da164962accdfb1a7`.

This focused pass rechecked only the two residuals in [comment 5294834660](https://github.com/J-Tech-Japan/Sekiban/pull/1127#issuecomment-5294834660) against the [SEK-G32 packet](https://github.com/J-Tech-Japan/SekibanIntentHost/blob/main/.intent-cli/issues/SEK-G32/packet.yaml), [review context](https://github.com/J-Tech-Japan/SekibanIntentHost/blob/main/.intent-cli/issues/SEK-G32/review-context.md), and [#1125](https://github.com/J-Tech-Japan/Sekiban/issues/1125).

### Residual 1 is closed

The production SQL Server route now takes the ordinary writable provider connection separately from `MvOptions.SqlServerInspectionConnectionString`, fails with typed `UnsupportedProviderCapability` before catalog inspection when the restricted capability is absent, and uses that distinct principal for schema and registry inspection. The SQL Server fixture starts from the normal writable container connection, provisions a `db_datareader` + `VIEW DEFINITION` inspector principal, deliberately sets its `ApplicationIntent=ReadWrite`, and proves a real `INSERT` fails with SQL error 229 while the row count is unchanged. The common production path separately records catalog reads and preserves registry entries. This is no longer the prior ApplicationIntent-marker-only proof.

### Residual 2 is still not closed: the real-store matrix does not kill uniqueness or generation-expression checks

The new four-provider `SchemaContractMatrix_UsesTheProductionInspector` is a useful real-store improvement and passes, but it does not meet the packet's **MISMATCH MUTATION MATRIX** requirement for every schema dimension:

- Its schema contains only a unique index and its compatible contract also requires `IsUnique=true`. The only index mismatch reverses the columns but retains `IsUnique=true` ([matrix lines 1252–1259](https://github.com/J-Tech-Japan/Sekiban/blob/7c5cb72f6dd15d15eedcc44da164962accdfb1a7/dcb/tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests/MvVerifyOnlyAndPolicyTests.cs#L1252-L1259)). Thus it proves ordered columns, not the `IsUnique` branch: an inspector that reports every index as unique still passes the compatible case and this mismatch. There is no real-store compatible non-unique-index case followed by its isolated uniqueness mismatch.
- Its only generated mismatch changes `IsGenerated` to `false` ([lines 1260–1269](https://github.com/J-Tech-Japan/Sekiban/blob/7c5cb72f6dd15d15eedcc44da164962accdfb1a7/dcb/tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests/MvVerifyOnlyAndPolicyTests.cs#L1260-L1269)). It never changes `GenerationExpression`. Moreover every fixture expression is the constant `7`, and SQLite/SQL Server expected values are literal `"7"` rather than independently read metadata ([lines 1348–1424](https://github.com/J-Tech-Japan/Sekiban/blob/7c5cb72f6dd15d15eedcc44da164962accdfb1a7/dcb/tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests/MvVerifyOnlyAndPolicyTests.cs#L1348-L1424)). A broken inspector that returns `"7"` for all generated columns therefore passes this matrix; it is not a non-vacuous expression proof.

Required unblock: extend the **real production-inspector matrix for all four providers** with (1) an actual non-unique index whose `IsUnique=false` contract passes and whose isolated `IsUnique=true` requirement returns exactly `RequiredIndexMissing` (retain the ordered-index mismatch separately), and (2) a non-constant generated expression whose matching contract passes and whose altered expression alone returns exactly `GeneratedSemanticsMismatch`. Assert the exact deterministic `Mismatches` and `Failure.Mismatches` lists for each case, and demonstrate that removing each provider's uniqueness/expression observation makes its named real-store test fail. This must cover SQLite's derivation or its precise typed unsupported result as applicable.

Focused local evidence was green but is necessary, not sufficient: the six real-store residual tests passed on both net9.0 and net10.0, and the eight compatibility tests passed on both frameworks. All four GitHub checks are green and the PR-scoped open code-scanning alert count is zero at this head.

PR #1135, opened 2026-08-17
SEK-G35: recover projection status heartbeats

### LINKED ISSUE #1133: Projection status heartbeat is frozen after the first write: every subsequent CAS update is rejected
## Summary

The projection status heartbeat written by `MultiProjectionGrain` is persisted only once, right after activation. Every subsequent update is rejected by the CAS check, so the row stays frozen at `Sequence=1, AppliedEventCount=0, phase=catchingUp` with an expired lease.

`IProjectionStatusReader` faithfully returns that frozen row, so an operations dashboard shows `progress 0 / <total events>` and a permanent "stale" state, while the projection itself keeps working normally. The heartbeat never recovers on its own.

Sekiban.Dcb **10.13.0**. Reproduced with **both the SQLite provider (local) and the Cosmos DB provider (Azure)**, so it does not look provider-specific.

## Symptom 1 — local (Sekiban.Dcb.Sqlite, single silo)

Every row in `dcb_projection_statuses` is still the initial write made at grain activation, unchanged for more than 30 minutes:

| ProjectorName | ProjectorVersion | Sequence | AppliedEventCount | Phase | RecordedAtUtc | LeaseExpiresAtUtc |
|---|---|---|---|---|---|---|
| KanyushaListProjection | 1.0.21-haigyogo-keizoku-uketsuke-date-20260729 | 1 | 0 | catchingUp | 2026-08-16T23:08:36Z | 23:09:36Z |
| AutoMailSendSummaryProjection | 1.0.1-perf | 1 | 0 | catchingUp | 23:14:09Z | 23:15:09Z |
| MailTemplateSendSummaryProjection | 1.7.1-perf | 1 | 0 | catchingUp | 23:20:53Z | 23:21:53Z |
| MailTemplateSendRecipientProjection | 1.5.1-perf | 1 | 0 | catchingUp | 23:22:36Z | 23:23:36Z |

`LastAppliedSortableUniqueId` is NULL on every row. The event store holds 1,161,249 events.

Log (`Sekiban.Dcb.Orleans.Grains.MultiProjectionGrain`, Warning, repeating every 30 seconds):

```
[KanyushaListProjection] Projection status heartbeat CAS conflict: Heartbeat CAS rejected: expected sequence 1.
```

**The `ProjectorVersion` of these rows matches the running code's `MultiProjectorVersion` exactly.** In other words the row exists under the expected key, its `Sequence` equals the expected value of 1, and the CAS is still rejected.

## Symptom 2 — Azure (Sekiban.Dcb.CosmosDb, App Service host)

Same behaviour, but the rejection reason has an extra clause:

```
[NendoKeiyakuMitsumoriProjection] Projection status heartbeat CAS conflict: Heartbeat CAS rejected: expected sequence 1, row is absent.
```

106 occurrences in the last 24 hours, every 30 seconds, still ongoing. Meanwhile **the status reader can read that projector's row without any problem** — the reader sees the row while the writer reports it as absent.

Other projectors in the same process log `Heartbeat CAS rejected: activation row already exists.` once or twice at activation and then update normally (for example `MailTemplateSendRecipientProjection` is at `sequence 5188` / caughtUp). Only some projectors get stuck permanently.

### Timeline (UTC) — the projector finished while the heartbeat stayed at 0

| Time | Event |
|---|---|
| 00:22:38 | `Resetting integrity guard: LastGoodSafeVersion was N but external snapshot is missing` → full rebuild starts with no snapshot |
| 00:22:38 | First heartbeat written (`Sequence=1`, `applied=0`, lease 00:23:38) |
| 00:23:08 onwards | Every heartbeat rejected with `expected sequence 1, row is absent` (30 s interval) |
| 00:25–00:34 | Rebuild load causes 468 × `Dropping expired message` and 6 × `TimeoutException` |
| 00:34:38 | **Rebuild completes; a snapshot with 2,575,916 processed events is persisted to the state store** |
| 00:52:24 | Status read through `IProjectionStatusReader` still reports `catchingUp 0/2576130` |
| 01:17:49 | CAS conflicts still ongoing |

## Impact

- Projection progress, phase and lease become unusable for operations and monitoring: the reported values are unrelated to reality.
- The affected projector is permanently displayed as "stale, progress 0", which is easily mistaken for an incident.
- No functional impact on the projection itself, which keeps processing events.

## Expected behaviour

- Heartbeats keep being written after the first one, with `Sequence` and `AppliedEventCount` advancing.
- After a rejected CAS, the writer re-reads the current stored value so the next attempt can succeed (today it retries forever with the same expected value and never recovers).
- If the target row does not exist, the writer falls back to an insert.

## Notes / questions

- The row key is `(ServiceId, ProjectorName, ProjectorVersion, ClusterId)` and does **not** include `ActivationId` (confirmed in the SQLite DDL). If the CAS also compares `ActivationId` or lease ownership, the failure is still reported with sequence wording, which hides the real mismatch. **Including both expected and actual values in the reason (expected/actual sequence, activationId, version) would make this much easier to diagnose.**
- On Cosmos, `row is absent` suggests the insert path and the update path resolve the id / partition key differently, given that the reader can read the same row.
- The expected sequence stays at 1 forever, which suggests the writer's cached expected sequence is never advanced after the initial write.

## Environment

- Sekiban.Dcb 10.13.0 (`Sekiban.Dcb.Orleans.WithoutResult`, `Sekiban.Dcb.Sqlite`, `Sekiban.Dcb.CosmosDb`)
- .NET 9, Orleans, single silo
- Local: SQLite provider, ~1.16M events
- Azure: Cosmos DB provider, App Service host, ~2.58M events

### LINKED ISSUE #1134: SEK-G35 dcb: projection status heartbeat freezes permanently
# SEK-G35 dcb: projection status heartbeat freezes permanently

Fixes #1133.

## Goal

Make the projection status heartbeat keep advancing. Today it stops after the
first write and never recovers, silently, on every storage provider — for two
independent reasons — and the reader then presents the frozen row as the
current value.

## Why This Slice Exists Now

Reported from a production-equivalent environment (#1133): a projector's status
froze at sequence 1 and was mistaken for an outage. The aic design thread
investigated against live Cosmos documents; Sekiban design confirmed every
claim in code and found that PostgreSQL carries the same defect as SQLite.

## Current Observed State

Verified on `dcb-v10.15.0`:

- **SQLite and PostgreSQL**: `INSERT ... SELECT ... WHERE @expectedSequence = 0
  ON CONFLICT ... DO UPDATE ...` — the `WHERE` gates the SELECT that *feeds*
  the insert, not the conflict branch. With expected ≥ 1 nothing is inserted,
  no conflict fires, `DO UPDATE` is never evaluated, and the write is
  rejected. The update path has been unreachable since 10.10.0.
- **Cosmos**: the status document id derives from `ProjectorVersion`, which the
  grain resolves from `_stateStore.Committed.ProjectorVersion` — snapshot
  metadata that is replaced inside a live activation. Once the local sequence
  is ≥ 1, every later heartbeat point-reads an id that does not exist →
  `Rejected(null, …)` → the rebase branch cannot run → permanent freeze.
  Live evidence: successful write at 00:22:38, `row is absent` from 00:23:08 —
  one heartbeat tick apart, so no activation boundary is involved.
- **Reader**: with no version filter the query returns rows of any version, so
  an orphaned old-version row is shown as current.

- **Core invariant**: identity must never depend on a value that can change
  during the activation's life.
- **Activation-pinned identity**: keep `ProjectorVersion` in the physical
  identity, but capture it ONCE after `_host` creation in `OnActivateAsync`
  from `_host.GetProjectorVersion()` (never from
  `_stateStore.Committed.ProjectorVersion`) and use it for the activation's
  whole life; host recreation verifies/logs disagreement but keeps the pin.
  Removing version from identity was rejected (rolling-deployment contention
  plus five-provider key migration).
- **Typed conflict payload** on `ProjectionStatusWriteResult` (reason:
  RowAbsent | RowAlreadyExists | SequenceMismatch | ProviderPreconditionFailed,
  plus expected/observed sequence, version and activation id), with the
  existing string property retained and its text derived from the payload.
- **Writer state machine**: RowAbsent with expected > 0 resets only the local
  fence to 0, keeps the pinned identity, and retries later through the
  provider-conditional create — never an unconditional create in the same
  operation. Applies to `MultiProjectionGrain` and to
  `MvProjectionStatusPublisher`'s fence.
- **Reader visualizes, never rejects**: expected version, observed row version
  and stale/orphan status are explicit; no automatic deletion in this slice.
- **Serialized V2** carries version disposition; V1 behaviour and vectors are
  frozen.
- No schema or key migration. Release: dcb-v10.16.0 (minor), one complete
  release.

## Target Repo / Path / Part

Target paths: `dcb/src dcb/tests docs/dcb_llm docs/dcb_llm_ja`
Target part: pinned writer identity in MultiProjectionGrain + SQLite/PostgreSQL update routing + typed conflict payload and rebase state machine across all providers + MV publisher fence parity + reader version disposition + serialized V2 + tests + docs.

## In Scope

- Everything in the Accepted Baseline, the twelve acceptance proofs below,
  EN/JA docs and the 10.16.0 release-note text.

## Out Of Scope

- Removing `ProjectorVersion` from the physical identity.
- Deleting or retaining orphan rows automatically (separate policy slice).
- Changing V1 serialized behaviour or vectors.
- Repeating the incident investigation — the mechanism is established.

The child PR must capture the writer identity once per activation from the
host projector version, make the SQLite and PostgreSQL update routes
reachable, return a typed conflict payload with expected/observed pairs from
every provider, implement the absent-to-rebase state machine (never an
unconditional create) in both production writers, classify rather than reject
version-mismatched rows on read, add the V2 serialized disposition while
freezing V1, and prove all of it with the shared store-contract suite plus the
killing tests below — additive public API only, no migration.

## Acceptance Criteria

- Repeat-write proof on InMemory, SQLite, PostgreSQL, Cosmos and Dynamo: expected 0 → Sequence 1, expected 1 → Sequence 2 with AppliedEventCount advancing.
- Success-path version change: after Sequence=1, mutating the committed version still commits Sequence=2 via the pinned identity; fails if the mutable value is used.
- Create-conflict route: writer B (local=0) gets typed RowAlreadyExists with pairs, rebases to 1, commits Sequence=2, and still advances after a later committed-version change; fails if a conflict only logs.
- Absent path: expected>0 against a removed row returns typed RowAbsent with no create; the fence resets to 0 and the next conditional create succeeds; fails if absent inserts unconditionally.
- Absent race: a competing create between signal and retry yields RowAlreadyExists, rebase, later update — never an overwrite.
- Mismatch diagnostics: identical typed reasons and exact expected/observed pairs from every provider; text derived from the payload, never "activation row".
- Reader rolling case: two fresh version rows both visible and classified; an expired old row is explicitly stale/orphan and never uniquely current.
- MV publisher parity: its fence consumes the same RowAbsent rebase contract and recovers from external row loss.
- Existing-row upgrade with no schema or key migration.
- SQL matrix incl. PostgreSQL; the SQLite case runs the exact statement shape on the minimum supported native version.
- Serialized V1 vectors byte/shape frozen; new diagnostics only in V2.
- Additive API baseline (reflection-pinned) + EN/JA docs + 10.16.0 release note stating that SQLite/PostgreSQL updates were unreachable since 10.10.0 and Cosmos froze after an in-activation version change.

## Verification

## Related Links

- Host design record: intents/sekiban/design/projection-status-heartbeat-freeze.md
- Consultation: reports/sek-g35-design-consult.md

## Knowledge Maintenance

- Intent placement: host design doc projection-status-heartbeat-freeze.md.
- Docs update: EN/JA projection status guidance — in scope.
- Closeout writeback expected: yes — pin capture point and host-recreation handling, any further writer with the same frozen-fence mode, the SQLite minimum-version parser finding, orphan retention follow-up.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

- Pin each grain activation to its host-provided projection-status identity and make both writers rebase safely after a missing row.
- Repair the fail-closed SQLite/PostgreSQL update route and return structured, provider-neutral CAS conflict diagnostics.
- Add V2 rolling-version status diagnostics while preserving V1 vectors and CLR signatures, with provider, writer, reader, and documentation coverage.

Closes #1134
Fixes #1133

## Verification

- dotnet build dcb/Sekiban.Dcb.slnx --artifacts-path /private/tmp/sek-g35-build.YAusWE -m:1
- Focused projection-status test suites on net9.0 and net10.0, including the real PostgreSQL Testcontainers matrix.

### DIFF
--- a/dcb/src/Sekiban.Dcb.Core/Actors/IProjectionStatusReader.cs
+++ b/dcb/src/Sekiban.Dcb.Core/Actors/IProjectionStatusReader.cs
@@ -36,6 +36,17 @@ public interface ISerializedProjectionStatusReader
         ProjectionStatusReadRequest? request = null,
         CancellationToken cancellationToken = default);
 
+    /// <summary>
+    ///     Optional V2 serialized observation surface. Existing implementers remain binary compatible; implementations
+    ///     that do not support V2 report a typed unsupported-version error rather than changing V1 behavior.
+    /// </summary>
+    Task<ResultBox<byte[]>> ReadSerializedV2Async(
+        ProjectionStatusReadRequest? request = null,
+        CancellationToken cancellationToken = default) =>
+        Task.FromResult(ResultBox.Error<byte[]>(
+            new UnsupportedSerializedProjectionStatusVersionException(
+                SerializedProjectionStatusEnvelopeV2.CurrentVersion)));
+
     Task<ResultBox<byte[]>> GetSerializedSnapshotsAsync(
         ProjectionStatusReadRequest? request = null,
         CancellationToken cancellationToken = default) =>
--- a/dcb/src/Sekiban.Dcb.Core/Actors/ProjectionStatusReader.cs
+++ b/dcb/src/Sekiban.Dcb.Core/Actors/ProjectionStatusReader.cs
@@ -62,9 +62,12 @@ public sealed class ProjectionStatusReader : IProjectionStatusReader
                     new UnauthorizedAccessException("Projection status ServiceId is owned by the server."));
             }
 
+            var expectedProjectorVersion = string.IsNullOrWhiteSpace(request?.ExpectedProjectorVersion)
+                ? null
+                : request!.ExpectedProjectorVersion;
             var rowsResult = await _statusStore.ListAsync(
                 request?.ProjectorName,
-                request?.ProjectorVersion,
+                expectedProjectorVersion is null ? request?.ProjectorVersion : null,
                 cancellationToken).ConfigureAwait(false);
             if (!rowsResult.IsSuccess)
             {
@@ -173,6 +176,11 @@ public sealed class ProjectionStatusReader : IProjectionStatusReader
                         IsFaulted = faulted,
                         FaultMessage = row.FaultMessage,
                         IsFresh = rowFresh,
+                        ExpectedProjectorVersion = expectedProjectorVersion,
+                        VersionDisposition = ResolveVersionDisposition(
+                            expectedProjectorVersion,
+                            row.ProjectorVersion,
+                            rowFresh),
                         SwitchKind = row.SwitchKind,
                         SwitchReason = row.SwitchReason,
                         SwitchedAtUtc = row.SwitchedAtUtc
@@ -194,6 +202,22 @@ public sealed class ProjectionStatusReader : IProjectionStatusReader
         {
             return ResultBox.Error<IReadOnlyList<ProjectionStatusSnapshot>>(ex);
         }
+    }
+
+    private static ProjectionStatusVersionDisposition ResolveVersionDisposition(
+        string? expectedProjectorVersion,
+        string observedProjectorVersion,
+        bool rowFresh)
+    {
+        if (!rowFresh)
+        {
+            return ProjectionStatusVersionDisposition.StaleOrOrphan;
+        }
+
+        return expectedProjectorVersion is not null &&
+               !string.Equals(expectedProjectorVersion, observedProjectorVersion, StringComparison.Ordinal)
+            ? ProjectionStatusVersionDisposition.VersionMismatch
+            : ProjectionStatusVersionDisposition.Current;
     }
 }
 
@@ -241,7 +265,8 @@ public sealed class SerializedProjectionStatusReader : ISerializedProjectionStat
         }
 
         var version = discriminator.GetValue();
-        if (version != SerializedProjectionStatusRequestEnvelopeV1.CurrentVersion)
+        if (version != SerializedProjectionStatusRequestEnvelopeV1.CurrentVersion &&
+            version != SerializedProjectionStatusRequestEnvelopeV2.CurrentVersion)
         {
             return ResultBox.Error<byte[]>(
                 new UnsupportedSerializedProjectionStatusVersionException(version));
@@ -249,18 +274,32 @@ public sealed class SerializedProjectionStatusReader : ISerializedProjectionStat
 
         try
         {
-            // Phase 2: bind only the already-discriminated V1 shape. Unknown fields, wrong-typed filters, and null
-            // roots are shape errors, never version errors.
-            var envelope = JsonSerializer.Deserialize<SerializedProjectionStatusRequestEnvelopeV1>(
+            if (version == SerializedProjectionStatusRequestEnvelopeV1.CurrentVersion)
+            {
+                // Phase 2: bind only the already-discriminated V1 shape. Unknown fields, wrong-typed filters, and null
+                // roots are shape errors, never version errors. This branch intentionally retains the frozen V1 path.
+                var envelope = JsonSerializer.Deserialize<SerializedProjectionStatusRequestEnvelopeV1>(
+                    utf8Json.Span,
+                    RequestJsonOptions);
+                if (envelope is null)
+                {
+                    return ResultBox.Error<byte[]>(
+                        new SerializedProjectionStatusShapeException("Projection status request envelope is null."));
+                }
+
+                return await ReadSerializedAsync(envelope.ToRequest(), cancellationToken).ConfigureAwait(false);
+            }
+
+            var v2Envelope = JsonSerializer.Deserialize<SerializedProjectionStatusRequestEnvelopeV2>(
                 utf8Json.Span,
                 RequestJsonOptions);
-            if (envelope is null)
+            if (v2Envelope is null)
             {
                 return ResultBox.Error<byte[]>(
                     new SerializedProjectionStatusShapeException("Projection status request envelope is null."));
             }
 
-            return await ReadSerializedAsync(envelope.ToRequest(), cancellationToken).ConfigureAwait(false);
+            return await ReadSerializedV2Async(v2Envelope.ToRequest(), cancellationToken).ConfigureAwait(false);
         }
         catch (JsonException ex)
         {
@@ -290,10 +329,36 @@ public sealed class SerializedProjectionStatusReader : ISerializedProjectionStat
         return ResultBox.FromValue(JsonSerializer.SerializeToUtf8Bytes(envelope, JsonOptions));
     }
 
+    /// <summary>
+    ///     Serializes V2 status observations. V1 remains frozen; V2 is the opt-in surface that includes expected versus
+    ///     observed projector versions and the stale/orphan classification for every row.
+    /// </summary>
+    public async Task<ResultBox<byte[]>> ReadSerializedV2Async(
+        ProjectionStatusReadRequest? request = null,
+        CancellationToken cancellationToken = default)
+    {
+        var result = await _reader.ReadAsync(request, cancellationToken).ConfigureAwait(false);
+        if (!result.IsSuccess)
+        {
+            return ResultBox.Error<byte[]>(result.GetException());
+        }
+
+        var envelope = SerializedProjectionStatusEnvelopeV2.Create(
+            _serviceIdProvider.GetCurrentServiceId(),
+            result.GetValue());
+        return ResultBox.FromValue(JsonSerializer.SerializeToUtf8Bytes(envelope, JsonOptions));
+    }
+
     /// <summary>Serializes the canonical V1 request vector used by endpoints and wire-contract tests.</summary>
     public static byte[] SerializeRequest(ProjectionStatusReadRequest? request = null) =>
         JsonSerializer.SerializeToUtf8Bytes(
             SerializedProjectionStatusRequestEnvelopeV1.Create(request),
+            RequestJsonOptions);
+
+    /// <summary>Serializes a V2 request with its optional expected-version observation.</summary>
+    public static byte[] SerializeRequestV2(ProjectionStatusReadRequest? request = null) =>
+        JsonSerializer.SerializeToUtf8Bytes(
+            SerializedProjectionStatusRequestEnvelopeV2.Create(request),
             RequestJsonOptions);
 
     public static ResultBox<SerializedProjectionStatusEnvelopeV1> Deserialize(ReadOnlySpan<byte> payload)
@@ -348,6 +413,64 @@ public sealed class SerializedProjectionStatusReader : ISerializedProjectionStat
         catch (Exception ex) when (ex is InvalidOperationException or NotSupportedException or OverflowException)
         {
             return ResultBox.Error<SerializedProjectionStatusEnvelopeV1>(
+                new SerializedProjectionStatusShapeException(ex.Message));
+        }
+    }
+
+    /// <summary>Strictly deserializes a V2 status envelope without changing the frozen V1 parser.</summary>
+    public static ResultBox<SerializedProjectionStatusEnvelopeV2> DeserializeV2(ReadOnlySpan<byte> payload)
+    {
+        try
+        {
+            using var document = JsonDocument.Parse(payload.ToArray());
+            if (document.RootElement.ValueKind != JsonValueKind.Object)
+            {
+                return ResultBox.Error<SerializedProjectionStatusEnvelopeV2>(
+                    new SerializedProjectionStatusShapeException("Projection status envelope root must be an object."));
+            }
+
+            var versionResult = ReadVersion(document.RootElement);
+            if (!versionResult.IsSuccess)
+            {
+                return ResultBox.Error<SerializedProjectionStatusEnvelopeV2>(versionResult.GetException());
+            }
+
+            var version = versionResult.GetValue();
+            if (version != SerializedProjectionStatusEnvelopeV2.CurrentVersion)
+            {
+                return ResultBox.Error<SerializedProjectionStatusEnvelopeV2>(
+                    new UnsupportedSerializedProjectionStatusVersionException(version));
+            }
+
+            var envelope = JsonSerializer.Deserialize<SerializedProjectionStatusEnvelopeV2>(payload, JsonOptions);
+            if (envelope is null || string.IsNullOrWhiteSpace(envelope.ServiceId) || envelope.Snapshots is null ||
+                envelope.Snapshots.Any(snapshot => snapshot is null || snapshot.Snapshot is null ||
+                    string.IsNullOrWhiteSpace(snapshot.Snapshot.ProjectorName) ||
+                    string.IsNullOrWhiteSpace(snapshot.Snapshot.ProjectorVersion) ||
+                    string.IsNullOrWhiteSpace(snapshot.Snapshot.ClusterId) ||
+                    string.IsNullOrWhiteSpace(snapshot.Snapshot.ActivationId) ||
+                    string.IsNullOrWhiteSpace(snapshot.Snapshot.Consistency) ||
+                    string.IsNullOrWhiteSpace(snapshot.ObservedProjectorVersion) ||
+                    !Enum.IsDefined(typeof(ProjectionStatusVersionDisposition), snapshot.VersionDisposition)))
+            {
+                return ResultBox.Error<SerializedProjectionStatusEnvelopeV2>(
+                    new SerializedProjectionStatusShapeException("Projection status envelope has an invalid V2 shape."));
+            }
+
+            return ResultBox.FromValue(envelope);
+        }
+        catch (UnsupportedSerializedProjectionStatusVersionException ex)
+        {
+            return ResultBox.Error<SerializedProjectionStatusEnvelopeV2>(ex);
+        }
+        catch (JsonException ex)
+        {
+            return ResultBox.Error<SerializedProjectionStatusEnvelopeV2>(
+                new SerializedProjectionStatusShapeException(ex.Message));
+        }
+        catch (Exception ex) when (ex is InvalidOperationException or NotSupportedException or OverflowException)
+        {
+            return ResultBox.Error<SerializedProjectionStatusEnvelopeV2>(
                 new SerializedProjectionStatusShapeException(ex.Message));
         }
     }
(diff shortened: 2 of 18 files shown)


tomohisa (commented):
# SEK-G35 review — APPROVED (workflow verdict; submitted as COMMENTED because the reviewer and PR author are the same GitHub account)

Reviewed head SHA: `37f8b5c3ab3a02e0442d96707591825218f61600`
Base: `main` @ `714cae1ce71d579de7f79015dcd9f1a06952b672` (merge base, PR is MERGEABLE, non-draft)
Execution unit: SEK-G35 (packet `implementation_issue_packet`, `target_repo: J-Tech-Japan/Sekiban`, `target_path: dcb/src dcb/tests docs/dcb_llm docs/dcb_llm_ja`, `dependencies: []`, `base_branch_policy: direct-main` / `expected_pr_base: main` — actual base is `main`)
Closing reference: `Closes #1134` (linked source issue) and `Fixes #1133` (origin report) are both present — G311 satisfied.

## What was actually verified, and what was not

**Verified by source mapping and by reading the tests against the production code paths at the exact head SHA**, plus **exact-head CI evidence**. `dcbTestsNet9`, `dcbTestsNet10`, `SonarCloud` and `SonarCloud Code Analysis` are all `success` on `37f8b5c3ab3a02e0442d96707591825218f61600`.

**Not run by this reviewer:** no local build, no local `dotnet test`, no manual/runtime execution, no real Cosmos DB or real DynamoDB session. The non-vacuity argument below is a source-level mutation analysis, not a locally executed mutation run. The real-engine claims (SQLite, PostgreSQL) rest on CI executing those suites at this head, which the workflow definition confirms (`run_test_dcb.yml` runs `Sekiban.Dcb.WithResult.Tests` and `Sekiban.Dcb.Postgres.Tests` on both net9.0 and net10.0), together with `PostgresTestFixture` starting a real `postgres:16-alpine` Testcontainer with no skip path.

## The point this slice exists for: is the repeat-write proof real?

Yes. `AssertProjectionStatusCasContractAsync` (`dcb/tests/Sekiban.Dcb.WithResult.Tests/ProjectionStatusRegistryTests.cs`) drives the **production** `IProjectionStatusStore` implementations — not a mocked write result — and the first two steps are exactly the case SEK-G24 omitted:

- `UpsertAsync(first, 0)` → `Committed`, `Sequence = 1`
- `UpsertAsync(first with { Sequence = 2, AppliedEventCount = 2 }, 1)` → `Committed`, `Current.Sequence == 2`, `Current.AppliedEventCount == 2`, and `ListAsync` re-reads the persisted row to confirm `Sequence == 2` / `AppliedEventCount == 2`

It is invoked against **real native SQLite** (`SqliteStatusStore_UsesReachableUpdateRouteAndFailsClosedForAbsentExpectedSequence`, real file DB, executing the exact production `INSERT ... SELECT ... WHERE ... ON CONFLICT ... DO UPDATE ... RETURNING` statement), the Core in-memory store, the Cosmos store and the Dynamo store. `PostgresProjectionStatusStoreTests.CasMatrix_UsesReachableUpdateAndConditionalCreateAgainstRealPostgres` repeats the same matrix against **real PostgreSQL 16** via Testcontainers.

### The four SQL CAS cases (review-context §2) — all four present on BOTH SQLite and PostgreSQL

| case | assertion |
|---|---|
| (a) no row, expected=0 → create succeeds | `Assert.True(created.GetValue().Committed)` |
| (b) row present, expected matches → update succeeds | `Assert.True(updated.GetValue().Committed)` + `Current.Sequence == 2` + `ListAsync` re-read |
| (c) row present, expected mismatched → rejected | `staleUpdate` → `SequenceMismatch`, `ExpectedSequence 1` / `ObservedSequence 2` |
| **(d) no row, expected>0 → rejected** | `missing` (unused `ClusterId`) with `expected = 1` → `RowAbsent`, `Current == null`, `ObservedSequence == null`, **and `ListAsync` proves no row was created** |

(d) is present in both suites, so the "silently bypass CAS by deleting the `WHERE`" failure mode is closed.

The fix itself is the right shape: `WHERE @expectedSequence = 0 OR EXISTS (SELECT 1 FROM dcb_projection_statuses WHERE <identity>)` keeps the `SELECT` that feeds the `INSERT` non-empty **only** when the row exists, so `ON CONFLICT ... DO UPDATE ... WHERE Sequence = @expectedSequence` becomes reachable, while an absent row with `expected > 0` still yields zero rows and an empty `RETURNING`. It also *retains* a `WHERE` clause on the `SELECT`, which matters — see the follow-up note below.

## Mutation analysis (would each killing test actually fail?)

- Revert the `OR EXISTS(...)` in `SqliteProjectionStatusStore` / `PostgresProjectionStatusStore` → case (b) `Assert.True(updated.GetValue().Committed)` fails in both suites. The tests are not green-by-construction.
- Replace the guard with an unconditional `INSERT` → case (d) `Assert.IsType<ProjectionStatusWriteConflict>(absent.GetValue().ConflictDetails)` and the `ListAsync` "no `missing-cluster` row" assertion both fail.
- Derive the heartbeat identity from `_stateStore.Committed.ProjectorVersion` again → `ProjectionStatusHeartbeat_PinsHostVersionAndRebasesMissingRowsWithoutUnconditionalCreate` fails at `Assert.Equal(2, advanced.Sequence)` and at `Assert.Empty(ListAsync("projection", "mutable-v2"))`.
- Let `RecreateHostForFullRebuild` rewrite the pin → `Assert.Equal("registered-v1", pinned.ProjectorVersion)` fails.
- Make `RowAbsent` create in the same operation (grain or MV publisher) → `Assert.Empty(...)` right after the row loss fails in `MultiProjectionGrainPersistPolicyTests` and in `Publisher_RebasesExternalRowLossAndCompetingCreate_UsingTheRealStoreContract`.
- Reinstate `"activation row already exists."` → `Assert.DoesNotContain("activation row already exists", ...)` fails on every provider running the shared contract.
- Drop any pinned CLR constructor → `ProjectionStatusPublicApi_RetainsLegacyClrConstructors` fails.
- Serialize any new field into V1 → `SerializedRequestAndResponse_UseFrozenV1GoldenVectors` fails on the exact byte string.

## Contract-by-contract mapping

**1. Activation-pinned identity (DECISION 1 / review-context §1).** `MultiProjectionGrain.OnActivateAsync` calls `CaptureProjectionStatusWriterIdentity(projectorName)` at `MultiProjectionGrain.cs:2197`, immediately after `_host = _actorHostFactory.Create(...)` at `:2192`, and the pin's version comes from `host.GetProjectorVersion()` (`:2733`–`:2741`). `_stateStore.Committed.ProjectorVersion` no longer appears anywhere in the heartbeat construction — `WriteProjectionStatusHeartbeatAsync` (`:2855`–`:2865`) reads all four identity fields from `writerIdentity`. `RecreateHostForFullRebuild` (`:2708`) calls `VerifyPinnedProjectionStatusWriterIdentityAfterHostRecreation`, which only logs a disagreement and **never** reassigns `_projectionStatusWriterIdentity`. `CaptureProjectionStatusWriterIdentity` is likewise write-once. The killing test mutates the committed version to `mutable-v2` *and* the registered host version to `registered-v2`, recreates the host, and still requires the row under `registered-v1` to advance to `Sequence = 2` with `mutable-v2` empty.

**2. Writer state machine (DECISION 3).** Grain: `MultiProjectionGrain.cs:2903`–`:2916` — `RowAbsent && expectedSequence > 0` resets **only** the local fence to 0 and performs no create in that operation; every other observed row adopts the exact observed sequence. MV publisher parity is present at `MvProjectionStatusPublication.cs:256`–`:267` with the identical branch, and is proven against the real in-memory store rather than a stub. Subsequent creates stay provider-conditional: SQL `INSERT`, Cosmos `CreateItem` (`CosmosProjectionStatusStore.cs:48`), Dynamo conditional put — and the absent→race→`RowAlreadyExists`→no-overwrite sequence is asserted in the grain test (`afterCreateRace.ActivationId == "competing-activation"`), the publisher test, the shared contract (`losingConflict.ObservedActivationId == "competing"`) and the PostgreSQL test.

**3. Typed conflict payload (DECISION 2).** `ProjectionStatusWriteConflict` carries reason / expected+observed sequence / expected+observed version / expected+observed activation id, and the human-readable text is derived via `ToCompatibilityReason()`. The shared contract asserts `ConflictReason == ToCompatibilityReason()` on every provider, so the string cannot drift from the payload. All five stores were converted — `grep` for `ProjectionStatusWriteResult.Rejected` shows no remaining legacy 2-argument call site in `dcb/src`, and `"activation row already exists"` no longer exists in the tree. `ProviderCondition` carries only `"conditional-replace"` / `"conditional-put"`; no Cosmos ETag is exposed.

**4. Reader visualizes, does not reject (DECISION 4).** `ProjectionStatusReader.ResolveVersionDisposition` classifies `Current` / `VersionMismatch` / `StaleOrOrphan` and rejects nothing; `Reader_ExposesRollingVersionsInV2WhileLeavingV1VectorsFrozen` keeps two fresh versions visible and classifies the expired one as stale/orphan. No automatic deletion was introduced anywhere — scanning every added line for `delete` / `drop` / `remove` / `truncate` finds only the test-setup `DROP TABLE IF EXISTS` in `PostgresProjectionStatusStoreTests.InitializeAsync` and documentation prose stating that no provider deletes rows.

**5. Serialized V2 / V1 frozen (DECISION 5).** The V1 golden-vector test is untouched, and it passes because every new snapshot field (`ExpectedProjectorVersion`, `VersionDisposition`, `ObservedProjectorVersion`, `IsStaleOrOrphan`) is `[JsonIgnore]`, as is `ProjectionStatusReadRequest.ExpectedProjectorVersion`. V2 lives in its own envelope and request types; `AcceptAsync` dispatches on the discriminator and keeps the V1 branch intact.

**6. Additive public API.** New members only: a fourth-arity `Rejected` overload (the `(ProjectionStatusHeartbeat?, string)` one is retained and remains unambiguous), `init` properties on existing positional records with no positional-parameter changes, a default-implemented `ReadSerializedV2Async` on `ISerializedProjectionStatusReader`, and new types. `ProjectionStatusPublicApi_RetainsLegacyClrConstructors` reflection-pins three constructors. No schema or key migration; the `dcb_projection_statuses` DDL and the `(ServiceId, ProjectorName, ProjectorVersion, ClusterId)` key are unchanged, so EXISTING-ROW UPGRADE holds. NuGet versions come from the release workflow input, so no in-repo version bump is expected here; the required release-note text is present in `docs/dcb_llm/13_common_issues.md` and its JA counterpart, and it does state that the update path was unreachable on PostgreSQL and SQLite since dcb-v10.10.0 and that Cosmos writers froze after an in-activation version change.

**Out-of-scope boundaries were not crossed:** version is still part of the physical identity, no orphan auto-deletion, no V1 change, no schema/physical-key migration, no re-investigation of the mechanism.

**Intent references consulted:** `intents/sekiban/design/projection-status-heartbeat-freeze.md` (CORE INVARIANT — identity must never depend on a value that can change during the activation's life). The diff is consistent with it: the only identity input is registered projector metadata captured once per activation.

**Nice catch worth recording:** the `MixedDynamoDb` `"="` comparison fix in `ProjectionStatusRegistryTests` removes a real vacuity in the existing test double — previously `actual.S == expected.S` with both sides `null` made any numeric mismatch compare equal, so the Dynamo CAS assertions could pass without the condition actually holding. Tightening the double *before* relying on it for the provider matrix is the right order of operations.

## Follow-ups (non-blocking; recorded here so they are durable rather than chat-only)

1. **AC "SQL PROVIDER MATRIX", second clause is only partly evidenced.** The SQLite test does execute the exact production statement shape against real native SQLite, but against whatever `Microsoft.Data.Sqlite`/SQLitePCLRaw bundles in CI — not a demonstrably *minimum supported* native SQLite build — and the parenthetical "confirm whether the original `WHERE` was motivated by a parser constraint that no longer applies" is not answered anywhere durable in the PR. I did not run the statement against a pinned minimum SQLite build. On reading: SQLite documents that in `INSERT ... SELECT ... ON CONFLICT`, the parser cannot tell whether `ON` opens the upsert or a join constraint unless the `SELECT` carries a `WHERE` clause, which is the most likely original motivation — and this fix *keeps* a `WHERE` clause, so the constraint is respected and no new minimum-version floor is introduced (`RETURNING` already required 3.35+ since SEK-G24; `EXISTS` adds nothing). Please record that confirmation in the SEK-G35 closeout learning, which already expects "the SQLite minimum-version parser finding".
2. **Reader: supplying both `ProjectorVersion` and `ExpectedProjectorVersion` in a V2 request silently drops the explicit filter** (`ProjectionStatusReader.cs`, `expectedProjectorVersion is null ? request?.ProjectorVersion : null`). This is consistent with "classification, not filtering" and cannot affect V1 callers (V1 has no `ExpectedProjectorVersion`), but it is surprising for a V2 caller that sets both and it is not documented. Worth a doc line or an explicit precedence decision in a later slice.
3. **Cosmos DB and DynamoDB parity evidence is test-double backed** (`InMemoryCosmosClient`, `DispatchProxy`-based `MixedDynamoDb`), not real-service backed. That is the established practice in this repo and the two defects that actually shipped broken (SQLite, PostgreSQL) are covered by real engines, so this does not block — but the Cosmos freeze reported in #1133 is only reproduced at the contract level, not against live Cosmos.
4. **Narrow, currently unreachable race in the SQL guard:** if some future code path deleted a status row concurrently between the `EXISTS` evaluation and the `INSERT` within the same statement, the write could land as an unconditional create. Nothing deletes status rows today — this slice explicitly forbids automatic deletion — so it is not reachable now. Please re-check this guard when the retention/cleanup policy slice is designed.

## Sonar

Quality Gate **passed** on this PR (`SonarCloud Code Analysis` = success, `SonarCloud` = "No new alerts in code changed by this pull request"), so the reliability conditions on new code are green — no exemption was needed or claimed for reliability/BUG. Two new issues are reported that did not breach the gate; structural duplication is operator-exempt and is not raised here.

## Verdict

Approved. Passing tests were treated as necessary but not sufficient: the approval rests on the packet contract mapping and intent-reference consistency above, on the four-case CAS matrix being present on both SQL providers including the absent/expected>0 rejection, and on the killing tests driving production writers rather than fakes.

PR #1143, opened 2026-08-21
fix(dcb): make catch-up checkpoints advance on fetched events

### LINKED ISSUE #1142: SEK-G37 dcb: catch-up checkpoints never fire on the hot-only path
# SEK-G37 dcb: catch-up checkpoints never fire on the hot-only path

Fixes #1138.

## Goal

Restore the resume-and-converge guarantee of catch-up: an interrupted catch-up
must continue from a durable intermediate checkpoint instead of replaying from
the same stale snapshot forever.

## Why This Slice Exists Now

Reported with a production-grade trace: consecutive non-empty catch-up batches
logging `PersistTriggered=False, PersistReason=none` throughout the hot tail.
A host with a large backlog and a caller timeout shorter than the full replay
can stay permanently in `catchingUp` with no error logged.

## Current Observed State

Verified on `dcb-v10.17.0` (`cca3c8d6`). Four defects compound:

- The hot-only persist trigger is a single exact-match test
  `_eventsProcessed % 5000 == 0`; the counter is mutated at six sites with
  data-dependent increments and is seeded from the restored snapshot, so its
  residue can become permanently unreachable.
- A non-empty batch whose events are all filtered takes an early return in
  both readers and never reaches the persist decision at all — the reported
  trace (`FetchedCount=500, FilteredCount=500, AppliedCount=0`) never
  consulted the trigger.
- Both outer loops treat `ProcessedCount == 0` as an empty batch and complete
  catch-up after five in a row, so an all-filtered stretch ends the run after
  2,500 fetched events — before any 5,000-event trigger could fire.
- The since-last-persist counter advances by applied count only, so batches
  that lose most fetched events to duplicate/position filtering accrue replay
  cost while advancing it least. The same weakness exists latently on the cold
  path.

- **Empty read = `FetchedCount == 0` only.** A non-empty zero-applied read
  advances the traversal cursor to the last fetched position, resets the
  empty-batch counter, and runs one common progress/decision/telemetry seam.
  `AppliedCount` no longer controls completion.
- **Common seam** for both readers, receiving last-fetched and nullable
  last-applied separately; one decision, at most one persist attempt, one
  summary reporting the trigger decision and the actual persist outcome
  separately.
- **Additive triggers only**: keep every existing trigger, reason and
  precedence; add a fetched-based count trigger on both paths and a
  time-based trigger on the hot-only path.
- **Thresholds by `UsedCold`**: cold reads use the existing configured
  options; hot-only and hybrid hot tail use private constants 5,000 events /
  5 minutes. No new public options.
- **Window normalization**: applied, fetched and time tracking reset together
  at logical catch-up-window boundaries (four defined transitions).
- **Durability gate**: the fetched trigger must be proven to advance the
  committed restart cursor consumed by a fresh activation; if it cannot,
  implementation stops and the slice returns to design.

## Target Repo / Path / Part

- Target part: catch-up persist decision and completion contract in `MultiProjectionGrain` — common post-fetch seam, FetchedCount-defined empty reads, additive fetched/time triggers with path-specific thresholds, window normalization, durability proof.

## In Scope

Everything in the Accepted Baseline, the fourteen acceptance criteria below,
and an EN/JA docs note on the new `PersistReason` strings and cadence change.

## Out Of Scope

- New public options or any `ColdEventStoreOptions` property.
- Orleans response-timeout / first-query-barrier structure.
- Cold export / segment-boundary behavior changes.
- General `PersistStateAsync` failure recovery (its result contract is pinned,
  not redesigned).

The child PR must redefine empty reads by fetched count, route every non-empty
batch of both readers through one common seam, add the fetched and time
fallbacks with path-specific thresholds while preserving every existing
trigger, normalize the tracking fields at window boundaries, prove restart
durability through a fresh activation, and ship the fourteen killing tests —
all with zero public-surface additions.

## Acceptance Criteria

- Enumerable zero-applied production path: all-filtered non-empty batches keep the run active through batch 9 and produce exactly one fetched-triggered attempt on batch 10; current code fails this.
- Streaming zero-applied production path: identical assertions through the streaming reader.
- Durability proof: committed cursor C0 → persist at C1 > C0 → committed restart record equals the contract value → a fresh activation's first authoritative read starts strictly after it; stop on failure per the durability gate.
- Hot threshold boundaries with an irregular series and an arbitrary seeded residue.
- Legacy trigger preservation (old modulo fires first where constructed; reason/precedence pinned).
- Discriminating counter reset cycle on hot and cold.
- Discriminating time reset cycle (a no-fire assertion between two fires).
- Window normalization: all four transitions, seeded so reset/preserve/no-mutation are distinguishable.
- Routing matrix: no-hybrid hot / hybrid hot tail / cold.
- Cold legacy matrix incl. fetched-only case, with and without the segment boundary.
- Frequency bound: exactly one attempt per crossing; attempts vs durable writes counted separately; unchanged-checkpoint guard behavior proven both enabled and disabled.
- Persist result contract: three-field reset after any completed return incl. short-circuit and failed ResultBox, with the no-durable-write outcome reported.
- Partial-filter cursor separation: next read starts after last fetched, not last applied.
- Public surface pin + EN/JA docs note.

## Verification

All new and existing dcb tests green; `git diff --check` clean; no public signature additions, removals or modifications.

## Related Links

- Host design record: intents/sekiban/design/catchup-checkpoint-hot-path.md
- Consultations: reports/sek-g37-design-consult.md, reports/sek-g37-design-rereview.md, reports/sek-g37-design-final.md

## Knowledge Maintenance

- Intent placement: host design doc catchup-checkpoint-hot-path.md.
- Docs update: EN/JA PersistReason/cadence note — in scope.
- Closeout writeback expected: yes — durability-gate outcome, seam location, remaining ProcessedCount-as-emptiness callers, PersistReason frequency change.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

- Define catch-up completion by `FetchedCount == 0`; non-empty zero-applied batches advance the traversal cursor and use the shared progress/decision/telemetry seam in both enumerable and streaming readers.
- Preserve the legacy hot modulo trigger and add private hot fetched/time fallbacks (5,000 / 5 minutes); add the fetched fallback to the existing cold policy selected by `UsedCold == true`.
- Normalize applied/fetched/time windows, preserve an inherited forced window, and report `PersistTriggered` separately from `PersistOutcome`.
- Add production-path killing tests, fresh-activation durability coverage, EN/JA documentation, and partial-filter cursor assertions without adding public options, interfaces, or serialized state members.

Fixes #1142

## Verification

- `dotnet test dcb/tests/Sekiban.Dcb.Orleans.Tests/Sekiban.Dcb.Orleans.Tests.csproj --framework net9.0 --filter FullyQualifiedName~MultiProjectionGrainCatchUpProductionPathTests --no-restore` — 10 passed.
- Same production-path filter on `net10.0` — 10 passed.
- `MultiProjectionGrainPersistPolicyTests` on `net9.0` and `net10.0` — 12 passed each.
- `MinimalOrleansTests.Hot_fetched_checkpoint_is_committed_and_consumed_by_fresh_activation` on `net9.0` and `net10.0` — 1 passed each; committed C1 was consumed by the fresh activation's first authoritative read.
- `MultiProjectionGrainCatchUpDiagnosticsTests` on `net9.0` — 6 passed.
- `git diff --check` — clean.

### DIFF
--- a/dcb/src/Sekiban.Dcb.Orleans.Core/Grains/CatchUpPositionContracts.cs
+++ b/dcb/src/Sekiban.Dcb.Orleans.Core/Grains/CatchUpPositionContracts.cs
@@ -82,8 +82,10 @@ internal sealed record CatchUpInvocationResult(
     SortableUniqueId? AuthoritativeReachedPosition);
 
 internal readonly record struct CatchUpBatchResult(
-    int ProcessedCount,
-    SortableUniqueId? AuthoritativeReadCursor);
+    int FetchedCount,
+    int AppliedCount,
+    SortableUniqueId? LastFetchedPosition,
+    SortableUniqueId? LastAppliedPosition);
 
 internal enum CatchUpProductionHookPoint
 {
--- a/dcb/src/Sekiban.Dcb.Orleans.Core/Grains/MultiProjectionGrain.cs
+++ b/dcb/src/Sekiban.Dcb.Orleans.Core/Grains/MultiProjectionGrain.cs
@@ -34,6 +34,11 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
     private const string EmptyLogValue = "empty";
     private const int CatchUpEventTypeSummaryTopN = 5;
     private const long CatchUpInformationElapsedThresholdMs = 1000;
+    private const int HotCatchUpPersistMaxFetchedEvents = 5_000;
+    private static readonly TimeSpan HotCatchUpPersistMaxInterval = TimeSpan.FromMinutes(5);
+    private const string PersistOutcomeNotAttempted = "not_attempted";
+    private const string PersistOutcomeDurableWrite = "durable_write";
+    private const string PersistOutcomeNoDurableWrite = "no_durable_write";
     private const int DefaultSnapshotEnvelopeSizeLimitBytes = 2 * 1024 * 1024;
     private const int SnapshotEnvelopeBase64ExpansionNumerator = 3;
     private const int SnapshotEnvelopeBase64ExpansionDenominator = 4;
@@ -211,8 +216,36 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
     private bool _hybridCatchUpCheckLogged;
     private HybridReadBatchMetadata? _lastHybridReadBatchMetadata;
     private long _eventsProcessedSinceLastCatchUpPersist;
+    private long _eventsFetchedSinceLastCatchUpPersist;
     private DateTime _lastCatchUpPersistUtc = DateTime.UtcNow;
+    private bool? _lastCatchUpUsedCold;
     private long _catchUpBatchSkipCount;
+    private string _lastPersistOutcome = PersistOutcomeNotAttempted;
+
+    private void ResetCatchUpPersistWindow(bool resetReadPath = false)
+    {
+        _eventsProcessedSinceLastCatchUpPersist = 0;
+        _eventsFetchedSinceLastCatchUpPersist = 0;
+        _lastCatchUpPersistUtc = DateTime.UtcNow;
+        if (resetReadPath)
+        {
+            _lastCatchUpUsedCold = null;
+        }
+    }
+
+    private void ObserveCatchUpReadPath(HybridReadBatchMetadata? metadata)
+    {
+        var usedCold = metadata?.UsedCold == true;
+        if (_lastCatchUpUsedCold.HasValue && _lastCatchUpUsedCold.Value != usedCold)
+        {
+            // A cold-to-hot (or hot-to-cold) transition starts a new logical catch-up window. The applied count,
+            // fetched count, and elapsed-time fallback must move together; carrying only one of them makes the next
+            // threshold dependent on the previous storage path.
+            ResetCatchUpPersistWindow();
+        }
+
+        _lastCatchUpUsedCold = usedCold;
+    }
 
     private sealed record PersistPolicySettings(
         int PersistBatchSize,
@@ -247,7 +280,10 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
         int SegmentCount,
         bool PersistTriggered,
         string PersistReason,
-        string EventTypeSummary);
+        string EventTypeSummary)
+    {
+        public string PersistOutcome { get; init; } = PersistOutcomeNotAttempted;
+    }
 
     private sealed record PersistCheckpoint(
         string ProjectorVersion,
@@ -454,7 +490,7 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
         _logger.Log(
             logLevel,
             MultiProjectionLogEvents.CatchUpBatchSummary,
-            "[{ProjectorName}] Catch-up batch summary. BatchNumber={BatchNumber}, StartPosition={StartPosition}, CurrentPosition={CurrentPosition}, LastAppliedPosition={LastAppliedPosition}, TargetPosition={TargetPosition}, RequestedMaxCount={RequestedMaxCount}, FetchedCount={FetchedCount}, FilteredCount={FilteredCount}, AppliedCount={AppliedCount}, PendingStreamEventsBefore={PendingStreamEventsBefore}, PendingStreamEventsAfter={PendingStreamEventsAfter}, ReadElapsedMs={ReadElapsedMs}, ApplyElapsedMs={ApplyElapsedMs}, PersistElapsedMs={PersistElapsedMs}, SafePromotionElapsedMs={SafePromotionElapsedMs}, TotalElapsedMs={TotalElapsedMs}, ReadSource={ReadSource}, ColdEventsRead={ColdEventsRead}, HotEventsRead={HotEventsRead}, ReachedColdSegmentBoundary={ReachedColdSegmentBoundary}, SegmentCount={SegmentCount}, PersistTriggered={PersistTriggered}, PersistReason={PersistReason}, EventTypeSummary={EventTypeSummary}",
+            "[{ProjectorName}] Catch-up batch summary. BatchNumber={BatchNumber}, StartPosition={StartPosition}, CurrentPosition={CurrentPosition}, LastAppliedPosition={LastAppliedPosition}, TargetPosition={TargetPosition}, RequestedMaxCount={RequestedMaxCount}, FetchedCount={FetchedCount}, FilteredCount={FilteredCount}, AppliedCount={AppliedCount}, PendingStreamEventsBefore={PendingStreamEventsBefore}, PendingStreamEventsAfter={PendingStreamEventsAfter}, ReadElapsedMs={ReadElapsedMs}, ApplyElapsedMs={ApplyElapsedMs}, PersistElapsedMs={PersistElapsedMs}, SafePromotionElapsedMs={SafePromotionElapsedMs}, TotalElapsedMs={TotalElapsedMs}, ReadSource={ReadSource}, ColdEventsRead={ColdEventsRead}, HotEventsRead={HotEventsRead}, ReachedColdSegmentBoundary={ReachedColdSegmentBoundary}, SegmentCount={SegmentCount}, PersistTriggered={PersistTriggered}, PersistReason={PersistReason}, PersistOutcome={PersistOutcome}, EventTypeSummary={EventTypeSummary}",
             projectorName,
             telemetry.BatchNumber,
             telemetry.StartPosition,
@@ -479,6 +515,7 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
             telemetry.SegmentCount,
             telemetry.PersistTriggered,
             telemetry.PersistReason,
+            telemetry.PersistOutcome,
             telemetry.EventTypeSummary);
     }
 
@@ -1299,6 +1336,7 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
 
     public async Task<ResultBox<bool>> PersistStateAsync()
     {
+        _lastPersistOutcome = PersistOutcomeNotAttempted;
         try
         {
             var startUtc = DateTime.UtcNow;
@@ -1306,6 +1344,7 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
 
             if (_host == null)
             {
+                _lastPersistOutcome = PersistOutcomeNoDurableWrite;
                 return ResultBox.Error<bool>(new InvalidOperationException("Projection host not initialized"));
             }
 
@@ -1335,6 +1374,7 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
             var shortCircuit = TryShortCircuitPersist(projectorName, checkpoint);
             if (shortCircuit is not null)
             {
+                _lastPersistOutcome = PersistOutcomeNoDurableWrite;
                 return shortCircuit;
             }
 
@@ -1355,6 +1395,7 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
             {
                 _lastError = snapshotWriteResult.GetException().Message;
                 _logger.LogWarning("[{ProjectorName}] {LastError}", projectorName, _lastError);
+                _lastPersistOutcome = PersistOutcomeNoDurableWrite;
                 return ResultBox.FromValue(false);
             }
 
@@ -1504,12 +1545,16 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
                 TryCompactAfterLargePersist(projectorName, envelopeSize);
             }
 
+            _lastPersistOutcome = externalStoreSaved
+                ? PersistOutcomeDurableWrite
+                : PersistOutcomeNoDurableWrite;
             return ResultBox.FromValue(true);
         }
         catch (Exception ex)
         {
             _lastError = $"Persistence failed: {ex.Message}";
             _logger.LogError(ex, "[{ProjectorName}] Persistence failed", GetProjectorName());
+            _lastPersistOutcome = PersistOutcomeNoDurableWrite;
             return ResultBox.Error<bool>(ex);
         }
     }
@@ -1539,6 +1584,7 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
                 {
                     _lastError = writeResult.GetException().Message;
                     _logger.LogWarning("[{ProjectorName}] Streaming snapshot write failed: {Error}", projectorName, _lastError);
+                    _lastPersistOutcome = PersistOutcomeNoDurableWrite;
                     return ResultBox.FromValue(false);
                 }
 
@@ -1612,6 +1658,9 @@ public class MultiProjectionGrain : Grain, IMultiProjectionGrain, ILifecyclePart
                     TryCompactAfterLargePersist(projectorName, tempFileSize);
                 }
 
+                _lastPersistOutcome = externalStoreSaved
+                    ? PersistOutcomeDurableWrite
+                    : PersistOutcomeNoDurableWrite;
                 return ResultBox.FromValue(true);
             }
             catch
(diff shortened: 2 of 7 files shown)


tomohisa (commented):
# SEK-G37 semantic review — REQUEST UPDATE (hard stop)

Reviewed exact head `75cb06ffd4118811929d90dc3241f559c52d94a0` against issue #1142, `SEK-G37` packet, implementation notes, and mandatory review context.

**Workflow verdict: request-update.** This is a COMMENTED same-account review; it is not an approval.

## Blocking implementation finding — AC3 durability gate

**AC3 requires the all-filtered / zero-applied path itself to prove C0 -> C1 durable restart consumption. That evidence is absent, so the guarantee must not be narrowed.**

`MinimalOrleansTests.Hot_fetched_checkpoint_is_committed_and_consumed_by_fresh_activation` is a good proof for a range of newly applied events, but it does not run the required zero-applied range. In contrast, the two zero-applied production-path tests seed every fetched event into `_processedEventIds`; their `ProductionCatchUpProjectionHost.GetSafeLastSortableUniqueIdAsync()` returns `string.Empty`, and they assert write-call counts only. They never inspect the committed `LastSortableUniqueId` or create a fresh activation.

The common seam advances `_catchUpProgress.CurrentPosition` to the last fetched position, but `PersistStateAsync()` commits the host safe position, not that traversal cursor. Therefore the current tests do not show that a zero-applied C1 is committed as `> C0`, nor that a fresh activation's first authoritative read begins after that C1. Per packet AC3(c)/(d) and review-context.md, this is a hard stop.

Required unblock: add an actual-state-store/fresh-activation test which starts from committed C0 before an all-filtered non-empty range, causes the fetched-triggered attempt at C1, asserts the committed restart record cursor is C1 (or at least `> C0` per the contract), then proves the fresh activation's first authoritative read is strictly after it. If that cannot pass on the existing safe-checkpoint mechanism, stop and return to design/operator for the durable traversal-cursor and serialized-state assessment; do not silently weaken the guarantee.

## Acceptance-criteria map

1. **PASS** — enumerable reader's zero-applied early return routes through `UpdateCatchUpProgressAfterBatch`; outer completion uses `FetchedCount`, and the production-path test reaches ten non-empty batches plus five true-empty reads.
2. **PASS** — equivalent independent streaming-reader path and production-path test exist.
3. **FAIL / hard stop** — the mandatory C0/C1 committed-record plus fresh-activation consumption proof is not exercised for the required zero-applied range (finding above).
4. **PASS** — private hot 5,000 fetched fallback has below/exact/overshoot coverage with an arbitrary `_eventsProcessed` residue; the legacy modulo cannot satisfy the fallback assertion.
5. **PASS** — `event_count_checkpoint` remains before the fetched fallback, with reason/precedence pinned.
6. **PASS** — hot and cold reset-cycle tests include a discriminating non-fire and exactly two observed crossings.
7. **PASS** — time reset test includes an immediate non-fire before a second aged-time firing.
8. **PASS** — non-zero seeded four-transition window test covers new run reset, forced inherited preservation, non-superseding early return, and completed-run cleanup.
9. **PASS** — policy selects cold only with `UsedCold == true`; hybrid hot tail remains on private 5,000/5-minute policy.
10. **PASS** — source decision order preserves cold segment/applied/interval precedence and adds fetched-only fallback after them.
11. **PASS** — production tests separately observe persist attempts (metadata calls) and durable writes, including unchanged-checkpoint guarded and unguarded cases.
12. **PASS** — short-circuit and failed `ResultBox` paths reset applied/fetched/time fields and report `no_durable_write`.
13. **PASS** — diff is confined to private grain/internal contract/test/docs work; no public constructor, grain interface, options property, or serialized-state member changed.
14. **PASS** — enumerable and streaming partial-filter tests show next read after last fetched while applied state/ids retain only applied events.

## Other review evidence

- Scope matches #1142 In Scope and the packet target paths; no public options, timeout/barrier restructuring, cold-export behavior change, or general persistence-recovery redesign. PR correctly says `Fixes #1142` and targets `main`.
- EN and JA docs both describe `FetchedCount == 0` completion, `fetched_count_checkpoint` / `time_checkpoint`, `UsedCold`, and separate `PersistTriggered` / `PersistOutcome` semantics.
- `git diff --check origin/main...origin/claude/sek-g37-catchup-checkpoints` is clean. Current exact-base pairing is `7e387e5e20d12f7d5686b6880b75583934accc75...75cb06ffd4118811929d90dc3241f559c52d94a0`.
- Re-run at exact head: production-path tests 10/10 on net9.0 and net10.0; persist-policy tests 12/12 on both; current applied-path fresh-activation test 1/1 on both; diagnostics 6/6 on both. These are necessary but not sufficient because AC3's required zero-applied durability proof is missing.
- Terminal exact-head CI is green: `dcbTestsNet9` and `dcbTestsNet10` (plus Sonar quality gate). Sonar annotations were triaged as null-forgiving/cognitive-complexity maintenance signals, not a reliability/BUG finding; no review threads were present.

The implementation fixes the reader reachability and FetchedCount completion defects, but it cannot receive an approve-equivalent verdict until AC3 proves durable restart consumption for the zero-applied path.

tomohisa (commented):
# SEK-G37 AC3 semantic rereview — REQUEST UPDATE

Reviewed exact head `9d8744f02414c2087e2a60b110389a7b77ecb13a` (base `7e387e5e20d12f7d5686b6880b75583934accc75`) against #1142, the operator-approved [AC3 amendment](https://github.com/J-Tech-Japan/Sekiban/issues/1142#issuecomment-5376352138), the updated `SEK-G37` packet and implementation notes, mandatory `review-context.md`, the design record, and the prior request-update review at `75cb06ffd4118811929d90dc3241f559c52d94a0`.

**Workflow verdict: request-update.** This is the required same-account `COMMENTED` review, not a GitHub approval.

## Blocking implementation finding — AC3b lossless reapplication is not proved

Classification: **implementation-finding** against amended AC3b / the packet durability gate. The repair is test-only, so the required correction is also test-only.

`Synthetic_id_tracking_only_does_not_advance_c0_and_fresh_activation_reapplies_the_range` proves several important negative properties: C0 is supplied to the first read, the synthetic activation takes the unchanged-checkpoint path (`SafeCheckpointCalls == 1`, no snapshot or persistent-state write, `no_durable_write`), and its persistent cursor remains C0. Its new `MultiProjectionGrain` has an empty ID cache and asks its event store to read from C0.

It does **not** yet prove the remaining AC3b requirement that the fresh activation re-applies the exact authoritative range and that state converges without loss. The fresh fixture constructs a separate `ProductionCatchUpEventStore` and `ProductionCatchUpProjectionHost`; that host's `AddSerializableEventsAsync` only adds `events.Count` to `AppliedEventCount` and overwrites `LastAppliedPosition`. The assertions check that count and the last position only. It retains neither the exact IDs/order nor a materialized projection value/state. An interior event omission/repetition, or a host which accepts calls without reaching the correct projection state, can still satisfy both assertions. Thus this is call/position evidence, not a convergence proof.

Required unblock: retain the required C0/no-write/unchanged-checkpoint/empty-cache assertions, but make the fresh fixture consume the same unchanged C0 durable record and an authoritative range whose exact IDs are tracked. Assert the complete re-applied ID sequence (not only its count/final position) and a deterministic materialized result/state equal to applying that range once from C0. A new `MultiProjectionGrain` instance is fine, but the test must establish the actual restart-state/source relationship and exact lossless convergence. Do not persist the traversal cursor or add public/serialized state to make this pass.

## AC3a evidence — PASS

`Hot_fetched_checkpoint_for_stream_preapplied_range_is_durable_and_consumed_by_fresh_activation` is the required production case:

- it starts with a real C0 in `SharedStateStore`, delivers the exact 5,000-event range through the Orleans production stream first, then writes that same range to the real authoritative event store so catch-up refetches it as ID-filtered/zero-applied;
- the stream events are deliberately aged and the first read is gated for a further 18 seconds against the 20-second safe window, making promotion eligibility a real precondition rather than treating stream delivery as durable proof;
- the actual state store record becomes C1, with `EventsProcessed == 5123` and `LastSortableUniqueId` equal to the range tail and strictly after C0; and
- after deactivation, a real fresh grain activation's first recorded authoritative `ReadAllSerializableEventsAsync` cursor equals C1. That API is exclusive, so the read begins strictly after committed C1 rather than replaying C0/C1.

This closes the former AC3a gap; it does not cure the independent AC3b convergence gap above. AC3 is therefore still a hard-stop failure overall.

## Original 14-AC map (AC3 is now provenance-split)

1. **PASS** — enumerable zero-applied production reader reaches the common seam, remains active through batch 9, and crosses exactly once at batch 10.
2. **PASS** — the independently exercised streaming zero-applied reader has the same production-path assertions.
3. **FAIL / hard stop** — AC3a passes as above; AC3b has correct non-advancement and no-write checks but lacks exact reapplication/state-convergence evidence.
4. **PASS** — irregular hot fetched threshold coverage uses a seeded modulo residue and below/exact/overshoot cases.
5. **PASS** — legacy hot modulo reason/precedence remains before the fetched fallback.
6. **PASS** — hot and cold reset cycles contain the discriminating intervening non-fire and exactly two crossings.
7. **PASS** — time reset includes the immediate non-fire before the second aged-time trigger.
8. **PASS** — non-zero seeded four-transition window normalization covers reset, forced inheritance, no mutation, and completed-run cleanup.
9. **PASS** — routing selects cold policy only for `UsedCold == true`; the hybrid hot tail retains private hot thresholds.
10. **PASS** — cold segment/applied/interval precedence is retained and fetched-only cases cover both boundary settings.
11. **PASS** — tests observe attempts separately from durable writes, including guarded unchanged and unguarded behavior.
12. **PASS** — completed success, unchanged short-circuit, and failed `ResultBox` paths reset all three fields and report the no-write outcome.
13. **PASS** — both reader paths advance the partial-filter next-read cursor after last fetched, while only applied events affect application tracking.
14. **PASS** — no public constructor/interface/options/serialized-state member changed; the existing EN/JA documentation remains aligned on `FetchedCount == 0`, reasons/cadence, `UsedCold`, and trigger/outcome separation.

## Scope, regression, and validation evidence

- The delta from prior reviewed head `75cb06f...` to this head is confined exactly to `MinimalOrleansTests.cs` and `MultiProjectionGrainCatchUpProductionPathTests.cs`; there is no production or documentation delta. The prior semantic PASS evidence for AC1–2 and AC4–14 is retained, and the helper changes do not alter those paths. The full PR remains within #1142's In Scope and does not cross its public-options, timeout/barrier, cold-export, or general persistence-recovery exclusions.
- The PR targets `main` and correctly says `Fixes #1142`. It remains consistent with `intents/sekiban/design/catchup-checkpoint-hot-path.md`: fetched traversal is distinct from a durable safe checkpoint, and a synthetic no-durable-effects range must not advance its restart cursor.
- `git diff --check 7e387e5e20d12f7d5686b6880b75583934accc75...9d8744f02414c2087e2a60b110389a7b77ecb13a` is clean.
- Re-run at exact head: the two repaired AC3 tests pass 2/2 on each of net9.0 and net10.0; `MultiProjectionGrainCatchUpProductionPathTests` passes 11/11 on each framework. These passing tests are necessary, not sufficient: the AC3b semantic assertion above is still absent.
- Terminal commit-scoped CI is green for `dcbTestsNet9`, `dcbTestsNet10`, `SonarCloud`, and `SonarCloud Code Analysis`, each reporting this exact head SHA.

tomohisa (commented):
# SEK-G37 final AC3b semantic rereview — APPROVE

Reviewed exact head `31bf9116a1d391e0f539ec2e501d4f6d9fb1914d` (base `7e387e5e20d12f7d5686b6880b75583934accc75`) against #1142, the operator-approved [AC3a/AC3b amendment](https://github.com/J-Tech-Japan/Sekiban/issues/1142#issuecomment-5376352138), updated `SEK-G37` `packet.yaml` and mandatory `review-context.md`, the design record, and prior review `#pullrequestreview-4998309877`.

**Workflow verdict: approve.** This is the required same-account `COMMENTED` review; it records the approve-equivalent workflow verdict rather than attempting GitHub's unavailable self-approval.

## AC3b durability/data-loss gate — PASS

The previous blocker is repaired entirely in `MultiProjectionGrainCatchUpProductionPathTests.cs`; the delta from prior reviewed head `9d8744f02414c2087e2a60b110389a7b77ecb13a` changes no production or documentation file.

`Synthetic_id_tracking_only_does_not_advance_c0_and_fresh_activation_reapplies_the_range` now proves both provenance-sensitive halves of AC3b:

- Synthetic zero-applied activation uses `unchangedC0Record` as its persistent record, preloads every distinct authoritative ID into its activation-local cache, and forbids host application. It reads C0, performs exactly one safe-checkpoint capture, reports `no_durable_write`, writes neither snapshot nor persistent state, and leaves the committed `LastSortableUniqueId` at C0. Cursor advancement remains a test failure, not success.
- The fresh `MultiProjectionGrain` has an empty cache and is constructed with the **same** `RecordingPersistentState<MultiProjectionGrainState>` object and the exact same authoritative event-list object (`Assert.Same` for both). Its first authoritative read is C0.
- The fixture compares the full ordered authoritative ID array with every ID received by the fresh host. Its deterministic materialized projection is `(C0, count, SHA-256 of C0 plus ordered IDs)` and must equal one application of that authoritative range; it must also differ from the empty C0 projection. Thus omission, duplication, reordering, or no state change fails rather than being hidden by a call count.
- The fresh run still has zero persistent-state writes and C0 remains committed, so no traversal cursor is converted into a restart cursor.

This satisfies the packet/design requirement that effects absent from a durable snapshot be replayed after restart rather than skipped.

## AC3a preservation — PASS

`MinimalOrleansTests.Hot_fetched_checkpoint_for_stream_preapplied_range_is_durable_and_consumed_by_fresh_activation` is unchanged by this repair and remains the production proof: stream delivery first, delayed safe-window eligibility, actual `SharedStateStore` C0 → C1 promotion, deactivation, and a real fresh activation whose first exclusive authoritative cursor is C1. It passes on net9.0 and net10.0.

## Original 14-AC map

1. **PASS** — enumerable zero-applied production batches remain active through 9 and cross exactly once at fetched batch 10.
2. **PASS** — the independent streaming reader has the equivalent production-path proof.
3. **PASS** — AC3a C1/fresh-read durability and AC3b C0/no-write/fresh exact replay are both proved as above.
4. **PASS** — hot fetched threshold boundaries use irregular batches and arbitrary modulo residue.
5. **PASS** — legacy hot modulo reason and precedence are preserved.
6. **PASS** — hot/cold counter reset cycles contain the discriminating intervening non-fire.
7. **PASS** — time reset includes the immediate non-fire before the second aged trigger.
8. **PASS** — non-zero seeded four-transition window normalization remains covered.
9. **PASS** — routing chooses cold thresholds only with `UsedCold == true`; hybrid hot tail stays private-hot.
10. **PASS** — cold segment/applied/interval precedence and fetched-only cases remain covered.
11. **PASS** — attempts and durable writes remain separately observed, including unchanged guarded/unguarded behavior.
12. **PASS** — success, unchanged, and failed-result persistence return paths reset all fields and report the no-write outcome.
13. **PASS** — partial-filter reads advance after last fetched on both readers while application tracking contains only applied events.
14. **PASS** — no public constructor/interface/options/serialized-state member changed; EN/JA documentation remains aligned on `FetchedCount == 0`, cadence/reasons, `UsedCold`, and trigger/outcome separation.

## Scope and validation

- The exact repair diff is one test file only; `git diff --check 7e387e5e20d12f7d5686b6880b75583934accc75...31bf9116a1d391e0f539ec2e501d4f6d9fb1914d` is clean. No source/docs drift, public API/options addition, serialized-state member, or traversal-cursor persistence was introduced. `PersistStateAsync` continues to commit the host safe position, not `_catchUpProgress.CurrentPosition`.
- Scope remains within #1142's In Scope, excludes its public-options, timeout/barrier, cold-export, and general failure-recovery exclusions, targets `main`, and correctly says `Fixes #1142`. It is consistent with `intents/sekiban/design/catchup-checkpoint-hot-path.md`'s separation of traversal cursor from durable safe checkpoint.
- Re-run at exact head: `MultiProjectionGrainCatchUpProductionPathTests` 11/11 on net9.0 and 11/11 on net10.0; `MultiProjectionGrainPersistPolicyTests` 12/12 on each; `MultiProjectionGrainCatchUpDiagnosticsTests` 6/6 on each; AC3a real-activation test 1/1 on each.
- Commit-scoped terminal CI is green on this exact SHA: `dcbTestsNet9`, `dcbTestsNet10`, `SonarCloud`, and `SonarCloud Code Analysis`.

Passing tests are necessary but not sufficient; the approval rests on the packet's AC3b non-advancement/lossless-replay contract and the explicit semantic proof above, together with the preserved prior evidence for the remaining criteria.

PR #1151, opened 2026-08-22
feat(dcb): scope projection fault lifecycle

### LINKED ISSUE #1140: A re-raised projection fault is indistinguishable from the original: no inner stack, and the stored position reads as if it were in the current batch
## Summary

Once a projection fault is recorded, every later call re-raises it as a `SekibanProjectionFaultException` built from the stored descriptor. That exception carries no inner exception and no original stack, and — more importantly — it is not distinguishable in a log from the first, genuine fault. Both render as "projector X is faulted at event Y, position Z: <message>" with a stack rooted at the apply entry point.

This is not only a diagnosability nicety. It produced a concrete misdiagnosis on a real incident, described below.

## The two paths

The first fault rethrows the ORIGINAL exception instance, stack intact:

`dcb/src/Sekiban.Dcb.Core/Actors/GeneralMultiProjectionActor.cs:788-801`

```csharp
private void CaptureFaultAndRethrow(Event ev, Exception ex)
{
    var original = ex.InnerException ?? ex;
    _fault ??= new ProjectionFaultDescriptor(
        ev.Id, ev.EventType, _projectorName, ev.SortableUniqueIdValue, original.Message, DateTime.UtcNow.Ticks);
    _fault.Annotate(original);
    ExceptionDispatchInfo.Capture(original).Throw();
}
```

Every subsequent call short-circuits at the top of the apply surface and throws a freshly constructed exception instead:

`dcb/src/Sekiban.Dcb.Core/Actors/GeneralMultiProjectionActor.cs:97-100,172-185`

```csharp
public async Task AddSerializableEventsAsync(...)
{
    ThrowIfFaulted();
    ...
}

private void ThrowIfFaulted()
{
    if (_fault is not null) throw CreateFaultException(_fault);
}

private static SekibanProjectionFaultException CreateFaultException(ProjectionFaultDescriptor fault)
{
    var exception = new SekibanProjectionFaultException(fault);
    fault.Annotate(exception);
    return exception;
}
```

`SekibanProjectionFaultException` takes only a descriptor — there is no `innerException` parameter:

`dcb/src/Sekiban.Dcb.Core/Actors/SekibanProjectionFaultException.cs`

```csharp
public SekibanProjectionFaultException(ProjectionFaultDescriptor fault) : base(fault.Describe()) => Fault = fault;
```

and `ProjectionFaultDescriptor` stores the failure as a plain string with no stack (`ProjectionFaultDescriptor.cs:24` — `string Message`).

The type's own doc comment already states this is the re-raise path. The problem is that nothing in the *emitted* exception says so.

## The misdiagnosis it caused

On a deployed host, a projection faulted with an `OutOfMemoryException` during a full rebuild and the descriptor was persisted. Later, a narrow catch-up was requested over a recent window — roughly 46,000 events spanning four days. It failed, and the log showed:

```
[<Projection>] Catch-up batch error.
  CurrentPosition=<position P1>,   ← start of the requested window
  TargetPosition =<position P2>,   ← end of the requested window

Sekiban.Dcb.Actors.SekibanProjectionFaultException:
  projector '<Projection>' is faulted at event <id> (<EventType>),
  position <position P0>:          ← FIVE MONTHS BEFORE P1
  Exception of type 'System.OutOfMemoryException' was thrown.
   at Sekiban.Dcb.Actors.GeneralMultiProjectionActor.AddSerializableEventsAsync(...)
   at ...ProcessStreamingSerializableBatch(... String startPosition ...)
   at ...ProcessSingleCatchUpBatch()
```

The reasonable reading — and the one the operator filed — is that the cold read returned an event far outside the requested window. That produced a bug report against `HybridEventStore` / the cold segment selection path.

It was wrong. The exception is the stored descriptor being re-raised at the top of `AddSerializableEventsAsync`; the buffer for that batch was never inspected. A correctly-filtered, in-window buffer produces exactly this log. Confirming that took a full audit of the cold read path (segment skip at `HybridEventStore.cs:849-851`, per-line filter in the JSONL handler at `JsonlColdSegmentFormatHandler.cs:49-65`, and the grain-side filter at `MultiProjectionGrain.cs:4129-4141`) purely to establish that the reported symptom could not be what it looked like.

A second consequence: the same report concluded the fault was memory-independent, because the process sat at 133–330 MB on a 16 GiB plan while "failing at the same event in about one second". That is also the short-circuit — no work is done, so no memory moves. The plan upgrade was never actually exercised.

## What would have prevented it

Anything that marks the re-raise as a re-raise. In rough order of value:

1. **Distinguish the re-raise in the message or a property** — e.g. "previously faulted at …", or an `IsReRaise` / `FaultedAtUtc` surfaced in `Describe()`. A reader then knows the position refers to a past fault, not the current batch.
2. **Carry the original stack.** Pass the original exception as `innerException` when the fault is first captured and store enough to reconstruct it, or record the original stack string on the descriptor. Today the descriptor keeps only `Message`.
3. **Log the first fault distinctly from subsequent ones**, so the one occurrence that carries the real stack is findable after the fact.

Item 1 is the one that actually prevents the misreading; items 2 and 3 shorten the investigation once someone is looking.

## Note on what is already there

The original stack is not lost at the moment of the fault — `CaptureFaultAndRethrow` rethrows the original instance, and a caller that logs the exception object records it. So the information exists in the log line for the *first* failure. The difficulty is that after that point every log line looks the same, and an operator arriving days later has no way to tell which one was first, or that the later ones are echoes.

## What is verified and what is not

Verified by reading `origin/main` (`cca3c8d64c5a5a013b8ce4e5197c160b26423864`): the two paths above, the exception and descriptor shapes, and the three cold-read filters that rule out the out-of-window reading.

Not verified: the deployed host's binaries were reported as 10.17.0; I have not confirmed they are byte-identical to this `origin/main`.

### LINKED ISSUE #1141: The reset token for ResetProjectionFaultAsync is not exposed by any read surface: recovery requires reading grain state directly or catching a deliberate query failure
## Summary

`IMultiProjectionGrain.ResetProjectionFaultAsync` is the only way an operator can clear a persisted projection fault, and it requires a three-part concurrency token: projector name, fault event id, and fault stream position. None of the read surfaces expose the last two. The documented way to obtain them is to catch a failing query's exception and read its `Data` — which means the recovery path for a stuck projection is only reachable by writing code that deliberately triggers and inspects a failure.

## The token

`dcb/src/Sekiban.Dcb.Orleans.Core/Grains/ResetProjectionFaultRequest.cs`

```csharp
public sealed record ResetProjectionFaultRequest(
    string ProjectorName,
    string FaultEventId,
    string FaultPosition);
```

The interface doc is explicit that it must not be synthesised:

`dcb/src/Sekiban.Dcb.Orleans.Core/Grains/IMultiProjectionGrain.cs:125-159`

> Acquire the token from the fault context surfaced on a failed query (the projection-fault error carries the event id, type, projector and position); do not synthesise it.

That instruction is correct and worth keeping — a stale token must not be accepted. The problem is that "a failed query's exception" is the *only* source.

## What the read surfaces expose

`MultiProjectionHealthStatus` carries `LastError` (a message string) and no fault identity at all:

`dcb/src/Sekiban.Dcb.Orleans.Core/Grains/MultiProjectionHealthStatus.cs:8-40` — `IsInitialized`, `HasProjectionActor`, `EventsProcessed`, `LastError`, `IsCatchUpActive`, `LastPersistTime`, `LastSortableUniqueId`, `PendingStreamEvents`, `StateRestoredAt`, `StateRestoreSource`, `IsHealthy`.

`FaultEventId` appears in only five files across `dcb/src`, none of which is a read surface:

```
dcb/src/Sekiban.Dcb.Core/Actors/ProjectionFaultDescriptor.cs
dcb/src/Sekiban.Dcb.Orleans.Core/Grains/CoordinatedGrainStateStore.cs
dcb/src/Sekiban.Dcb.Orleans.Core/Grains/MultiProjectionGrain.cs
dcb/src/Sekiban.Dcb.Orleans.Core/Grains/MultiProjectionGrainState.cs
dcb/src/Sekiban.Dcb.Orleans.Core/Grains/ResetProjectionFaultRequest.cs
```

The values are durably present — `MultiProjectionGrainState` persists `FaultEventId` / `FaultEventType` / `FaultPosition` / `FaultMessage` / `FaultedAtUtcTicks` (`MultiProjectionGrainState.cs:95-107`) through `[PersistentState("multiProjection", "OrleansStorage")]` (`MultiProjectionGrain.cs:268`). There is simply no method that returns them.

The projection-status surface does not carry them either, so an operations dashboard built on `IProjectionStatusReader` can show that a projector is faulted and why, but not the identity needed to clear it.

## Why this bites in practice

On a real incident the operator could see `phase: faulted` and the fault message on their dashboard, tried the obvious remediations — process restart, deleting the projection-status row, reverting the projector version, resizing the host — and concluded in a written report that no recovery mechanism existed. `ResetProjectionFaultAsync` had been there the whole time; it was unreachable because nothing surfaced its inputs.

Recovering the token required reading the Orleans grain-state document directly out of the backing store. That works, but it is a database-level operation being used to drive a supported API, and it differs per storage provider.

The alternative the docs point to — trigger a query, catch the exception, read `Exception.Data["Sekiban.Projection.FaultEventId"]` and `["Sekiban.Projection.FaultPosition"]` (`ProjectionFaultDescriptor.cs:38-42`) — means every host that wants an operator recovery button has to implement a deliberate failure-and-inspect step. That is an odd shape for the intended path, and it is easy to get subtly wrong.

## Suggested direction

Add a read surface that returns the current persisted fault identity for a projector — projector name, fault event id, fault event type, fault position, fault message, faulted-at time. Either as fields on `MultiProjectionHealthStatus`, or as a dedicated admin-plane read next to `ResetProjectionFaultAsync`.

The concurrency property is unaffected: the reset already validates the token against the current persisted descriptor inside the single-writer gate, so a token read a moment earlier and then invalidated is still correctly rejected. Exposing the value for reading does not weaken that check — it just lets an operator see what they are about to act on.

If keeping the admin plane deliberately narrow is preferred, a `TryGetProjectionFaultAsync()` returning the descriptor (or none) on the same admin-only surface would be enough.

## What is verified and what is not

Verified by reading `origin/main` (`cca3c8d64c5a5a013b8ce4e5197c160b26423864`): the request shape, the interface documentation, the full set of files mentioning `FaultEventId`, the health-status record's fields, and the grain-state persistence attribute.

Additionally verified against a deployed host: the persisted fault document exists in the Orleans grain-state store with all three token fields populated, and reading it there was what made the reset reachable.

Not verified: whether the deployed binaries are byte-identical to this `origin/main`; they were reported as 10.17.0.

### LINKED ISSUE #1144: A persisted projection fault survives a projector version change: the restore path never compares the recorded ProjectorVersion
## Summary

A persisted projection fault survives a projector version change. The grain-state document that holds it is keyed by projector name only, and the restore path does not compare the recorded `ProjectorVersion` against the running one. Changing the projector version — the ordinary way to make a projection rebuild under different code — therefore does not escape a fault recorded under the previous version.

This matters because bumping or reverting the projector version is one of the first things an operator reaches for when a projection is stuck, and it silently does nothing.

## The restore path ignores the version

`dcb/src/Sekiban.Dcb.Orleans.Core/Grains/MultiProjectionGrain.cs:3404-3427`

```csharp
private void RestoreProjectionFaultIfPersisted()
{
    var state = _stateStore.Committed;
    if (state?.FaultEventId is null || _host is null)
    {
        return;
    }

    if (!Guid.TryParse(state.FaultEventId, out var eventId))
    {
        return;
    }

    var fault = new ProjectionFaultDescriptor(
        eventId,
        state.FaultEventType ?? string.Empty,
        GetProjectorName(),
        state.FaultPosition ?? string.Empty,
        state.FaultMessage ?? string.Empty,
        state.FaultedAtUtcTicks);

    _projectionFault = fault;
    _host.RestoreFault(fault);
}
```

`_stateStore.Committed` carries a `ProjectorVersion` (it is written on every persist, `MultiProjectionGrain.cs:1454`, `:1571`), but this method never reads it. It is called unconditionally during activation (`MultiProjectionGrain.cs:2446`).

## Version-scoped state is cleared; the fault is not

The activation path already recognises that derived state must not survive a version change. When no external snapshot exists for the running version, it resets the integrity-guard fields:

`dcb/src/Sekiban.Dcb.Orleans.Core/Grains/MultiProjectionGrain.cs:2386-2402`

```csharp
// Reset integrity guard fields when external snapshot is missing.
// Without this, LastGoodSafeVersion from a previous successful run
// permanently blocks persist because catch-up starts at safeVersion=0.
if (_stateStore.Committed.LastGoodSafeVersion > 0)
{
    _logger.LogWarning("Resetting integrity guard: LastGoodSafeVersion was {LastGood} but external snapshot is missing. ...");
    await _stateStore.ExecuteWriteAsync(GrainStateWriteKind.MetadataMaintenance, s =>
    {
        s.LastGoodSafeVersion = 0;
        s.LastGoodPayloadBytes = 0;
        s.LastGoodOriginalSizeBytes = 0;
        s.LastGoodEventsProcessed = 0;
    });
}
```

`FaultEventId`, `FaultEventType`, `FaultPosition`, `FaultMessage` and `FaultedAtUtcTicks` live in the same `MultiProjectionGrainState` (`MultiProjectionGrainState.cs:95-107`) and are not touched here or anywhere else on a version change. So the four high-water-mark fields are treated as version-scoped derived state while the fault descriptor beside them is treated as permanent.

## Observed consequence

From an operator report on a deployed host. A projection faulted under version `1.0.22`. Among the remediations attempted was reverting the projector code to `1.0.21` and redeploying — a version for which a healthy snapshot still existed. After the redeploy the grain read the same grain-state document, restored the `1.0.22` fault, and failed at the identical event id and position. The projection-status row then showed `projectorVersion: 1.0.21` with `isFaulted: true` and the `1.0.22` fault message.

The operator's summary of that attempt:

> プロジェクタのバージョンを 1.0.22 から 1.0.21 へ戻すコードリバートと再デプロイ … 毎回まったく同じイベント ID・同じ位置で即座に失敗

The fault was only escaped by deleting the grain-state document by hand.

## Why the current behaviour is defensible but incomplete

Keeping the fault across restarts is clearly intentional — `ThrowIfFaulted` exists so a retry cannot re-run the pre-poison events, and re-raising on activation keeps a query failing closed rather than answering from a partial state. None of that is in question.

What seems unintended is that a *different projector implementation* inherits it. The event that poisoned version 1.0.22 may well fold cleanly under 1.0.21 or under 1.0.23; the whole point of a version change is that the fold logic differs. Carrying the fault across that boundary removes the most natural recovery path and leaves only direct document deletion, which is storage-provider specific and cannot clear the live actor.

## Suggested direction

Compare the recorded `ProjectorVersion` against the running one in `RestoreProjectionFaultIfPersisted`, and skip the restore (clearing the fault fields as version-scoped derived state) when they differ — the same treatment `LastGoodSafeVersion` already gets. A log line naming both versions would make the transition auditable.

If carrying the fault across versions is deliberate, it would help to say so at the field or method level, and to make the version-crossing visible when the fault is re-raised, so an operator reverting a version can tell that the fault they are seeing predates the code they are running.

A regression test worth adding: record a fault under version A, restart the grain with version B, and assert the fault does not gate version B's first apply.

## Verification status

Verified by reading `origin/main` (`cca3c8d64c5a5a013b8ce4e5197c160b26423864`): the restore method and its unconditional call site, the presence of `ProjectorVersion` on the committed state, the integrity-guard reset that omits the fault fields, and the fault fields' definition on `MultiProjectionGrainState`.

Additionally verified against the deployed host: the grain-state document id contains the projector name and no version component, and the document carried the fault fields together with `ProjectorVersion: 1.0.22-…` while `1.0.21` code was running.

Not verified: whether the deployed binaries are byte-identical to this `origin/main`; they were reported as 10.17.0.

### LINKED ISSUE #1150: SEK-G39 dcb: projection fault lifecycle — re-raise identity, reset-token read surface, version scoping
# SEK-G39 dcb: projection fault lifecycle — re-raise identity, reset-token read surface, version scoping

Fixes #1140.
Fixes #1141.
Fixes #1144.

## Goal

Make a persisted projection fault diagnosable, recoverable, and version-scoped:
a re-raised fault must say it is a re-raise, the reset token must be readable
through a supported surface, and a fault must not survive the projector
version change that is the operator's primary remediation.

## Why This Slice Exists Now

All three defects compounded one real incident: a re-raise was misread as a
fresh cold-read bug (false bug report, false memory-independence conclusion),
the operator concluded no recovery mechanism existed because the reset token
was unreachable, and reverting the projector version silently did nothing.

## Current Observed State

Verified on `origin/main` (SEK-G37/G38 merged):

- First fault rethrows the original exception (stack intact); every later call
  builds `SekibanProjectionFaultException` from the stored descriptor with no
  stack and a message identical to a first fault.
- `ProjectionFaultExceptionConverter.ConvertFromSurrogate` does not re-apply
  the `Data` annotations on the client side.
- No read surface exposes `FaultEventId` / `FaultPosition`; the documented
  source is failing a query on purpose and reading `Exception.Data`.
- `RestoreProjectionFaultIfPersisted` never compares the recorded
  `ProjectorVersion` with the running one, while the adjacent `LastGood*`
  fields are treated as version-scoped.

- `IsReRaise` additive get-only property, always true on that type; the first
  fault never carries the marker. Distinct first-fault log event id.
- Annotations (four existing keys + re-raise key) applied idempotently on
  construction and surrogate reconstruction; proven client-side.
- `TryGetProjectionFaultAsync()` as a **default interface method** whose
  default body returns explicit unsupported (never "no fault"); returns a new
  `[GenerateSerializer]` DTO, four distinguishable result states, read from
  `_stateStore.Committed` only.
- Version-mismatch restore performs one awaited `MetadataMaintenance` clear,
  published before catch-up/query/reset/admin-read; write failure fails the
  activation (fail-before-serving, per SEK-G38 discipline). Match restores as
  today.
- Original-stack persistence is gated on frozen 10.17.0 binary vectors; if
  they cannot be produced, that part is split out (pre-approved).

## Target Repo / Path / Part

- Target part: fault lifecycle across `SekibanProjectionFaultException`, the exception surrogate/converter, `IMultiProjectionGrain` admin read with Orleans DTO, and the version-scoped restore transition in `MultiProjectionGrain`.

## In Scope

Everything in the Accepted Baseline and the seventeen acceptance criteria
below, with EN/JA docs.

## Out Of Scope

- Automatic fault recovery or auto-reset.
- Widening `MultiProjectionHealthStatus`.
- Recovering stacks for faults persisted before this fix (impossible).
- Any change to reset validation semantics (read surface must not weaken it).

The child PR must mark re-raises distinctly (message, property, Data key,
log ids) surviving client-side Orleans deserialization; add the committed-only
admin read as a default interface method with an Orleans DTO and four result
states; clear persisted fault fields durably on a version-mismatch restore
with fail-before-serving ordering; keep same-version restore intact; gate
stack persistence on frozen 10.17 vectors or split it out; and prove all of it
with the seventeen criteria — additive public API only.

## Acceptance Criteria

- Re-raise kills the misdiagnosis: windowed catch-up on a persisted fault yields a machine-distinguishable re-raise whose position reads as outside the window.
- First fault unchanged: original instance, stack, four annotations, no marker.
- Client-side deserialization: marker and Data keys survive actual client-side Orleans deserialization; idempotent annotation on both construction paths.
- First-fault log id exactly once; re-raises distinct.
- Four admin-read states distinguishable; null/none = no committed descriptor only.
- Committed-only proof: live-but-uncommitted fault reads as no committed fault.
- Proxy vs legacy fake: fake returns explicit unsupported; generated proxy reaches the real grain (integration test).
- A-to-B stale-token race, full six steps, zero mutation, B retained.
- Token round-trip incl. legacy Exception.Data path.
- Version-mismatch transition: cleared durably before B's first apply; admin read reports no fault.
- Mismatch do-it-twice: old fault never reappears.
- Same-version restore still fails closed.
- Fail-before-serving on injected clear failure: no apply, fields unchanged, no successful no-fault response.
- Both directions + audit with both exact versions; distinct failure telemetry.
- Stack persistence only with frozen 10.17 vectors, field-id pins, bounded round trip — else split out.
- Additive public API; descriptor positional count unchanged; reflection pins.
- EN/JA docs (re-raise reading, token flow, version clearing, upgraded-silo requirement).

## Verification

## Related Links

- Host design record: intents/sekiban/design/projection-fault-observability.md
- Consultations: reports/sek-g39-design-consult.md, reports/sek-g39-design-final.md, reports/sek-g39-design-final2.md

## Knowledge Maintenance

- Intent placement: host design doc projection-fault-observability.md.
- Docs update: EN/JA fault lifecycle guidance — in scope.
- Closeout writeback expected: yes — stack split-out outcome, activation ordering vs G38 retirement, converter idempotency shape, remaining live-vs-committed conflations.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
## Summary

- Makes persisted projection-fault re-raises machine distinguishable and preserves typed annotations across real Orleans ResultBox transport.
- Adds committed-only admin fault reads and version-scoped one-write durable clearing before replacement activation work.
- Covers reset-token races, fail-before-serving, both version directions, client deserialization, and EN/JA operator docs.

## Validation

- Focused fault suites: net9.0 and net10.0 passed (37 reset + 11 production/transport + 10 integrity each).
- `git diff --check` passed.

## Scope note

OriginalStackTrace persistence is intentionally split out: reliable frozen dcb-v10.17.0 state and surrogate binary vectors are unavailable in this repository/tag, so no state or surrogate stack field was added.

Closes #1150
Fixes #1140
Fixes #1141
Fixes #1144


tomohisa (commented):
Verdict: **APPROVE**

Reviewed SEK-G39 / PR #1151 read-only at exact head `0f92265fb4296ce622aae0978428dd53998d79ba` against issue #1150, the canonical `packet.yaml`, mandatory `review-context.md`, `implementation.md`, and the referenced `projection-fault-observability.md` design. Tests are necessary, not sufficient: this approval is based on the production paths and mutation-killing discriminants below as well as their execution.

## AC1-AC17 evidence matrix

| AC | Production code path and killing discriminant | Test/execution evidence | Result |
|---|---|---|---|
| **AC1 Re-raise misdiagnosis** | `SekibanProjectionFaultException` uses `DescribeReRaise()` and `GeneralMultiProjectionActor.CreateFaultException`; `IsReRaise` and the Data marker identify a stored fault. A later tail event differs from the persisted fault position, killing attribution to the requested window. | `ReRaise_DeserializedOnTheClient_HasTheMarkerAndAllFaultAnnotations` uses a replacement activation and real client proxy, then proves the stored position is outside the later tail. | **PASS** |
| **AC2 First fault unchanged** | Both `CaptureFaultAndRethrow` boundaries annotate the original four keys and rethrow with `ExceptionDispatchInfo`; only descriptor-reconstructed exception instances receive the re-raise marker. | `PoisonEvent_FaultsTheProjection_AndSurfacesTheOriginalException` proves non-wrapper original, stack present, four annotations, and no marker. | **PASS** |
| **AC3 Client Orleans deserialize/idempotency** | `ResultBoxSurrogateConverter` transports a typed `ProjectionFaultExceptionSurrogate`; the exception constructor and `ConvertFromSurrogate` both call idempotent `AnnotateReRaise`, while `ConvertToSurrogate` preserves the typed descriptor. | The actual `TestCluster.Client` result is a client-side `SekibanProjectionFaultException` with all five Data entries; the converter test applies annotation a third time and still has exactly five entries. A server-only object cannot satisfy the real proxy test. | **PASS** |
| **AC4 Dedicated logs** | Actor-local EventIds 1401/1402 separate first observation (original exception/stack) from reconstructed re-raises; the `_fault is null` transition gates the first log. | `FirstFault_LogsTheOriginalOnce_AndReRaisesUseADistinctEventId` finds exactly one 1401 with the same thrown exception and two distinct 1402 re-raises without an exception. | **PASS** |
| **AC5 Four read states** | `TryGetProjectionFaultAsync` returns success+DTO or canonical success+none; malformed committed state becomes read error; the interface default is explicit `NotSupportedException`. | Real-proxy fault/no-fault, direct legacy fake unsupported, and malformed-committed read-failure tests independently distinguish all four states. | **PASS** |
| **AC6 Committed-only** | Admin read takes only `_stateStore.Committed`; it never reads `_projectionFault` or host state. | `LiveButUncommittedFault_AdminReadReportsOnlyNoCommittedDescriptor` rejects the first descriptor write, proves the live query is faulted, then observes supported+no-committed-fault. Reading live state would fail this test. | **PASS** |
| **AC7 Proxy vs default body** | Default interface body always returns unsupported; generated Orleans proxy dispatches to the grain implementation. | `LegacyDirectImplementation_DefaultAdminReadIsExplicitlyUnsupported` versus `ProjectionFaultAdminRead_UsesTheRealProxyForFaultAndNoFault`; a proxy falling into the default body cannot return either successful real-grain result. | **PASS** |
| **AC8 Six-step A-to-B race** | Existing reset validation remains inside the committed single-writer transition. | `AdminReadTokenA_DoesNotClearDurablyPersistedFaultB_ThenTokenBResets`: read A; reset A; durably establish/read B; submit A; assert rejection, zero provider writes, zero external deletes, exact B retained and freshly readable; reset B. The simpler clear/resubmit test also remains. | **PASS** |
| **AC9 Token round-trip** | DTO identity maps verbatim to the unchanged reset request; legacy Data keys remain. | `AdminReadToken_RoundTripsToReset_AndTheLegacyExceptionDataTokenStillWorks` proves canonical token acceptance, no-fault/read behavior, stale resubmit rejection, and legacy Data-token acceptance. | **PASS** |
| **AC10 Version mismatch** | `RestoreProjectionFaultIfPersistedAsync` compares committed and running versions before fault restoration, performs one awaited `MetadataMaintenance` candidate write clearing exactly five fault fields, and returns only after the new committed snapshot is published; catch-up starts afterward. | Bump/revert theory records exactly one provider write, inspects that committed record for surgical five-field clear, then observes the real projector's first B apply after that write and a successful supported+no-fault admin read. | **PASS** |
| **AC11 Do it twice** | The clear is provider-durable, not actor-local. | The same mismatch test deactivates B, waits for the provider deactivation write, creates a second fresh B activation, reads no fault again, and observes no second transition log. | **PASS** |
| **AC12 Same-version restore** | Equality branch reconstructs the descriptor into the fresh host without a clear write. | `SameVersion_ReactivationRestoresTheFaultAndContinuesFailingClosed` observes fault DTO, unchanged write count, no clear audit, and continued failed query. | **PASS** |
| **AC13 Fail before serving** | A mismatch write exception is logged as 1028 and escapes activation before first-query gate, background catch-up, subscription, or any method response. | Injected provider failure yields an RPC/activation exception, zero new successful writes/history, exact committed fault retention, zero real-projector apply, no successful no-fault result, failure telemetry only, and no success audit. | **PASS** |
| **AC14 Both directions/audit** | Ordinal inequality handles either direction; success log includes persisted and running versions, while write failure uses a separate EventId. | Theory seeds `0.9 -> 1.0` and `2.0 -> 1.0`, asserts both exact strings in the single 1027 audit, and the failure case independently pins 1028. | **PASS** |
| **AC15 Stack gate/split-out** | The pre-approved split-out fired, as stated in the PR body. No `OriginalStackTrace`/stack persistence property, grain-state field, or exception-surrogate stack field was added; existing state ids 0-20 and exception-surrogate ids 0-5 were not extended for stack. | Full diff/source search confirms no stack persistence link or weakened frozen-vector substitute. Because the gated feature is absent, frozen 10.17 vectors are not claimed or required for this slice. | **PASS** |
| **AC16 Additive API** | Existing descriptor six-argument ctor/Deconstruct and exception ctor remain; additions are the get-only marker, default interface method, transport DTOs, annotation helpers, and ResultBox surrogate fields. No public member was removed. | `ProjectionFaultPublicShape_IsAdditive_AndDescriptorKeepsSixPositionalFields` pins six ctor/Deconstruct parameters and read-only marker; base-to-head public-surface diff confirms additive signatures. | **PASS** |
| **AC17 EN/JA docs** | Both common-issues guides document re-raise interpretation, committed read -> reset flow, read outcome semantics, upgraded-silo requirement, and version-change durable clearing/failure. | Direct EN/JA diff inspection covers every required topic and keeps automatic recovery/reset explicitly excluded. | **PASS** |

## Compatibility and harness assessment

- `ResultBoxSurrogate<T>` keeps ids 0-3 unchanged and adds only ids 4 (`IsProjectionFault`) and 5 (`ProjectionFault`). New code reading an old payload sees id 4's default `false` and follows the unchanged legacy error-string path; old code reading a new payload can skip unknown 4/5 while the converter still supplies ids 2/3. This is additive old-payload compatibility. AC15's frozen-vector gate applies to the split-out stack/state/surrogate persistence, not this additive typed-error envelope, so a new frozen vector is not required by the packet. A field-id pin would be useful future drift protection, but its absence is not a current contract failure because the base/current schema diff directly proves 0-3 unchanged and 4/5 unique.
- The real-cluster harness uses separate per-class ranges (46k, 53k, 59k), PID-derived pairs, and collections marked `DisableParallelization=true`; silo shutdown occurs between tests. Both local TFM runs completed all 48 real-cluster focused cases without a port failure. The setup is acceptably reliable for this CI scope.

## Cross-cutting verification

- Packet/design alignment: the implementation follows SEK-G39's re-raise identity, committed-only read, unchanged reset validation, and version-scoped durable-clear ordering. Relative to SEK-G38, external restore/integrity retirement completes first; G39's fault restoration/mismatch clear runs after `base.OnActivateAsync` but still before catch-up, subscription, and request serving.
- Out-of-scope boundaries were not crossed: no automatic recovery/reset, no `MultiProjectionHealthStatus` expansion, no historical-stack recovery, no reset-token validation weakening, and no release action.
- PR metadata: base `main`, non-draft, exact head above; body retains `Closes #1150`, `Fixes #1140`, `Fixes #1141`, and `Fixes #1144`.
- Local exact-head execution: focused Orleans production/transport/reset/persist-failure suites passed **48/48** on net9 and **48/48** on net10; actor integrity passed **10/10** on net9 and **10/10** on net10. `git diff --check` is clean.
- Exact-head CI: `dcbTestsNet9`, `dcbTestsNet10`, `SonarCloud`, and `SonarCloud Code Analysis` are all SUCCESS. Sonar's four findings are CODE_SMELL/maintainability only, not reliability/BUG: three boolean assertion-style findings are informational, and the suggestion to make the required instance `IsReRaise` API static is rejected as contrary to AC1/AC16. Quality gate passed and there are no security hotspots.

No implementation finding remains. No push, merge, or label mutation was performed.

PR #1155, opened 2026-08-22
perf(dcb): defer out-of-order repair to consumption

### LINKED ISSUE #1146: RebuildSafeState re-folds the entire retained safe history per out-of-order event, making a rebuild-from-beginning quadratic in event count
## Summary

In the non-incremental mode that a rebuild-from-`beginning` selects, every out-of-order safe event re-folds the **entire** retained safe history from the initial payload. Each occurrence is O(N) in the events retained so far, so a rebuild with a steady fraction of out-of-order arrivals costs O(N²) overall.

Measured on `origin/main`: doubling the event count roughly quadruples the time. Extrapolating the measured curve to a 2.6M-event rebuild gives about three hours — with a projector whose fold is a single integer addition.

## The path

`dcb/src/Sekiban.Dcb.Core/MultiProjections/DualStateProjectionWrapper.cs:186-191`

```csharp
else if (!_useIncrementalSafePromotion)
{
    // Fresh path still retains the FULL safe history in _allSafeEvents, so a global re-sort places the
    // out-of-order arrival correctly (no compacted baseline, no data loss).
    RebuildSafeState(domainTypes);
    _safeVersion++;
}
```

`dcb/src/Sekiban.Dcb.Core/MultiProjections/DualStateProjectionWrapper.cs:368-410`

```csharp
private void RebuildSafeState(DcbDomainTypes domainTypes)
{
    if (_allSafeEvents.Count == 0) { return; }

    var allEvents = _allSafeEvents.Values.ToList();
    allEvents.Sort((a, b) => string.Compare(
        a.SortableUniqueIdValue, b.SortableUniqueIdValue, StringComparison.Ordinal));

    var rebuiltProjector = _types.GenerateInitialPayload(_projectorName);
    ...
    foreach (var ev in allEvents)
    {
        var tags = ev.Tags.Select(tagString => domainTypes.TagTypes.GetTag(tagString)).ToList();
        var projected = _types.Project(_projectorName, newSafeProjector, ev, tags, domainTypes, /* zero threshold */);
        ...
    }
    ...
}
```

Per occurrence this is a full copy of the retained dictionary, an O(N log N) sort over 30-character strings, a fresh initial payload, and a re-fold of every retained event — plus a `Tags.Select(...).ToList()` allocation per event inside the loop.

The mode is selected by which constructor runs. A rebuild from an initial payload takes the non-incremental branch:

`dcb/src/Sekiban.Dcb.Core/MultiProjections/DualStateProjectionWrapper.cs:77`

```csharp
_useIncrementalSafePromotion = false;
```

while a wrapper restored from a snapshot sets it to `true` (`:103`) and signals an authoritative rebuild instead of re-folding locally (`:191-197`).

## Measurement

A temporary probe on `origin/main` (`7e387e5e`), using a minimal projector whose `Project` is `Total + Amount`, run on an Apple Silicon Mac. 1% of events arrive with a position earlier than the current safe head; the rest are in order. Everything is inside the safe window, so no buffering is involved.

**Total time for a full pass**

| Events | Elapsed | Ratio vs previous |
| ---: | ---: | ---: |
| 1,000 | 2 ms | — |
| 2,000 | 9–12 ms | ~5x |
| 4,000 | 28–30 ms | ~3x |
| 8,000 | 103–107 ms | ~3.6x |

Doubling N multiplies elapsed time by roughly four.

**Cost of one out-of-order event, against retained history**

| Retained events | One out-of-order event |
| ---: | ---: |
| 1,000 | 0.8 ms |
| 4,000 | 3.6 ms |
| 8,000 | 8.5 ms |

Linear in the retained set, as the code implies.

**Extrapolation** from the 8,000-event point, assuming the quadratic holds:

| Events | Estimated |
| ---: | ---: |
| 100,000 | ~16 s |
| 500,000 | ~7 min |
| 2,600,000 | ~3 h |

That is for the trivial projector. A projector holding hundreds of thousands of entities does materially more work per fold, and the re-fold walks all of them every time.

## Why the retained set can grow without bound

`_allSafeEvents` is cleared only by `CompactSafeHistory()` (`DualStateProjectionWrapper.cs:234-244`), and the grain calls that only after a durable write succeeds (`MultiProjectionGrain.cs:1490-1492`, `:1593-1595`). When a persist is refused — for instance by the safeVersion regression guard discussed in #1139 — compaction is skipped while promotion still runs, so the retained set grows for the whole rebuild and each re-fold gets progressively more expensive.

The two issues compound: #1139 removes the bound on N, and this issue makes the cost quadratic in N.

## Field context

A host running this library reported a rebuild where per-batch apply time was orders of magnitude worse than the same data applied without the cold-storage read path. The application-side projection code is not the cause there — that application removed per-event dictionary copying from its projectors in an earlier optimisation round (a documented change that took one projector from ~44 to ~10,800 events/sec) and the current code still folds in place. The remaining quadratic term is this one, inside the library.

That report also observed that the same environment applied other projections at normal speed while one projection was ~30x slower, which argues against the storage path being the differentiator.

I am **not** claiming this fully explains that host's numbers — I have not reproduced their measurement, and their environment differs. The probe above establishes only that the quadratic exists on `origin/main` and that its constant grows with the retained history.

## Suggested direction

The comment on the branch states the intent: keep full history so an out-of-order arrival can be placed by a global re-sort without data loss. Correct, but the implementation re-derives the whole state to place one event.

Options worth considering:

- Fold the out-of-order event into a position-ordered structure and re-fold only from that position forward, rather than from the initial payload.
- Bound the retained history independently of persist success, so the re-fold cost has a ceiling even when checkpointing is blocked.
- Detect a run of in-order arrivals and compact opportunistically, rather than only at a successful durable write.

A regression guard worth adding: assert that a rebuild's elapsed time grows sub-quadratically across two event counts, so a future change cannot silently reintroduce this.

## Verification status

Verified on `origin/main` (`7e387e5e`): the code paths quoted above, and the timing table, which came from a probe file added to `dcb/tests/Sekiban.Dcb.WithResult.Tests/`, run, and then removed (the working tree was left clean).

The probe uses a synthetic 1% out-of-order rate. I have not measured what out-of-order rate real catch-up produces, so the extrapolated wall-clock figures should be read as illustrating the growth curve, not as predictions for any particular deployment.

### LINKED ISSUE #1154: SEK-G41 dcb: defer out-of-order repair to consumption
# SEK-G41 dcb: defer out-of-order repair to consumption

Fixes #1146.

## Goal

Remove the O(N²) cost of out-of-order arrivals during a rebuild, without
weakening the losslessness the current full re-fold provides.

## Why This Slice Exists Now

Measured on `origin/main`: doubling the event count roughly quadruples the
time; extrapolating to a 2.6M-event rebuild gives about three hours with a
projector whose fold is a single integer addition.

## Current Observed State

In the non-incremental mode selected by rebuild-from-`beginning`, **every**
out-of-order safe event triggers `RebuildSafeState`: a full copy of the
retained dictionary, an O(N log N) ordinal sort, a fresh initial payload, and
a re-fold of every retained event — plus a per-event tag-list allocation
inside the loop.

`SEK-G37`/`SEK-G38` bound N *between* checkpoints (G37 established the
fallbacks; G38 restored the blocked durable writes and the compaction that
follows them), but no bound exists before the first successful persist, while
persists keep failing, or on short-circuits with no durable write.

- **Dirty + repair-on-consume**: arrivals insert into a sorted structure in
  O(log N) and set a dirty flag only; repair happens exactly once at each
  consumption boundary through one private seam, reusing `RebuildSafeState`.
- **Total order** `(SortableUniqueId, EventId)`, ordinal.
- **Seven-point consumption inventory** including the direct unsafe getter,
  zero-event promotion, and an atomic repair→clear→switch before compaction.
- **Suppression rule**: a `ProcessEvent` that sets dirty must not run
  `ReconcileServedState` at the end of that event; a real consumption request
  performs `repair → atomic publish → ReconcileServedState → consume`.
- **Dirty is not only a payload state**: version, last ids and served/unsafe
  state are staged and published atomically with the payload.
- **Repair failure publishes nothing** and blocks compaction/switch/persist,
  keeping the causing event attributable.

## Target Repo / Path / Part

- Target paths: `dcb/src dcb/tests`
- Target part: `DualStateProjectionWrapper` non-incremental mode — sorted retained history with a dirty flag, one private repair seam at seven consumption boundaries, atomic payload+metadata publication, suppressed per-event reconcile while dirty.

## In Scope

Everything in the Accepted Baseline and the eight acceptance criteria below.

## Out Of Scope

- Changes to incremental mode.
- The snapshot restore path (`#1145`, SEK-G42).
- New public options or any public API change.
- Partial re-fold from the insertion point (rejected: a reusable prefix state
  is not guaranteed under the general projector-fold contract).

The child PR must replace per-arrival re-folding with a dirty flag plus a
single private repair seam reached from all seven consumption boundaries,
adopt the `(SortableUniqueId, EventId)` total order, suppress the per-event
reconcile while dirty and enforce the `repair → publish → reconcile → consume`
order, publish payload and metadata atomically, keep repair failure
non-publishing, and prove all of it with the eight criteria below — public API
unchanged.

## Acceptance Criteria

- Losslessness + ties: deferred repair equals immediate full re-fold exactly, including identical-SortableUniqueId ties.
- Suppression and order: no publication or reconcile before the first consume; the first consume repairs and reconciles exactly once each; a second consume advances neither; the order is repair → publish → reconcile → consume.
- Consumption-boundary matrix (primary killing test): each of the seven production entry points driven individually from a dirty seed, checking payload/version/position consistency, exactly one repair, and nothing published on failure; includes zero-event promotion, compaction-before-repair, the direct unsafe getter, and failure/retry.
- Complexity by deterministic fold count: zero refolds during insertion, exactly one rebuild per consume, total folds linear in N (old implementation is ~1+2+…+N and must fail); repeat consume adds zero; a new out-of-order epoch adds exactly one; compare N vs 2N with consumption count fixed. General complexity stated as O(N log N + N × consumption epochs).
- Wall-clock retained only as auxiliary evidence, never the primary gate.
- G37/G38 interaction preserved (post-checkpoint compaction and incremental switch).
- Per-event allocation removal proven by counting tag resolutions.
- Public API unchanged (reflection pins).

## Verification

All new and existing dcb tests green; `git diff --check` clean; no public signature additions, removals or modifications.

## Related Links

- Host design record: intents/sekiban/design/ooo-refold-quadratic.md
- Consultations: reports/sek-g41-design-consult.md, reports/sek-g41-design-final.md, reports/sek-g41-design-final2.md

## Knowledge Maintenance

- Intent placement: host design doc ooo-refold-quadratic.md.
- Docs update: not required (internal wrapper behavior; the release note lands with dcb-v10.19.0).
- Closeout writeback expected: yes — seam location and boundary reachability, scope of the reconcile suppression, staging representation, measured fold counts, any further consumption path found.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Replace arrival-time re-folding with sorted retained history and one private consumption seam.

Closes #1154
Fixes #1146

### DIFF
--- a/dcb/src/Sekiban.Dcb.Core/Actors/GeneralMultiProjectionActor.cs
+++ b/dcb/src/Sekiban.Dcb.Core/Actors/GeneralMultiProjectionActor.cs
@@ -210,14 +210,9 @@ public class GeneralMultiProjectionActor
 
         if (_singleStateAccessor is IDualStateAccessor dualAccessor)
         {
-            try
-            {
-                dualAccessor.PromoteBufferedEvents(safeWindowThreshold, _domain);
-            }
-            catch (Exception)
-            {
-                // Head-status reads are observational; ignore best-effort safe-window promotion failures here.
-            }
+            // A deferred safe-history repair is a consumption boundary. Returning an old head after its repair failed
+            // would falsely advertise a stale payload/position as current, so propagate the failure to the caller.
+            dualAccessor.PromoteBufferedEvents(safeWindowThreshold, _domain);
 
             var current = new ProjectionPosition(
                 dualAccessor.UnsafeVersion,
@@ -627,14 +622,15 @@ public class GeneralMultiProjectionActor
         int offloadThresholdBytes = int.MaxValue,
         CancellationToken cancellationToken = default)
     {
-        // Promote buffered events before building a safe snapshot
+        // Promote buffered events before building a safe snapshot. A deferred-repair failure must fail closed: emitting a
+        // snapshot from the old payload would allow persistence and later compaction to discard the retained repair input.
         try
         {
             ForcePromoteBufferedEvents();
         }
-        catch
+        catch (Exception exception)
         {
-            // Ignore promotion errors to avoid blocking persistence
+            return ResultBox.Error<SnapshotPersistenceData>(exception);
         }
         var envelopeRb = await BuildSnapshotEnvelopeAsync(
             canGetUnsafeState,
@@ -992,7 +988,14 @@ public class GeneralMultiProjectionActor
             if (canGetUnsafeState)
             {
                 // Promote buffered events before reading unsafe state
-                try { dualAccessor.PromoteBufferedEvents(safeWindowThreshold, _domain); } catch { }
+                try
+                {
+                    dualAccessor.PromoteBufferedEvents(safeWindowThreshold, _domain);
+                }
+                catch (Exception exception)
+                {
+                    return Task.FromResult(ResultBox.Error<MultiProjectionState>(exception));
+                }
 
                 statePayload = (IMultiProjectionPayload)dualAccessor.GetUnsafeProjectorPayload();
                 lastSortableId = dualAccessor.UnsafeLastSortableUniqueId;
--- a/dcb/src/Sekiban.Dcb.Core/Actors/GeneralMultiProjectionActor.cs
+++ b/dcb/src/Sekiban.Dcb.Core/Actors/GeneralMultiProjectionActor.cs
@@ -211,8 +211,16 @@ public class GeneralMultiProjectionActor
         if (_singleStateAccessor is IDualStateAccessor dualAccessor)
         {
             // A deferred safe-history repair is a consumption boundary. Returning an old head after its repair failed
-            // would falsely advertise a stale payload/position as current, so propagate the failure to the caller.
-            dualAccessor.PromoteBufferedEvents(safeWindowThreshold, _domain);
+            // would falsely advertise a stale payload/position as current, so capture the actual replay-failing event
+            // before propagating the original failure to the caller.
+            try
+            {
+                dualAccessor.PromoteBufferedEvents(safeWindowThreshold, _domain);
+            }
+            catch (Exception exception)
+            {
+                CaptureDeferredRepairFaultAndRethrow(exception);
+            }
 
             var current = new ProjectionPosition(
                 dualAccessor.UnsafeVersion,
@@ -630,6 +638,9 @@ public class GeneralMultiProjectionActor
         }
         catch (Exception exception)
         {
+            // ForcePromoteBufferedEvents has already recorded deferred replay provenance through the actor's normal
+            // first-observed-fault path. Preserve this same exception instance for the snapshot ResultBox rather than
+            // wrapping or capturing it a second time.
             return ResultBox.Error<SnapshotPersistenceData>(exception);
         }
         var envelopeRb = await BuildSnapshotEnvelopeAsync(
@@ -794,15 +805,47 @@ public class GeneralMultiProjectionActor
     /// </summary>
     private void CaptureFaultAndRethrow(Event ev, Exception ex)
     {
+        var original = CaptureFault(ev.Id, ev.EventType, ev.SortableUniqueIdValue, ex);
+        ExceptionDispatchInfo.Capture(original).Throw();
+    }
+
+    /// <summary>
+    ///     Captures a deferred safe-history replay failure at a consumption boundary. The wrapper records the exact
+    ///     event whose fold failed on the original exception before it leaves <c>RebuildSafeState</c>; it is deliberately
+    ///     not the earlier arrival that happened to mark history dirty. This feeds the same descriptor/log discipline as
+    ///     the normal per-event path without adding a public or serialized contract.
+    /// </summary>
+    private void CaptureDeferredRepairFaultAndRethrow(Exception exception)
+    {
+        var original = CaptureDeferredRepairFault(exception);
+        ExceptionDispatchInfo.Capture(original).Throw();
+    }
+
+    private Exception CaptureDeferredRepairFault(Exception exception)
+    {
+        if (!DeferredSafeRepairFaultAttribution.TryGetReplayEvent(
+                exception,
+                out var eventId,
+                out var eventType,
+                out var position))
+        {
+            return exception;
+        }
+
+        return CaptureFault(eventId, eventType, position, exception);
+    }
+
+    private Exception CaptureFault(Guid eventId, string eventType, string position, Exception ex)
+    {
         var original = ex.InnerException ?? ex;
 
         if (_fault is null)
         {
             _fault = new ProjectionFaultDescriptor(
-                ev.Id,
-                ev.EventType,
+                eventId,
+                eventType,
                 _projectorName,
-                ev.SortableUniqueIdValue,
+                position,
                 original.Message,
                 DateTime.UtcNow.Ticks);
             // The original exception is available only at this boundary. Emit it exactly once, with its real stack;
@@ -817,33 +860,13 @@ public class GeneralMultiProjectionActor
         }
 
         _fault.Annotate(original);
-        ExceptionDispatchInfo.Capture(original).Throw();
+        return original;
     }
 
     /// <summary>The deserialize-boundary counterpart of <see cref="CaptureFaultAndRethrow" />, before an Event exists.</summary>
     private void CaptureSerializableFaultAndRethrow(SerializableEvent se, Exception ex)
     {
-        var original = ex.InnerException ?? ex;
-
-        if (_fault is null)
-        {
-            _fault = new ProjectionFaultDescriptor(
-                se.Id,
-                se.EventPayloadName,
-                _projectorName,
-                se.SortableUniqueIdValue,
-                original.Message,
-                DateTime.UtcNow.Ticks);
-            _logger.LogError(
-                ProjectionFaultFirstObserved,
-                original,
-                "Projection fault first observed: {ProjectorName}, EventId: {EventId}, Position: {Position}",
-                _fault.ProjectorName,
-                _fault.EventId,
-                _fault.Position);
-        }
-
-        _fault.Annotate(original);
+        var original = CaptureFault(se.Id, se.EventPayloadName, se.SortableUniqueIdValue, ex);
         ExceptionDispatchInfo.Capture(original).Throw();
     }
 
@@ -883,8 +906,17 @@ public class GeneralMultiProjectionActor
     {
         if (_singleStateAccessor is IDualStateAccessor dualAccessor)
         {
-            var threshold = GetSafeWindowThreshold();
-            dualAccessor.PromoteBufferedEvents(threshold, _domain);
+            try
+            {
+                var threshold = GetSafeWindowThreshold();
+                dualAccessor.PromoteBufferedEvents(threshold, _domain);
+            }
+            catch (Exception exception)
+            {
+                // Snapshot and grain persistence call this entry point before capturing a checkpoint. A deferred replay
+                // fault therefore has to establish the actor descriptor here, before those callers fail closed.
+                CaptureDeferredRepairFaultAndRethrow(exception);
+            }
             return;
         }
 
@@ -910,7 +942,14 @@ public class GeneralMultiProjectionActor
     {
         if (_singleStateAccessor is IDualStateAccessor dualAccessor)
         {
-            dualAccessor.CompactSafeHistory();
+            try
+            {
+                dualAccessor.CompactSafeHistory();
+            }
+            catch (Exception exception)
+            {
+                CaptureDeferredRepairFaultAndRethrow(exception);
+            }
         }
     }
 
@@ -935,12 +974,19 @@ public class GeneralMultiProjectionActor
     {
         if (_singleStateAccessor is IDualStateAccessor dualAccessor)
         {
-            var maxThreshold = SortableUniqueId.MaxValue;
-            dualAccessor.PromoteBufferedEvents(maxThreshold, _domain);
-            _logger.LogDebug(
-                "[SafePromotion] projector={ProjectorName} force-promote-all invoked threshold={Threshold}",
-                _projectorName,
-                maxThreshold.Value);
+            try
+            {
+                var maxThreshold = SortableUniqueId.MaxValue;
+                dualAccessor.PromoteBufferedEvents(maxThreshold, _domain);
+                _logger.LogDebug(
+                    "[SafePromotion] projector={ProjectorName} force-promote-all invoked threshold={Threshold}",
+                    _projectorName,
+                    maxThreshold.Value);
+            }
+            catch (Exception exception)
+            {
+                CaptureDeferredRepairFaultAndRethrow(exception);
+            }
             return;
         }
 
@@ -994,7 +1040,8 @@ public class GeneralMultiProjectionActor
                 }
                 catch (Exception exception)
                 {
-                    return Task.FromResult(ResultBox.Error<MultiProjectionState>(exception));
+                    return Task.FromResult(ResultBox.Error<MultiProjectionState>(
+                        CaptureDeferredRepairFault(exception)));
                 }
 
                 statePayload = (IMultiProjectionPayload)dualAccessor.GetUnsafeProjectorPayload();
@@ -1022,7 +1069,15 @@ public class GeneralMultiProjectionActor
             else
             {
                 // Promote buffered events and get safe state
-                dualAccessor.PromoteBufferedEvents(safeWindowThreshold, _domain);
+                try
+                {
+                    dualAccessor.PromoteBufferedEvents(safeWindowThreshold, _domain);
+                }
+                catch (Exception exception)
+                {
+                    return Task.FromResult(ResultBox.Error<MultiProjectionState>(
+                        CaptureDeferredRepairFault(exception)));
+                }
 
                 statePayload = (IMultiProjectionPayload)dualAccessor.GetSafeProjectorPayload();
                 var safeSortableId = dualAccessor.SafeLastSortableUniqueId;
--- a/dcb/src/Sekiban.Dcb.Core/MultiProjections/DualStateProjectionWrapper.cs
+++ b/dcb/src/Sekiban.Dcb.Core/MultiProjections/DualStateProjectionWrapper.cs
@@ -160,14 +160,20 @@ public class DualStateProjectionWrapper<T>
         if (isInSafeWindow)
         {
             // Still inside the safe window: hold it in the buffer. The served state is (safe baseline + ordered buffer).
-            // While either state is deferred, even an apparently in-order tail must stay retained until consumption; folding
-            // it now would make the result depend on arrival order again.
+            // Only a deferred SAFE-history repair suppresses the normal unsafe-window reconcile. An ordinary unsafe
+            // out-of-order arrival must still reconcile immediately so a fold failure is captured at this ProcessEvent
+            // boundary with the offending event's fault attribution.
             var inOrder = !_safeHistoryDirty && !_servedStateDirty && IsStrictlyAfterServedHead(evt);
             _bufferedEvents[evt.Id] = evt;
 
             if (inOrder)
             {
                 ProjectUnsafeInOrder(evt, safeWindowThreshold, domainTypes);
+            }
+            else if (!_safeHistoryDirty)
+            {
+                ReconcileServedState(safeWindowThreshold, domainTypes);
+                _servedStateDirty = false;
             }
             else
             {
@@ -187,7 +193,8 @@ public class DualStateProjectionWrapper<T>
             {
                 ProjectSafeInOrder(evt, domainTypes);
                 _safeVersion++;
-                _servedStateDirty = true;
+                ReconcileServedState(safeWindowThreshold, domainTypes);
+                _servedStateDirty = false;
             }
             else
             {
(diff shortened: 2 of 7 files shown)


tomohisa (commented):
# Semantic review — REQUEST-UPDATE

Reviewed the packet, mandatory `review-context.md`, implementation notes, issue #1154, and the full production diff at exact head `8aca07c284dbbe758d0f5ed336fb3b0b2008beb6`. Tests are green, but they are necessary rather than sufficient. One AC3 contract failure remains, so the semantic verdict is **REQUEST-UPDATE**.

## Blocking finding

1. **AC3 / packet `REPAIR FAILURE`: deferred repair does not preserve the precise projector-failing event, and it does not create the actor fault descriptor / first-observed fault log.** `RebuildSafeState` knows the event currently being folded (`DualStateProjectionWrapper.cs:646-659`), but it throws the projector exception without annotating that event. `ConsumeServedState` then calls `AttachDeferredRepairAttribution` (`:438-450`), which always reports `_dirtyCausingEventId` recorded by `MarkSafeHistoryDirty` (`:454-475`). Those can be different events: process `high`, then `low` to make history dirty, configure the projector to fail on `high` only during the deferred replay, and consume. The actual failing event is `high`, but the exception reports `low`. Moreover, the actor's only `ProjectionFaultDescriptor`/`ProjectionFaultFirstObserved` creation is `CaptureFaultAndRethrow` on the per-event apply path (`GeneralMultiProjectionActor.cs:716-820`); failures from `GetProjectionHeadStatusAsync`, `GetStateFromSingleAccessorAsync`, and `BuildSnapshotForPersistenceAsync` do not pass through it, so `CurrentFault` remains null and no precise first-observed fault record is produced.

   The current tests configure failure on `seed.Low`, which is also the event that made the history dirty (`DualStateProjectionWrapperDeferredRepairTests.cs:69-95, 124-155, 251-316`), so they cannot distinguish correct failing-event attribution from this implementation. Required killing evidence: make the dirty-causing event and the replay-failing event distinct; drive at least a real actor query and persistence boundary; assert the original exception is preserved, the failing event's id/position (not the dirty trigger's) reaches exception data plus `CurrentFault`/first-observed log, prior payload and all safe/unsafe metadata remain unchanged, dirty/history remain retryable, and compaction/switch/persist do not run. This is an implementation finding under AC3 and the packet's `REPAIR FAILURE` clause.

## AC1–AC8 evidence matrix

| AC | Production code path | Killing discriminant / executed evidence | Result |
|---|---|---|---|
| AC1 Losslessness + ties | `_allSafeEvents` is a `SortedDictionary<SafeEventOrderKey, Event>`; `SafeEventOrderKey.CompareTo` orders sortable id ordinal then event id, and `RebuildSafeState` folds that retained order (`DualStateProjectionWrapper.cs:18-23, 582-590, 625-670`). Duplicate event ids are rejected by `_processedEventIds` before insertion. | `DeferredRepair_IsLosslessForSortableUniqueIdTies` compares immediate and reverse-arrival deferred results including payload, last id, position and version. Boundary/buffer/duplicate behavior is also exercised by the existing wrapper suites. Removing the event-id tie-breaker or arrival-independence changes `ABC` and fails. | **PASS** |
| AC2 Suppression and order | Dirty safe arrivals return from `ProcessEvent` without `ReconcileServedState` (`:210-231`). `ConsumeServedState` rebuilds, publishes safe payload/version/last id/position, reconciles unsafe state, then returns to the consumer (`:386-451`). Clean safe and unsafe in-order branches retain immediate projection/reconciliation (`:160-225`). | `DirtyProcessEvent_SuppressesReconcileUntilTheFirstRealConsumeThenReconcilesOnce`: after high/buffered/low/middle, fold/tag counts stay 2 and published safe/unsafe remain `C`/`CU`; first getter produces `ABCU` with exactly 3 rebuild folds + 1 reconcile fold; repeat getter adds zero. This kills event-end reconcile, stale-payload incremental folding, missing reconcile, duplicate reconcile, and consume-before-reconcile. | **PASS** |
| AC3 Seven boundaries + atomic failure | Shared private seam is `ConsumeServedState`. (1) safe getter -> `ProcessBufferedEvents` -> seam; unsafe getter -> seam. (2) payload getters -> remembered-context seam; actor query/snapshot/persist promote before capture/getter. (3) dirty `ProcessEvent` suppresses direct reconciliation; dirty reconciliation occurs only inside the seam. (4) `ProcessBufferedEvents` calls the seam unconditionally, including zero promotions. (5) grain `PersistStateAsync` force-promotes before `CapturePersistCheckpointAsync`; actor snapshot promotes before envelope/position capture. (6) `CompactSafeHistory` consumes before clear/switch. (7) readiness includes both dirty flags, while the actor promotes before evaluating it. Rollback restores eight safe/unsafe payload/metadata fields. | Six wrapper production entries and actor head-status are driven from dirty state for success/failure; snapshot and grain persistence tests prove fail-before-snapshot/write/compaction. These kill omitted forwarding, zero-event gating, clear-before-repair, and partial publication. **But the distinct failing-event counterexample above survives, and no actor fault descriptor/log assertion exists.** | **FAIL** |
| AC4 Deterministic complexity | Dirty insertion is sorted-map insertion plus flag only; rebuild enumerates the already ordered map once (`:188-231, 625-670`). General complexity is `O(N log N + N × consumption epochs)`. | `DeferredRepair_FoldAndTagResolutionCountsAreLinearAndEachDirtyEpochRepairsOnce`: N=64 / 2N=128 gives 65 / 129 folds after one consume, two fold starts; repeat consume adds zero; next dirty epoch adds one fold start. The old `1+2+...+N` arrival rebuild fails these exact counts. | **PASS** |
| AC5 Wall-clock auxiliary | No production behavior relies on timing and no timing assertion is used as the gate. | AC4's deterministic counters, not elapsed time, are the primary evidence. | **PASS** |
| AC6 G37/G38 interaction | Grain repair is fail-closed before checkpoint capture/write (`MultiProjectionGrain.cs:1450-1464`); successful durable paths still call host compaction, whose wrapper path repairs before clearing and switching (`DualStateProjectionWrapper.cs:273-286`). | Existing G38 real-wrapper convergence test was updated to require the actual `SortedDictionary<,>` and still observes retained count >0 after durable checkpoint and 0 after compaction. Compaction suite plus `PersistStateAsync_WhenDeferredRepairFails...` distinguish success from fail-before-write/compact. | **PASS** |
| AC7 Per-arrival allocation/tag work | No tags are resolved in the dirty insertion branches; `ResolveTags` is called only for actual folds (`DualStateProjectionWrapper.cs:479-519, 625-670`). | The AC2/AC4 counting projector pins tag resolutions to fold calls: only 1 during adversarial insertion, N during repair, none on repeat consume. Reintroducing the old per-arrival full loop increases both deterministic counters quadratically. | **PASS** |
| AC8 Public API unchanged | Diff adds only private fields/types/helpers and changes no public/interface/options/serialized-state declaration. `IDualStateAccessor` remains unchanged. | `PublicDualStateSurface_IsFrozenWithoutDeferredRepairMembers` pins the wrapper's declared public method names and the accessor method inventory; full diff inspection found no public signature addition/removal/modification. | **PASS** |

## Seven-boundary call-graph assessment

1. `GetSafeProjection` -> `ProcessBufferedEvents` -> `ConsumeServedState`; `GetUnsafeProjection` -> `ConsumeServedState` directly.
2. Both explicit payload getters -> `ConsumeUsingRememberedContext` -> seam. `GeneralMultiProjectionActor.GetStateFromSingleAccessorAsync`, `BuildSnapshotForPersistenceAsync`, and grain persist paths force promotion before reading/capturing.
3. The only dirty repair/reconcile path is the seam. Direct `ReconcileServedState` calls remain only for clean in-order/ordinary unsafe behavior; dirty safe paths return without them.
4. `PromoteBufferedEvents` -> `ProcessBufferedEvents`; line 365 consumes outside `eventsToProcess.Count > 0`, proving zero-event reachability.
5. `MultiProjectionGrain.PersistStateAsync`: force promote -> checkpoint capture -> short-circuit/write. Actor snapshot: force promote -> envelope -> safe position. Repair failure exits before these stages.
6. `CompactSafeHistory`: remembered-context consume -> rebuild guard -> clear retained set -> rebuild duplicate set -> incremental switch. Throwing repair prevents every later step.
7. `IsServedIdenticalToSafe` is false for rebuild, safe-dirty, served-dirty, or buffered states. The actor's unsafe query path promotes/repairs before reading payload+metadata and then evaluates readiness.

## Verification and scope

- Focused wrapper/compaction: `dotnet test ...Sekiban.Dcb.WithResult.Tests.csproj -f net9.0|net10.0 --filter "FullyQualifiedName~DualStateProjectionWrapperDeferredRepairTests|FullyQualifiedName~DualStateProjectionWrapperCompactionTests"` — **24/24 PASS** on each TFM.
- Focused Orleans persistence/G38: `dotnet test ...Sekiban.Dcb.Orleans.Tests.csproj -f net9.0|net10.0 --filter "FullyQualifiedName~MultiProjectionGrainPersistPolicyTests|FullyQualifiedName~MultiProjectionGrainWatermarkRetirementTests"` — **31/31 PASS** on each TFM.
- Exact-head CI: `dcbTestsNet9`, `dcbTestsNet10`, `SonarCloud`, and `SonarCloud Code Analysis` are terminal **SUCCESS** at the reviewed SHA.
- Diff hygiene: full base/head dcb diff check reported no whitespace errors (only pre-existing CRLF normalization warnings).
- Base is `main`; head remained exact; PR is non-draft and mergeable. Body contains `Closes #1154` followed by `Fixes #1146`.
- Scope remains in `dcb/src` and `dcb/tests`; no snapshot-restore feature, public option, release/tag, or partial-refold implementation was added. The changes are consistent with `intents/sekiban/design/ooo-refold-quadratic.md`, but the precise fault-attribution clause is not yet met.

**Verdict: REQUEST-UPDATE at `8aca07c284dbbe758d0f5ed336fb3b0b2008beb6`.**

tomohisa (commented):
Semantic re-review at exact head `063a70871a7b2eed9b7c846e9b094e6c31b51ec4`

**Verdict: APPROVE-READY.** The prior AC3 blocker is resolved. I reviewed the repair diff from `8aca07c284dbbe758d0f5ed336fb3b0b2008beb6`, then retraced the full PR against the packet and mandatory review context. Tests were treated as killing evidence, not as a substitute for the production-path review.

## Prior AC3 finding

The production replay loops now annotate the *actual event being replayed* immediately when projection fails: `RebuildSafeState` and `ReconcileServedState` attach replay event id/type/position before rethrowing the original exception. The low event that first made the wrapper dirty is retained separately under the dirty-attribution keys. `GeneralMultiProjectionActor` extracts the replay annotation at the real query, head-status, promotion, compaction, and snapshot/persistence boundaries and funnels it through the existing `_fault is null` gate.

The new distinct-event tests use low=A as the dirty cause and high=B as the replay failure. They prove `Assert.Same(original, failure)`, B's id/type/position in exception data, `CurrentFault`, and `ProjectionFaultDescriptor`, A only in dirty-attribution data, and exactly one `ProjectionFaultFirstObserved` log. The real `GetStateAsync` actor query and `BuildSnapshotForPersistenceAsync` paths are both exercised. A repeated boundary does not recapture or emit another first-observed log.

Failure atomicity is also direct: payload plus safe/unsafe version, last-id, and position stay at the prior values; dirty flags and retained history remain; the replay stops at B; reconcile publication, compaction, incremental switching, snapshot/checkpoint creation, and provider write do not happen. After removing the injected B fault and clearing the actor fault for rebuild, retry consumes the retained history and converges to the exact expected state without event loss. This kills the former low/high conflation and the former actor-boundary provenance gap.

## AC matrix

| AC | Production path | Killing discriminant and executed evidence | Result |
|---|---|---|---|
| AC1 | `ProcessEvent` marks out-of-order non-incremental state dirty and suppresses event-end `ReconcileServedState`; clean in-order handling keeps the existing reconcile path. | `DirtyProcessEvent_SuppressesReconcileUntilTheFirstRealConsumeThenReconcilesOnce`: multiple OOO inputs publish neither served nor unsafe state and do not advance reconcile count; the first consume repairs and reconciles once. | PASS |
| AC2 / 2b | All dirty consumption goes through the one private `ConsumeServedState` seam: repair -> atomic payload/metadata publication -> reconcile -> consume. Retained events are ordered by `SortableUniqueId` then ordinal `EventId`. | Production-entry matrix, suppression test, linear fold/tag-resolution count test, and sortable-id tie test kill eager re-folding, repeated-consume work, and nondeterministic ties. Counts scale with retained input plus consumption epochs rather than arrival-time quadratic re-folding. | PASS |
| AC3 | Wrapper replay loops annotate the exact failing replay event; actor capture turns that into the first durable/current fault at real query and persistence boundaries while preserving the original exception. Rollback restores all published fields and keeps history dirty/retryable. | Distinct low/high wrapper matrix, actor head-status, real actor state-read, compaction, and real actor snapshot-persistence tests prove high provenance, low dirty provenance, original identity, one first-observed log, no stale result/write/compaction/switch, and exact retry convergence. `MultiProjectionGrainPersistPolicyTests.PersistStateAsync_WhenDeferredRepairFails_DoesNotWriteOrCompactAndReturnsTheOriginalFailure` proves zero provider writes. | PASS (prior blocker resolved) |
| AC4 | Publication inside the seam assigns payload, safe/unsafe version, last id, and position as one commit after successful replay; in-order events arriving while dirty only extend sorted retained history. | Entry-point and failure matrices snapshot every published field before failure; success paths assert exact final payload/metadata. The tie-order test kills arrival-order or partial-metadata publication. | PASS |
| AC5 | `GetSafeProjection`, direct `GetUnsafeProjection`, explicit payload access via remembered context, `PromoteBufferedEvents` (including zero-event), snapshot/persist promotion, `CompactSafeHistory`, and actor head/readiness status all reach the shared seam. | `DirtyHistory_IsConsumedExactlyOnceThroughEachWrapperProductionEntryPoint` plus the actor head-status tests exercise the full seven-boundary inventory. This is backed by tracing the public actor/grain call graph, not helper assignability alone. | PASS |
| AC6 | `CompactSafeHistory` consumes before clearing history/enabling incremental mode; exceptions leave dirty/history intact and prevent the later steps. | `RepairFailure_PublishesNothingRetainsCauseAndBlocksCompactionUntilRetry` and the compaction suite observe retained real wrapper history after failure, then successful retry and compaction. | PASS |
| AC7 | Grain persistence calls `ForcePromoteBufferedEvents` before checkpoint capture, short-circuit, snapshot serialization, and storage write. Actor snapshot persistence also promotes before envelope/position capture. | Actor snapshot-persistence failure test proves no snapshot/checkpoint progress; grain persist-policy test proves promotion once, snapshot write 0, compact 0, provider write 0. Existing G37 trigger behavior is not used to manufacture the result. | PASS |
| AC8 | Changes remain private/internal to `DualStateProjectionWrapper`, `GeneralMultiProjectionActor`, and focused tests; no public API, interface, option, or serialized-state member was added. G38 durable-checkpoint-before-compaction ordering remains intact. | Frozen public-surface test passes; focused G38 watermark/real-wrapper compaction suite passes; full diff is limited to the expected three production/test areas plus the prior PR scope. PR body retains `Closes #1154` then `Fixes #1146`. | PASS |

## Seven production consumption boundaries

1. Safe/unsafe projection access: safe flows through buffered processing; direct unsafe access calls the seam.
2. Explicit payload getters: remembered-context consumption; actor queries and snapshots promote before reading.
3. Event-end reconcile: suppressed while dirty; clean in-order behavior remains unchanged.
4. Promotion: the seam is outside the non-empty-buffer conditional, covering zero-event promotion.
5. Checkpoint/snapshot capture: promotion precedes checkpoint capture, serialization, and write.
6. Compaction: consume succeeds before retained-history clear and incremental switch.
7. Readiness/head evaluation: actor promotion precedes `IsServedIdenticalToSafe` evaluation.

## Verification

- `dotnet test dcb/tests/Sekiban.Dcb.WithResult.Tests/Sekiban.Dcb.WithResult.Tests.csproj -f net9.0 --filter "FullyQualifiedName~DualStateProjectionWrapperDeferredRepairTests|FullyQualifiedName~DualStateProjectionWrapperCompactionTests"` — 25/25 passed.
- Same command with `-f net10.0` — 25/25 passed.
- `dotnet test dcb/tests/Sekiban.Dcb.Orleans.Tests/Sekiban.Dcb.Orleans.Tests.csproj -f net9.0 --filter "FullyQualifiedName~MultiProjectionGrainPersistPolicyTests|FullyQualifiedName~MultiProjectionGrainWatermarkRetirementTests"` — 31/31 passed.
- Same command with `-f net10.0` — 31/31 passed.
- Full PR patch applied to exact base `b057951cfa3624bdecba0210a00bc37cab19aa2e` with `git apply --check --whitespace=error-all`: passed.
- Exact-head terminal checks: `dcbTestsNet9`, `dcbTestsNet10`, `SonarCloud`, and `SonarCloud Code Analysis` are all completed/success. Run `32609031150` attempt 1 had the reported unrelated net10 `MaterializedViewUnitTests.CatchUpWorker_UsesPollDelayWhenNothingProcessed` cancellation; attempt 2 completed both net9/net10 successfully at this same SHA.
- Base is `main`; no unrelated public surface or scope expansion found.

No remaining semantic finding against AC1-AC8.

PR #1157, opened 2026-08-23
SEK-G42 dcb: stream offloaded snapshot restore

### LINKED ISSUE #1145: Snapshot restore holds the serialized payload and the deserialised graph at once, so a large projection can OOM at the last step even with UseStreamingSnapshotIO
## Summary

Snapshot restore holds the fully materialised payload byte array and the deserialised object graph alive at the same time. `UseStreamingSnapshotIO` — which defaults to `true` — removes the buffering on the way in, but not this final overlap, so a large projection can still OOM at the last step of restore on a host whose heap ceiling is close to the steady-state size.

This is reported as a real failure, not a theoretical one: it is the first event in an incident that ended with a projection unrecoverable for two days.

## Where the overlap is

`dcb/src/Sekiban.Dcb.Core/Actors/GeneralMultiProjectionActor.cs:266-276`

```csharp
private void RestorePayload(SerializableMultiProjectionState state)
{
    var payloadBytes = state.GetPayloadBytes();
    var projTypeRb = _types.GetProjectorType(state.ProjectorName);
    if (!projTypeRb.IsSuccess) throw projTypeRb.GetException();

    var safeThreshold = GetSafeWindowThreshold();
    var deserializeResult = _types.Deserialize(state.ProjectorName, _domain, safeThreshold.Value, payloadBytes);
    if (!deserializeResult.IsSuccess) throw deserializeResult.GetException();

    var loadedPayload = deserializeResult.GetValue();
    ...
```

`GetPayloadBytes()` produces the whole decompressed payload as a single array, and `Deserialize` then builds the complete object graph from it. Both are live until `RestorePayload` returns. On the reported host the serialized payload was 173,479,838 B, and the in-memory object graph for a projection of that shape is substantially larger than its serialized form (a separate local measurement on the same projector put the serialized size at roughly 478 B per entity while process RSS grew by about 9 KB per entity — a ~19x expansion; treat the exact ratio as indicative, not as a constant).

## Streaming is already on and does not cover this

`dcb/src/Sekiban.Dcb.Core/Actors/GeneralMultiProjectionActorOptions.cs:121`

```csharp
public bool UseStreamingSnapshotIO { get; set; } = true;
```

The reported host does not override it. The failing stack confirms the streaming entry point was in use and still ended in the materialising step:

```
System.OutOfMemoryException
   at Sekiban.Dcb.Actors.GeneralMultiProjectionActor.RestorePayload(SerializableMultiProjectionState state)
   at Sekiban.Dcb.Actors.GeneralMultiProjectionActor.SetCurrentState(SerializableMultiProjectionState state)
   at Sekiban.Dcb.Actors.GeneralMultiProjectionActor.SetSnapshotAsync(SerializableMultiProjectionStateEnvelope envelope, CancellationToken cancellationToken)
   at Sekiban.Dcb.Runtime.Native.NativeProjectionSnapshotHandler.RestoreSnapshotFromStreamAsync(Stream source, CancellationToken cancellationToken)
```

`RestoreSnapshotFromStreamAsync` takes a `Stream`, but the value it hands down is a `SerializableMultiProjectionState`, so the stream is consumed into a materialised state before the deserializer sees it. The setting reduces transport buffering; it does not change the shape of this step.

## Measured memory on the reported host

| Phase | Process `MemoryWorkingSet` |
| --- | --- |
| Steady state after recovery (this projector active, others not) | 1,586–1,719 MB |
| Peak while rebuilding from `beginning` | 4,500–5,306 MB |
| Steady state with all projectors active | 3,764–4,561 MB |
| GC heap ceiling on the host tier (B3, 7 GiB) | ~5.2 GB |

The restore OOM occurred at 4.3–4.8 GB. The projection had grown because the projector version immediately before the incident added roughly 70 retained columns per entity, which raised both the serialized size and the restore peak.

## Why it is worth addressing rather than sizing around

Restoring costs meaningfully more than holding. On this host the steady state fits in well under half the ceiling while the restore peak reaches it. That means a host can run a projection for weeks and then fail to start with it — the failure appears at activation, which is also when the fallback is a full rebuild from `beginning`, the most expensive path available.

In the reported incident that fallback then interacted with two other issues (#1139 and #1144) and the projection could not be recovered by any supported means for two days. The restore OOM was the first domino.

## Suggested direction

Feed the decompression stream directly to the deserializer instead of materialising the payload first, so the byte array never exists as a whole. Where the serializer supports it, deserializing from a `Stream` / `ReadOnlySequence` would remove one full copy of the serialized size from the peak.

If a full-array API is required by the current serializer surface, an intermediate improvement would be to release the array as early as possible and to size the restore path's allocations from the recorded `OriginalSizeBytes` so a restore that cannot fit fails with a clear, actionable error before it exhausts the heap — a diagnosable refusal is more useful than an `OutOfMemoryException` whose stack is then discarded (see the fault-reporting issue filed alongside this one).

## Verification status

Verified by reading `origin/main` (`cca3c8d64c5a5a013b8ce4e5197c160b26423864`): `RestorePayload`'s materialise-then-deserialize sequence, the default value of `UseStreamingSnapshotIO`, and that the streaming restore entry point terminates in `SetCurrentState(SerializableMultiProjectionState)`.

The memory figures, the serialized payload size and the failing stack are from an operator report on a deployed host running 10.17.0; I have not confirmed those binaries are byte-identical to this `origin/main`, and I have not reproduced the restore OOM locally. The per-entity expansion ratio is from a separate local measurement on a different data set and is offered only as an order of magnitude.

### LINKED ISSUE #1156: SEK-G42 dcb: stream snapshot restore end to end
# SEK-G42 dcb: stream snapshot restore end to end

Fixes #1145.

## Goal

Stop holding the whole decompressed snapshot payload alongside the object
graph it is being turned into, so a large projection stops OOMing at the last
step of restore.

## Why This Slice Exists Now

Reported as a real failure: a 173,479,838 B payload, with a separate
measurement putting the serialized size at roughly 478 B per entity while
process RSS grew about 9 KB per entity. It was the first event in an incident
that ended with a projection unrecoverable for two days.

## Current Observed State

`RestorePayload` materializes the whole decompressed payload as a single array
and then builds the complete graph; both are live until it returns.
`UseStreamingSnapshotIO` (default true) removes the inbound buffering but not
this final overlap.

Upstream matters too: the production path resolves an offloaded payload into a
full `byte[]` before `RestorePayload` ever runs, so a registry overload alone
cannot remove the array.

- **Limited, verifiable guarantee**: under a stream-capable restore, Sekiban
  does not materialize an additional contiguous `byte[]`/`string` proportional
  to the full uncompressed payload length. This buys back roughly the
  serialized size; the graph expansion remains and dominates. **"No OOM" is
  not promised.** Inline Base64 envelopes are outside the guarantee and become
  an observable fallback; offloaded payloads are the guaranteed case.
- **End-to-end seam**: the resolver hands the actor an opened non-seekable
  stream instead of inlining; the actor deserializes from it and publishes
  payload and metadata only on success; the buffered path runs only when the
  capability is absent; failures fail the activation before serving.
- **Optional capability interface** leaving `ICoreMultiProjectorTypes`
  untouched (a default interface method requires a 10.18-compiled binary
  fixture), detected **per projector** across AOT, reflection JSON and custom
  serialization.
- **Absent falls back; a genuine failure does not** (a buffered retry doubles
  memory and hides corruption).
- **Stream contract**: async + `CancellationToken`, non-seekable, read from the
  current position, caller-owned (callee never disposes).

## Target Repo / Path / Part

- Target part: end-to-end streaming restore for offloaded snapshots — the optional stream-deserialization capability, the state-store-to-actor seam, capability-absent fallback separated from genuine failure, and fail-before-serving.

## In Scope

Everything in the Accepted Baseline and the eleven acceptance criteria below,
with EN/JA docs.

## Out Of Scope

- Shrinking the object graph itself, or changing the compression format.
- The symmetric change on the save side (restore leads; save is a follow-up).
- Any guarantee for inline Base64 envelopes beyond observable fallback.

The child PR must carry an opened stream from the state store to the actor for
offloaded payloads, add an optional per-projector stream-deserialization
capability without touching the existing registry interface, separate
capability absence (fallback) from genuine failure (propagate), fail the
activation before serving on restore errors, and prove it with a structural
production-path test rather than a read-size test — public API additive only.

## Acceptance Criteria

- Primary structural proof: production entry point succeeds with a non-seekable, Length/Seek-forbidding, read-capped stream without passing through `ReadPayloadBytesAsync`/`GetPayloadBytes()`; seam counters or structural inspection show no whole-buffer aggregation API is reached; process-isolated allocation/peak ceilings as supporting evidence; an intentionally buffered registry reliably fails it.
- Real path traversed: JSON+gzip goes through `GZipStream` and `JsonSerializer.DeserializeAsync` (or the AOT `JsonTypeInfo` equivalent).
- Equivalence matrix across reflection/AOT/custom serialization, gzip/raw legacy, v9/v10, offloaded/inline, with an explicit guaranteed-vs-fallback table.
- Binary compatibility: a 10.18-compiled external registry assembly loads and is invoked on 10.19.
- Absent vs failure counterexamples, including the one-per-attempt fallback log without payload content.
- Stream contract: cancellation, non-seekable, caller-owned, partial reads, truncated/corrupt payloads.
- Fail-before-serving: no query served before restore completes; nothing partial or stale published on failure.
- Restore caller inventory table; the entry point from the production OOM trace exercised by a killing test.
- CI size split: 16–32 MB controlled fixture in normal CI; 128–200 MB smoke nightly/manual, evaluating absence of full-payload materialization rather than "no OOM".
- Additive public API with reflection pins; existing registry interface untouched.
- EN/JA docs stating the limited guarantee, the offloaded-vs-inline boundary, and fallback conditions.

## Verification

## Related Links

- Host design record: intents/sekiban/design/restore-memory-overlap.md
- Consultations: reports/sek-g42-design-consult.md, reports/sek-g42-design-final.md

## Knowledge Maintenance

- Intent placement: host design doc restore-memory-overlap.md.
- Docs update: EN/JA restore guarantee and fallback guidance — in scope.
- Closeout writeback expected: yes — caller inventory outcome, per-projector detection needs, measured ceilings, whether the inline exclusion held, async reshape scope.

## Base Branch Policy

Policy: `direct-main`

tomohisa:
Closes #1156
Fixes #1145

## Summary
- carry offloaded snapshot streams from resolver through actor to optional per-projector deserializers
- retain a one-time observable fallback for capability-absent external registries and propagate real stream failures
- add production-path, compatibility, structural-memory, and net9/net10 coverage plus EN/JA guidance

## Validation
- focused WithResult streaming restore tests: 18 passed on net9.0 and net10.0
- focused production NativeProjectionActorHost tests: 2 passed on net9.0 and net10.0
- release build includes the separately compiled 10.18 registry fixture

### DIFF
--- a/.github/workflows/dcb_streaming_restore_memory.yml
+++ b/.github/workflows/dcb_streaming_restore_memory.yml
@@ -11,6 +11,7 @@ jobs:
     timeout-minutes: 35
     env:
       SEKIBAN_STREAM_RESTORE_SMOKE: '1'
+      SEKIBAN_STREAM_RESTORE_CONTROLLED_CEILING: '1'
 
     steps:
       - uses: actions/checkout@v4
@@ -30,6 +31,12 @@ jobs:
 
       - name: Build focused memory fixture
         run: dotnet build dcb/tests/Sekiban.Dcb.WithResult.Tests/Sekiban.Dcb.WithResult.Tests.csproj -c Release -f net10.0 --no-restore -p:GeneratePackageOnBuild=false --verbosity minimal
+
+      - name: Run controlled 16-32 MiB small-graph fixture in an isolated process
+        run: |
+          # This separate test host owns the allocation ceiling for the controlled fixture. The normal PR suite still
+          # runs the same fixture's supported-path and intentionally-buffered counter-proof; this is supporting evidence.
+          timeout 10m dotnet test dcb/tests/Sekiban.Dcb.WithResult.Tests/Sekiban.Dcb.WithResult.Tests.csproj -c Release --no-build -f net10.0 --filter "Category=StreamingRestoreControlledMemory" --logger "console;verbosity=normal"
 
       - name: Run 128-200 MiB restore smoke under a process ceiling
         run: |
--- a/.github/workflows/dcb_streaming_restore_memory.yml
+++ b/.github/workflows/dcb_streaming_restore_memory.yml
@@ -1,0 +1,39 @@
+name: DCB Streaming Restore Memory Smoke
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '17 3 * * 0'
+
+jobs:
+  streamingRestoreMemory:
+    runs-on: ubuntu-latest
+    timeout-minutes: 35
+    env:
+      SEKIBAN_STREAM_RESTORE_SMOKE: '1'
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup .NET 9
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 9.0.x
+
+      - name: Setup .NET 10
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 10.0.x
+
+      - name: Restore focused dependencies
+        run: dotnet restore dcb/tests/Sekiban.Dcb.WithResult.Tests/Sekiban.Dcb.WithResult.Tests.csproj
+
+      - name: Build focused memory fixture
+        run: dotnet build dcb/tests/Sekiban.Dcb.WithResult.Tests/Sekiban.Dcb.WithResult.Tests.csproj -c Release -f net10.0 --no-restore -p:GeneratePackageOnBuild=false --verbosity minimal
+
+      - name: Run 128-200 MiB restore smoke under a process ceiling
+        run: |
+          # The selected test process is isolated from normal PR CI. 6 GiB virtual-address limit and 30 minute wall
+          # timeout are supporting safety ceilings, while the test records peak RSS and selected-path telemetry.
+          ulimit -Sv 6291456
+          timeout 30m dotnet test dcb/tests/Sekiban.Dcb.WithResult.Tests/Sekiban.Dcb.WithResult.Tests.csproj -c Release --no-build -f net10.0 --filter "Category=StreamingRestoreManualMemory" --logger "console;verbosity=normal"
--- a/dcb/src/Sekiban.Dcb.Core.Model/Domains/AotMultiProjectorTypes.cs
+++ b/dcb/src/Sekiban.Dcb.Core.Model/Domains/AotMultiProjectorTypes.cs
@@ -17,7 +17,6 @@ public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes, IStreamin
 {
     private static volatile bool _debugBypassProject;
     private readonly Dictionary<string, ProjectorRegistration> _projectors = new();
-    private readonly StreamingMultiProjectorTypesSupport _streamingRestore = new();
     private static readonly bool DebugBypassProject =
         Environment.GetEnvironmentVariable("SEKIBAN_WASM_DEBUG_BYPASS_MULTI_PROJECT") == "1";
     private static bool _debugBypassProjectSwitch;
@@ -79,7 +78,7 @@ public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes, IStreamin
                 var jsonBytes = GzipCompression.Decompress(data);
                 return JsonSerializer.Deserialize(jsonBytes, typeInfo)!;
             });
-        _streamingRestore.RegisterJsonTypeInfo(name, typeInfo);
+        StreamingMultiProjectorTypesSupport.RegisterJsonTypeInfo(this, name, typeInfo);
     }
 
     /// <inheritdoc />
@@ -179,28 +178,9 @@ public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes, IStreamin
         return ResultBox.Error<IMultiProjectionPayload>(new Exception($"Projector not found: {projectorName}"));
     }
 
-    public bool SupportsStreamDeserialization(string projectorName) => _streamingRestore.Supports(projectorName);
+    public bool SupportsStreamDeserialization(string projectorName) => StreamingMultiProjectorTypesSupport.Supports(this, projectorName);
 
-    public Task<ResultBox<IMultiProjectionPayload>> DeserializeFromStreamAsync(
-        string projectorName,
-        DcbDomainTypes domainTypes,
-        string safeWindowThreshold,
-        Stream source,
-        CancellationToken cancellationToken = default)
-    {
-        if (!_projectors.ContainsKey(projectorName))
-        {
-            return Task.FromResult(ResultBox.Error<IMultiProjectionPayload>(
-                new Exception($"Projector not found: {projectorName}")));
-        }
-
-        return _streamingRestore.DeserializeAsync(
-            projectorName,
-            domainTypes,
-            safeWindowThreshold,
-            source,
-            cancellationToken);
-    }
+    public Task<ResultBox<IMultiProjectionPayload>> DeserializeFromStreamAsync(string projectorName, DcbDomainTypes domainTypes, string safeWindowThreshold, Stream source, CancellationToken cancellationToken = default) => StreamingMultiProjectorTypesSupport.DeserializeWithProjectorNotFoundAsync(this, projectorName, domainTypes, safeWindowThreshold, source, cancellationToken);
 
     public ResultBox<IMultiProjectionPayload> DeserializeJson(
         string projectorName,
--- a/dcb/src/Sekiban.Dcb.Core.Model/Domains/AotMultiProjectorTypes.cs
+++ b/dcb/src/Sekiban.Dcb.Core.Model/Domains/AotMultiProjectorTypes.cs
@@ -75,7 +75,9 @@ public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes, IStreamin
             },
             Deserialize: (domainTypes, safeWindowThreshold, data) =>
             {
-                var jsonBytes = GzipCompression.Decompress(data);
+                // Legacy raw UTF-8 snapshots and gzip snapshots must remain equivalent to the stream helper, which
+                // chooses gzip only from its two-byte marker.
+                var jsonBytes = GzipCompression.DecompressIfGzip(data);
                 return JsonSerializer.Deserialize(jsonBytes, typeInfo)!;
             });
         StreamingMultiProjectorTypesSupport.RegisterJsonTypeInfo(this, name, typeInfo);
--- a/dcb/src/Sekiban.Dcb.Core.Model/Domains/AotMultiProjectorTypes.cs
+++ b/dcb/src/Sekiban.Dcb.Core.Model/Domains/AotMultiProjectorTypes.cs
@@ -17,6 +17,7 @@ public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes, IStreamin
 {
     private static volatile bool _debugBypassProject;
     private readonly Dictionary<string, ProjectorRegistration> _projectors = new();
+    private readonly StreamingMultiProjectorTypesSupport _streamingRestore = new();
     private static readonly bool DebugBypassProject =
         Environment.GetEnvironmentVariable("SEKIBAN_WASM_DEBUG_BYPASS_MULTI_PROJECT") == "1";
     private static bool _debugBypassProjectSwitch;
@@ -27,8 +28,7 @@ public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes, IStreamin
         string Version,
         Func<DcbDomainTypes, string, IMultiProjectionPayload, SerializationResult> Serialize,
         Func<Stream, DcbDomainTypes, string, IMultiProjectionPayload, SerializationSizeInfo> SerializeToStream,
-        Func<DcbDomainTypes, string, byte[], IMultiProjectionPayload> Deserialize,
-        Func<Stream, DcbDomainTypes, string, CancellationToken, Task<IMultiProjectionPayload>> DeserializeFromStream);
+        Func<DcbDomainTypes, string, byte[], IMultiProjectionPayload> Deserialize);
 
     /// <summary>
     ///     Register a multi-projector with AOT-compatible JsonTypeInfo.
@@ -78,12 +78,8 @@ public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes, IStreamin
             {
                 var jsonBytes = GzipCompression.Decompress(data);
                 return JsonSerializer.Deserialize(jsonBytes, typeInfo)!;
-            },
-            DeserializeFromStream: async (source, _, _, cancellationToken) =>
-                await StreamSnapshotPayloadDeserializer.DeserializeJsonAsync(source, typeInfo, cancellationToken)
-                    .ConfigureAwait(false)
-                ?? throw new InvalidOperationException($"Failed to deserialize {typeof(TProjector).Name}")
-        );
+            });
+        _streamingRestore.RegisterJsonTypeInfo(name, typeInfo);
     }
 
     /// <inheritdoc />
@@ -183,41 +179,27 @@ public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes, IStreamin
         return ResultBox.Error<IMultiProjectionPayload>(new Exception($"Projector not found: {projectorName}"));
     }
 
-    public bool SupportsStreamDeserialization(string projectorName) => _projectors.ContainsKey(projectorName);
+    public bool SupportsStreamDeserialization(string projectorName) => _streamingRestore.Supports(projectorName);
 
-    public async Task<ResultBox<IMultiProjectionPayload>> DeserializeFromStreamAsync(
+    public Task<ResultBox<IMultiProjectionPayload>> DeserializeFromStreamAsync(
         string projectorName,
         DcbDomainTypes domainTypes,
         string safeWindowThreshold,
         Stream source,
         CancellationToken cancellationToken = default)
     {
-        if (string.IsNullOrWhiteSpace(safeWindowThreshold))
+        if (!_projectors.ContainsKey(projectorName))
         {
-            return ResultBox.Error<IMultiProjectionPayload>(
-                new ArgumentException("safeWindowThreshold must be supplied"));
+            return Task.FromResult(ResultBox.Error<IMultiProjectionPayload>(
+                new Exception($"Projector not found: {projectorName}")));
         }
 
-        if (source is null)
-        {
-            return ResultBox.Error<IMultiProjectionPayload>(new ArgumentNullException(nameof(source)));
-        }
-
-        if (!_projectors.TryGetValue(projectorName, out var registration))
-        {
-            return ResultBox.Error<IMultiProjectionPayload>(new Exception($"Projector not found: {projectorName}"));
-        }
-
-        try
-        {
-            return ResultBox.FromValue(
-                await registration.DeserializeFromStream(source, domainTypes, safeWindowThreshold, cancellationToken)
-                    .ConfigureAwait(false));
-        }
-        catch (Exception ex)
-        {
-            return ResultBox.Error<IMultiProjectionPayload>(ex);
-        }
+        return _streamingRestore.DeserializeAsync(
+            projectorName,
+            domainTypes,
+            safeWindowThreshold,
+            source,
+            cancellationToken);
     }
 
     public ResultBox<IMultiProjectionPayload> DeserializeJson(
--- a/dcb/src/Sekiban.Dcb.Core.Model/Domains/AotMultiProjectorTypes.cs
+++ b/dcb/src/Sekiban.Dcb.Core.Model/Domains/AotMultiProjectorTypes.cs
@@ -13,7 +13,7 @@ namespace Sekiban.Dcb.Domains;
 ///     AOT-compatible implementation of ICoreMultiProjectorTypes.
 ///     Uses JsonTypeInfo for serialization instead of reflection.
 /// </summary>
-public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes
+public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes, IStreamingMultiProjectorTypes
 {
     private static volatile bool _debugBypassProject;
     private readonly Dictionary<string, ProjectorRegistration> _projectors = new();
@@ -27,7 +27,8 @@ public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes
         string Version,
         Func<DcbDomainTypes, string, IMultiProjectionPayload, SerializationResult> Serialize,
         Func<Stream, DcbDomainTypes, string, IMultiProjectionPayload, SerializationSizeInfo> SerializeToStream,
-        Func<DcbDomainTypes, string, byte[], IMultiProjectionPayload> Deserialize);
+        Func<DcbDomainTypes, string, byte[], IMultiProjectionPayload> Deserialize,
+        Func<Stream, DcbDomainTypes, string, CancellationToken, Task<IMultiProjectionPayload>> DeserializeFromStream);
 
     /// <summary>
     ///     Register a multi-projector with AOT-compatible JsonTypeInfo.
@@ -77,7 +78,11 @@ public sealed class AotMultiProjectorTypes : ICoreMultiProjectorTypes
             {
                 var jsonBytes = GzipCompression.Decompress(data);
                 return JsonSerializer.Deserialize(jsonBytes, typeInfo)!;
-            }
+            },
+            DeserializeFromStream: async (source, _, _, cancellationToken) =>
+                await StreamSnapshotPayloadDeserializer.DeserializeJsonAsync(source, typeInfo, cancellationToken)
+                    .ConfigureAwait(false)
+                ?? throw new InvalidOperationException($"Failed to deserialize {typeof(TProjector).Name}")
         );
     }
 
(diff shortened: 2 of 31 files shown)


tomohisa (commented):
Semantic review at exact head `0fbc02c86f07cfc6813c82cc6238aedd6dd467f9`

**Verdict: REQUEST-UPDATE.** The core offloaded seam is directionally correct: production no longer calls `ResolveInlineAsync`, the primary JSON path reaches `GZipStream` plus `JsonSerializer.DeserializeAsync`, and `SnapshotRestoreTeeReadStream` copies the second payload only to a bounded-buffer `FileStream`. However, five packet requirements are not yet proved or enforced. Green tests are necessary, not sufficient.

## Required updates

1. **AC1(c), primary structural proof — the controlled memory counter-proof is not the required discriminant.** `Controlled_18MiB_offloaded_gzip_fixture_uses_the_same_nonbuffering_production_seam` builds one 18 MiB Base64 string and keeps the source actor, source graph, blob `byte[]`, and target restore in the same test process. The restored graph therefore is not small relative to wire size, there is no process-isolated allocation/peak ceiling, and `wholePayloadAggregations: 0` is a literal supplied by the test rather than an observation of the production seam. The 143 MiB workflow is process-isolated and logs peak RSS, but uses the same large-string graph and its 6 GiB limit would not kill an added 143 MiB whole-payload buffer. The intentionally buffered registry only increments its own test-owned counter, so it does not prove that a production `MemoryStream`/`ToArray` regression would be caught. Add the packet's small-graph/large-wire fixture and a meaningful isolated allocation/peak ceiling or a production-seam counter/structural guard that the intentionally buffered control and a production aggregation mutation both fail. Read-size caps may remain supporting evidence only.

2. **AC3, equivalence matrix — the required cross-product and metadata counter-proofs are absent.** The current tests separately sample reflection gzip, AOT gzip, custom raw, reflection raw, and inline v9/v10, but there is no explicit matrix across reflection/AOT/custom × gzip/raw legacy × v9/v10 × offloaded/inline. More importantly, successful offloaded cases generally assert payload only; they do not compare stream versus legacy byte paths for safe threshold, version, last event id, and last sortable id. The docs table classifies broad routes, not the required equivalence matrix. Add a parameterized matrix/table with exact payload and tracking metadata assertions, plus a projector that records the safe threshold supplied to both deserializers. A regression that drops/changes any one metadata field or routes one matrix cell to the wrong guarantee/fallback must fail.

3. **AC5, absent-vs-failure — open and read failures lack independent zero-fallback killing tests.** Capability absence and capability-level deserialize failure are covered, and corrupt gzip covers decompression, but no blob `OpenReadAsync` throw or source `ReadAsync` throw is injected through the production resolver/host path. Source inspection shows those exceptions currently propagate, but the packet explicitly requires open/read/decompress/deserialize counterexamples with zero buffered retry. Add distinct original-exception tests for open and mid-stream read failure, asserting zero legacy `Deserialize(byte[])`, zero capability-absent log, and no second attempt.

4. **AC7, fail-before-serving — a failed restore clears the only guard and makes the previous state usable again.** `SetCurrentStateCoreAsync` unconditionally clears `_snapshotRestoreInProgress` in `finally` and records no restore fault/quarantine. The truncated/corrupt tests then call `GetStateAsync` and explicitly obtain the pre-restore payload; `AddEventsAsync` would likewise proceed after the exception. The production host test only blocks query/apply/persist while a restore that eventually succeeds is still pending; it does not test a terminal restore failure through the host/grain activation boundary. This does not kill stale serving or catch-up/apply after failure. Keep atomic rollback, but make the activation/caller fail closed until the documented recovery/rebuild or a successful restore, and add a real failure-path call-chain test proving query, catch-up/apply, promotion, and persistence cannot use the old/partial actor state.

5. **AC10, additive public API inventory — reflection pins are incomplete.** A local API diff confirms no breaking change in Core.Model/Core, but identifies additive surfaces not pinned by `Streaming_restore_public_surface_is_additive_and_the_legacy_registry_contract_is_frozen`: public `StreamSnapshotPayloadDeserializer` and its two methods; the new `IStreamingMultiProjectorTypes` members on `AotMultiProjectorTypes`; the corresponding public AOT WithoutResult and WithResult/WithoutResult `SimpleMultiProjectorTypes` implementations; and the complete `ResolvedSnapshotRestore` method/constructor accessibility shape. Pin and inventory every new public type/member/interface implementation, not only the two interfaces, three properties, resolver method, and actor method currently asserted. `ICoreMultiProjectorTypes` itself is byte-for-byte source unchanged and the frozen 10.18 assembly does load/invoke, which should remain as counter-proofs.

## AC1–AC11 matrix

| AC | Production path | Killing evidence reviewed | Result |
|---|---|---|---|
| AC1 | `NativeProjectionSnapshotHandler` -> `ResolveForRestoreAsync` -> `SetResolvedSnapshotAsync`; offloaded state carries a non-seekable stream, and `DeserializeStreamingPayloadPairAsync` tees only to an 81,920-byte-buffered temporary file. No `ResolveInlineAsync`, `ReadAllBytesAsync`, `GetPayloadBytes`, `MemoryStream`, `ToArray`, or full string occurs on the supported branch. | Production host and capped-stream tests prove the route/read behavior; buffered fake returns false. The mandatory small-graph/large-wire isolated ceiling and production aggregation mutation discriminant are missing as described above. | **FAIL** |
| AC2 | `StreamingMultiProjectorTypesSupport.RegisterReflection` calls `StreamSnapshotPayloadDeserializer`, whose gzip branch constructs `GZipStream` and calls `JsonSerializer.DeserializeAsync`; AOT registration uses the `JsonTypeInfo<T>` overload. | Real reflection gzip and AOT gzip tests restore exact payloads through the actual built-in registries with zero byte-deserialize calls. | **PASS** |
| AC3 | Stream success and byte fallback converge through the same wrapper construction and final metadata publication block. Per-projector registration removes streaming support for non-opted-in custom serializers. | Individual route samples pass, but the required explicit cross-product and exact safe-threshold/version/last-id equivalence assertions do not exist. | **FAIL** |
| AC4 | New optional interfaces are separate from `ICoreMultiProjectorTypes`; old registries take the fallback. | `ICoreMultiProjectorTypes` source is unchanged; `Frozen_10_18_external_registry_binary_loads_and_is_invoked...` loads the separately built 10.18 NuGet-based assembly and observes two calls. | **PASS** |
| AC5 | Unsupported projector/registry selects the sole `ReadAllBytesAsync` branch and logs once; a selected streaming capability never enters that branch after an error. | Absence, capability error, and corrupt gzip are covered. Distinct production open/read throws with original exception and zero fallback are missing. | **FAIL** |
| AC6 | Prefix buffering is two bytes, preserves current position, never seeks, forwards cancellation, and leaves the underlying source open; resolver owner disposes it. | Cancellation, partial capped reads, current-position prefix, Length/Seek prohibition, caller ownership, truncated and corrupt payload tests pass. | **PASS** |
| AC7 | Actor publishes payload plus unsafe id/position/version/catch-up only after both graph deserializations and wrapper creation. In-flight guards cover reads, apply, promotion/compaction, head and persistence. | In-flight production host test passes, and failure rollback is atomic. Terminal failure clears the guard and the tests then serve the previous state; no production failure call-chain kills stale query/catch-up/apply. | **FAIL** |
| AC8 | Production callers found: grain -> native host/handler; offline builder; inline primitive; direct legacy state methods; explicit `ResolveInlineAsync` adapter. Both offloaded production callers await the shared resolver/actor seam. | EN/JA caller-inventory tables match the traced source call graph; the native incident/OOM entry is exercised for raw v10 and outer-gzip v9 envelopes. | **PASS** |
| AC9 | Ordinary test category uses an 18 MiB fixture; separate weekly/manual workflow uses a 143 MiB fixture, timeout, virtual-memory limit and telemetry. | Size/category split and opt-in workflow are present; no no-OOM assertion is made. AC1's quality of the controlled memory counter-proof remains failed separately. | **PASS** |
| AC10 | Optional public contracts are additive; existing interface is unchanged. | Frozen binary and partial reflection test pass; API diff shows no removals, but the additive inventory/pins omit public surfaces listed in finding 5. | **FAIL** |
| AC11 | EN/JA docs state the limited offloaded-only guarantee, explicitly exclude inline Base64, describe absence fallback versus genuine failure, temp-file clone, caller ownership, inventory and CI split. | Both language sections match `intents/sekiban/design/restore-memory-overlap.md`; no no-OOM, save-side, compression-change, or graph-shrinking claim found. | **PASS** |

## Verification and scope

- `dotnet test dcb/tests/Sekiban.Dcb.WithResult.Tests/Sekiban.Dcb.WithResult.Tests.csproj -f net9.0 --filter "FullyQualifiedName~StreamingSnapshotRestoreTests"` — 18/18 passed.
- Same command for `net10.0` — 18/18 passed.
- `dotnet test dcb/tests/Sekiban.Dcb.Orleans.Tests/Sekiban.Dcb.Orleans.Tests.csproj -f net9.0 --filter "FullyQualifiedName~StreamingSnapshotRestoreProductionPathTests"` — 2/2 passed.
- Same command for `net10.0` — 2/2 passed.
- Full PR patch applied to exact base `c9d1d87df0bde5d6a39d7c7a8573186c2888c2fb` with `git apply --check --whitespace=error-all`: passed.
- Exact-head terminal checks: `dcbTestsNet9`, `dcbTestsNet10`, `SonarCloud`, and `SonarCloud Code Analysis` all completed successfully. Sonar reports 15 new code smells and no BUG/reliability issue; none changes this semantic verdict.
- Base is `main`; body retains `Closes #1156` followed by `Fixes #1145`.
- Scope remains restore-side: no compression-format change, graph shrinking, save-side guarantee, release/tag, or no-OOM promise was added.

All five findings are implementation findings tied to the current packet; no host-metadata or intent ambiguity finding is being posted here.

tomohisa (commented):
Semantic re-review at exact head `8a98cd11dbedd0311462662d5e47243a51d6cf06`

**Verdict: REQUEST-UPDATE.** The repair genuinely resolves the prior AC5, AC7, and AC10 blockers, and most of AC3. The current production implementation is stream-first and the focused suites are green, but AC1's required path-wide aggregation mutation proof and AC3's exact safe-threshold equivalence proof remain incomplete. Tests are necessary, not sufficient.

## Required updates

1. **AC1(b/c/d), primary structural proof is still not path-complete.** `Supported_stream_restore_state_machines_have_no_whole_payload_aggregation_api_reference` scans only the direct IL member references of three entry state machines. It does not recursively follow referenced production helpers and does not scan `SnapshotRestoreTeeReadStream` or the private `StreamSnapshotPayloadDeserializer.PrefixBufferedStream.CreateAsync` state machine. Consequently, moving/adding `MemoryStream` + `ToArray`, `ReadAllBytes*`, or another whole-payload aggregation inside either helper survives the structural test. The production counter also increments only through `StreamReadHelper.ReadAllBytesAsync`, so those mutations remain reported as zero. The intentionally buffered control does now enter the real scope and produces count 2, which is useful, but it proves only the shared helper case. In addition, `Controlled_small_graph_large_wire...` is still an in-process functional test with no allocation/peak ceiling; the isolated 143 MiB job still uses a correspondingly large Base64 graph and a 6 GiB ceiling, so it does not supply the packet's small-graph/large-wire isolated supporting counter-proof. Make the scanner/counter cover the transitive resolver → actor/tee → stream-deserializer path (or add equivalent mutations/tests), and add the specified controlled isolated ceiling.

2. **AC3, exact safe-threshold equivalence is not established.** The 18 representable matrix cells now give exact payload, version, event id, sortable id, catch-up/safe-state, and routing assertions (with gzip-v9 correctly identified as unrepresentable). However, `AssertRecordedThresholdsWithinWindow` only checks that each recorded threshold falls between two independently generated wall-clock values. It never asserts stream and byte paths receive the identical threshold, nor that the two stream deserializations receive the same threshold. A mutation that recomputes a threshold per primary/clone call, or substitutes any different in-window value, remains green. Capture/assert the exact actor-selected threshold for both byte and stream paths, including primary and clone, so wrong threshold routing is killed as required by AC3.

## Prior blocker disposition and AC1–AC11 matrix

| AC | Production path | Killing evidence reviewed | Result |
|---|---|---|---|
| AC1 | `NativeProjectionSnapshotHandler.RestoreSnapshotFromStreamAsync` → `SnapshotEnvelopeResolver.ResolveForRestoreAsync` → `GeneralMultiProjectionActor.SetResolvedSnapshotAsync`; supported restore tees to an 81,920-byte-buffered temporary `FileStream`, never the inline resolver. | Non-seekable/capped source, small-graph 25 MiB wire, actor scope counter, and buffered control count 2 pass. Direct-only IL scan and the unrelated large-graph 6 GiB smoke do not kill the transitive/helper aggregation mutations described above. | **FAIL** |
| AC2 | Reflection/AOT registrations call `StreamSnapshotPayloadDeserializer`; gzip constructs real `GZipStream`, then reflection/AOT `JsonSerializer.DeserializeAsync` overloads. | Reflection and AOT raw/gzip matrix cells restore exact payload through real registries; no fake piecewise-reader substitute. | **PASS** |
| AC3 | Actor computes one restore threshold and publishes payload plus version/event/sortable metadata only after both graphs deserialize. | 18 representable registry × format × wire × delivery cells kill payload/version/id/routing changes, but threshold is range-only rather than exact as finding 2 explains. | **FAIL** |
| AC4 | Optional capability remains separate from `ICoreMultiProjectorTypes`; legacy registries use the explicit absence fallback. | `ICoreMultiProjectorTypes` is byte-identical to base; frozen 10.18 fixture loads and is invoked in the focused suite. | **PASS** |
| AC5 | Blob open occurs in resolver; once capability is selected, stream errors propagate without entering the sole absence fallback. | Distinct production-host `OpenReadAsync` and third-read failures preserve the same injected exception; open count 1, stream call 0/1, byte calls 0, no absence log or retry. Prior blocker **resolved**. | **PASS** |
| AC6 | Two-byte prefix wrapper preserves current position, never seeks, forwards cancellation, and leaves the caller-owned stream open; resolver owner disposes it. | Existing cancellation, partial-read, non-seekable/Length, current-position, ownership, truncated and corrupt cases remain green. | **PASS** |
| AC7 | `_snapshotRestoreFailure` quarantines an already-published accessor; every query/apply/head/promotion/compaction/snapshot boundary consults the shared blocker; only successful replacement restore clears it. | Production host test proves atomic rollback and blocks query, metadata, catch-up, apply, promotion, compaction, both write paths (zero bytes), then proves successful recovery. Separate tests preserve fresh-actor mismatch empty/rebuild behavior and quarantine replacement mismatch. Prior blocker **resolved**. | **PASS** |
| AC8 | Production incident caller and builder both await resolver → actor; inline primitive, direct state APIs, and `ResolveInlineAsync` remain explicit compatibility paths. | Source call graph matches the EN/JA inventory; production host exercises raw-v10 and outer-gzip-v9 incident routes. | **PASS** |
| AC9 | Ordinary suite owns the 16–32 MiB fixture; weekly/manual workflow owns the 143 MiB smoke, timeout, telemetry and virtual limit. | Category/workflow split remains correct; it makes no no-OOM claim. AC1's quality of the controlled ceiling is failed separately. | **PASS** |
| AC10 | New interfaces/helper/restore input and registry forwarding are additive; repair adds no further API. | Reflection pins cover both helper overloads, both interfaces, all five concrete/internal registry implementations across WithResult/WithoutResult/Core/AOT, full `ResolvedSnapshotRestore` property/constructor/disposal shape, resolver and actor methods. Local old-head→new-head API diff reports no change; frozen 10.18 invoke remains green. Prior blocker **resolved**. | **PASS** |
| AC11 | EN/JA docs limit the guarantee to capability-present offloaded restore, exclude inline Base64, describe absence-vs-failure, disk tee, ownership, quarantine, caller inventory, and CI split. | Both language sections retain the design wording; no no-OOM, save-side, graph-shrinking, or compression-change promise. | **PASS** |

## Verification, compatibility, and scope

- WithResult `FullyQualifiedName~StreamingSnapshotRestoreTests`: net9 **39/39**, net10 **39/39** passed.
- WithResult `FullyQualifiedName~GeneralMultiProjectionActorCatchingUpTests.SetCurrentState`: net9 **2/2**, net10 **2/2** passed.
- Orleans `FullyQualifiedName~StreamingSnapshotRestoreProductionPathTests`: net9 **5/5**, net10 **5/5** passed.
- WithoutResult `FullyQualifiedName~StreamingSnapshotRestorePublicInventoryTests`: net9 **1/1**, net10 **1/1** passed.
- Full PR patch applies to exact base `c9d1d87df0bde5d6a39d7c7a8573186c2888c2fb` with `git apply --check --whitespace=error-all`.
- Exact-head checks are terminal green: `dcbTestsNet9`, `dcbTestsNet10`, `SonarCloud`, and `SonarCloud Code Analysis`; SonarCloud reports zero new BUG issues.
- Base is `main`; body retains `Closes #1156` followed by `Fixes #1145`.
- Scope remains restore-side and additive: no save-side guarantee, compression-format change, graph shrinking, no-OOM promise, tag, or release.

These are implementation/test-evidence findings tied specifically to AC1 and AC3; no host-metadata or intent ambiguity finding is included.

tomohisa (commented):
## Semantic verdict: APPROVE-READY

Reviewed the complete PR at exact head `83514b5b57b1abf22fd3e4ef7e7ae48214956018` against the SEK-G42 packet/design, with the repair diff from `8a98cd11dbedd0311462662d5e47243a51d6cf06` inspected first. The prior AC1 and AC3 blockers are resolved. I found no remaining semantic blocker; tests are used below as killing discriminants, not as a substitute for tracing the production path.

### AC evidence matrix

| AC | Production path and killing discriminant | Result |
|---|---|---|
| AC1 — structural proof | `NativeProjectionSnapshotHandler -> SnapshotEnvelopeResolver.ResolveForRestoreAsync -> GeneralMultiProjectionActor.RestorePayloadAsync -> DeserializeStreamingPayloadPairAsync -> SnapshotRestoreTeeReadStream -> StreamSnapshotPayloadDeserializer` carries the offloaded source without `ResolveInlineAsync`, `ReadPayloadBytesAsync`, or `GetPayloadBytes`. The repaired recursive IL walk normalizes nested async state machines and reaches the resolver, actor, tee `ReadAsync`, `PrefixBufferedStream.CreateAsync`, and stream support/deserializer graph. Its self-counterexamples hide `MemoryStream`/`ToArray`, `File.ReadAllBytesAsync`, and `GetPayloadBytes` behind async helpers and are all rejected; the forbidden inventory also covers equivalent aggregation families. The same-wire 25 MiB small-graph fixture, isolated with `SEKIBAN_STREAM_RESTORE_CONTROLLED_CEILING=1`, measured supported allocations at 162,569,544 B (net9) / 196,127,440 B (net10), below 256 MiB, with production aggregation counter 0; the intentionally buffered control measured 787,996,648/787,996,808 B and counter 2. This would fail if aggregation were reintroduced anywhere in the closed supported graph. It is supporting evidence only, not a no-OOM claim. | **PASS** |
| AC2 — real JSON+gzip path | Built-in reflection/AOT streaming implementations route through `StreamSnapshotPayloadDeserializer`, which constructs `GZipStream` for gzip and invokes `JsonSerializer.DeserializeAsync`/`JsonTypeInfo`; the production host test wraps the capability seam but delegates to that real implementation and requires stream calls 2, byte calls 0. A fake chunk reader cannot satisfy these assertions. | **PASS** |
| AC3 — equivalence matrix | `RestorePayloadAsync` now calls `SelectSnapshotRestoreSafeWindowThreshold()` exactly once and forwards that exact string to byte or stream primary/clone calls. The 18 representable cells cover reflection/AOT/custom × raw-v9/raw-v10/gzip-v10 × offloaded/inline (gzip-v9 is explicitly unrepresentable). Every cell asserts exact payload, version, last event/sortable IDs, exact call counts, and equality of every recorded primary/clone threshold to the actor-selected value. The transitive structural pin permits exactly one `GetSafeWindowThreshold` under restore and none in the stream-pair graph, killing helper-side recomputation. | **PASS** |
| AC4 — binary compatibility | `ICoreMultiProjectorTypes` remains unchanged; the optional capability is separate. `Frozen_10_18_external_registry_binary_loads_and_is_invoked_through_the_buffered_compatibility_path` loads the separately compiled 10.18 fixture and invokes it on the current runtime, proving the legacy buffered route remains callable. | **PASS** |
| AC5 — absence versus failure | Capability absence takes exactly one observable buffered fallback/log with projector/registry, format, and `reason=capability-absent`, without payload content. Distinct production-host tests inject `OpenReadAsync` failure and mid-stream `ReadAsync` failure; both require the original exception, zero byte fallback, zero absence log, and no retry. Present deserialize/corrupt/truncated failures likewise cannot select the buffered route. | **PASS** |
| AC6 — stream contract | The production seam accepts the caller token, reads a non-seekable stream from its current position under partial reads, never calls `Length`/`Seek`, and leaves caller-owned streams undisposed at the callee boundary. Cancellation, current-position, ownership, truncation, and corruption tests distinguish each failure mode. | **PASS** |
| AC7 — fail before serving | The real `NativeProjectionActorHost` test blocks an in-progress restore and proves query, apply/catch-up, head/promotion, and persistence cannot proceed. Terminal failure latches quarantine across query/apply/catch-up/promotion/compaction/persist, publishes neither prior nor partial state, and successful restore clears the latch. Separately, a fresh actor version mismatch retains the legacy empty-state/rebuild behavior, while a mismatch after a published checkpoint quarantines that checkpoint; the two `SetCurrentState` tests prevent conflating these states. | **PASS** |
| AC8 — caller inventory | The production OOM entry is exercised through `NativeProjectionActorHost` and `NativeProjectionSnapshotHandler`, not a helper. The EN/JA inventory classifies native state-store restore, `MultiProjectionStateBuilder`, inline APIs, and explicit fallback entries, including async waiting and the offloaded-only guarantee. | **PASS** |
| AC9 — CI size split | Normal tests retain the 16–32 MiB controlled structural/counter fixture. `.github/workflows/dcb_streaming_restore_memory.yml` runs the controlled ceiling case in its own process and keeps the 143 MiB smoke weekly/manual with timeout and virtual-memory ceiling, recording path/counters rather than asserting “no OOM.” | **PASS** |
| AC10 — additive public API | Reflection inventory pins `IStreamingMultiProjectorTypes`, both stream methods/signatures, all WithResult/WithoutResult/AOT/reflection implementations, the complete `ResolvedSnapshotRestore` shape, and the stream deserializer surface. `ICoreMultiProjectorTypes` is frozen unchanged. The repair adds only a private actor field/helper and tests/docs; no new public delta was introduced after the prior review. The 10.18 load/invoke counter-proof remains green. | **PASS** |
| AC11 — EN/JA docs | Both languages state the limited offloaded guarantee, explicitly exclude inline Base64, explain capability-absent fallback versus capability failure, and describe normal controlled evidence versus separate large smoke. The repaired wording accurately adds the isolated same-wire buffered ceiling control without making a no-OOM promise. | **PASS** |

### Focused execution evidence

- `dotnet test ...Sekiban.Dcb.WithResult.Tests.csproj -f net9.0/net10.0 --filter FullyQualifiedName~StreamingSnapshotRestoreTests`: **42/42 passed on each TFM**. This includes the recursive aggregation scanner/self-counterexamples, exact-threshold matrix, real gzip/JSON paths, fallback/failure split, stream contract, compatibility fixture, public inventory, and controlled fixture.
- Same controlled test in a dedicated process with `SEKIBAN_STREAM_RESTORE_CONTROLLED_CEILING=1`: **1/1 passed on net9 and net10**, with the allocation/counter telemetry reported under AC1.
- `dotnet test ...Sekiban.Dcb.Orleans.Tests.csproj -f net9.0/net10.0 --filter FullyQualifiedName~StreamingSnapshotRestoreProductionPathTests`: **5/5 passed on each TFM** (real host path, distinct open/read failures, terminal quarantine/recovery).
- `dotnet test ...Sekiban.Dcb.WithResult.Tests.csproj -f net9.0/net10.0 --filter FullyQualifiedName~GeneralMultiProjectionActorCatchingUpTests.SetCurrentState`: **2/2 passed on each TFM**.
- Full PR patch applies cleanly to base `c9d1d87df0bde5d6a39d7c7a8573186c2888c2fb` with `git apply --check --whitespace=error-all`.

GitHub exact-head checks are terminal green: `dcbTestsNet9`, `dcbTestsNet10`, `SonarCloud`, and `SonarCloud Code Analysis`; the Sonar pull-request BUG query returns 0. Base is `main`; the body retains `Closes #1156` followed by `Fixes #1145`. Scope remains restore-side streaming/capability, tests, workflow, and EN/JA docs; there is no save-side/compression/graph-shrinking or no-OOM expansion.

**Final semantic verdict: APPROVE-READY at `83514b5b57b1abf22fd3e4ef7e7ae48214956018`.**

Issue #1193, opened 2026-09-08
DCB: bind expected-tag CAS to state reads from one dynamic command execution


joemphilips:
## Summary

Please document or add a supported typed-command path that derives PostgreSQL
expected-tag positions from the state used by that command's single handler
invocation. This is an integration question and feature request. It is not a
report that the PostgreSQL CAS transaction accepts a stale expectation.

Version: DCB 10.20.0. All restored DCB package metadata records name commit
`b7f2386c41f90712ac65633efaa0f2b03b64c882`.
Related: #1147 and its released multi-tag CAS capability.

## Concrete command shape

A normal state-dependent command does the following inside its handler:

```text
read tag A through ICommandContext.GetStateAsync
if the operation is already complete:
    return no event
if the state names another affected entity B:
    read tag B through the same context
append one or more events with the affected consistency tags
```

The affected tag set and no-op branch depend on those reads. We want the
database to compare the positions from those same reads. The handler must run
once per attempt. On a conflict, a new attempt must make a new decision.

## Observed source behavior

- `OrleansDcbExecutor` implements
  `ISerializedExpectedTagPositionSekibanDcbExecutor`, but not the typed
  `IConditionalCommandExecutor` interface.
- `GeneralSekibanExecutor` implements the typed interface. A public constructor
  accepts an actor accessor, so composition with an Orleans accessor appears
  possible. This does not by itself address expectation capture.
- `CommandExecutionOptions.ExpectedTagPositions` is supplied before the
  handler. There is no decision-completion expectation factory in that type.
- `CoreGeneralSekibanExecutor.ExecuteAsyncCore` validates entry shape and
  provider capability before invoking the handler. After the handler, it
  validates equality with the emitted consistency-tag set. A no-event result
  validates against an empty set.
- The reservation path already reads the command context's accessed tag
  states. The expected-position path does not derive its specification from
  those observations.

Consequently, a nonempty expectation list supplied for the potential append
does not match the no-event branch. An empty list does not match the append
branch. Pre-reading the state in a caller also duplicates the handler's
decision reads and requires a way to keep their versions bound.

This report is based on the pinned source. We have not run a new PostgreSQL
reproduction for this API-shape question.

## Requested supported contract

Please first confirm whether an existing supported pattern covers this case.
If not, an explicit opt-in mode could derive the expectation set after one
handler invocation, from its captured state positions and emitted consistency
tags. This is a suggested shape, not a request to enable CAS by default.

[...]

PR #1198, opened 2026-09-09
feat(dcb): derive expected tag positions from state reads

### LINKED ISSUE #1197: SEK-G65 dcb: derive expected tag positions from command state reads
## Goal

Derive exact expected tag positions from the successful state reads made by one command-handler invocation, after its emitted consistency-tag set is known.

## Why This Slice Exists Now

Issue1193 needs dynamic A/B state reads and a no-event branch without duplicated handler execution, undocumented mutation or caller-built execution plumbing. This is the next serial implementation after G56/G57/G58 closeout; authoring does not start implementation.

## Current Observed State

Existing options accept pre-supplied ExpectedTagPositions. CoreGeneralCommandContext reservation tracking overwrites observations and also tracks existence-only evidence; it cannot certify the full set of compatible payload reads. Typed TagState conversions may contain an empty SUID placeholder. Existing PostgreSQL CAS validates an exact emitted-tag set transactionally and remains the write boundary.

- External issue1193 by joemphilips,2026-09-08T02:28:14Z. Operator authorized investigation/packet design and explicitly reserved the next serial implementation slot after G56/G57/G58 individual closeouts. This is a deferred packet only: do not publish/promote an issue or begin implementation before all three closeouts. G56/G57 complete; G58 repair/review/CI is still in progress at packet authoring.
- Fable accepted design candidate41d9bd1 through canonical sek-issue1193-b3-text-rereview after B1/B2/B3 corrections; preserve that accepted contract and N9. Fable independently approved packetbd4d1a5 with P1-a/P1-b refinements now incorporated; this remains design-only approval behind the serial publication gate.
- Source investigation at main664d775 and issue pinb7f2386 found no dynamic decision-read capture API. Explicit ExpectedTagPositions works as pre-supplied total specification; this is an additive integration gap, not evidence of a broken PostgreSQL CAS transaction. G57/G58 changes are MV-specific; revalidate command source against actual landing main when implementing.
- Selected public selector name is DeriveExpectedTagPositionsFromStateReads; init-only false default. Public StateReadRecoveryStatus enum and ExpectedTagPositionConflictException.StateReadRecovery property are additive, preserving legacy constructor/pairs/default. No new serialized protocol, schema, package version or release.

## Target Repo / Path / Part

Repository: J-Tech-Japan/Sekiban

- Target paths: dcb/src/Sekiban.Dcb.Core dcb/src/Sekiban.Dcb.WithResult/Actors dcb/src/Sekiban.Dcb.WithoutResult/Actors dcb/src/Sekiban.Dcb.Orleans.WithResult dcb/src/Sekiban.Dcb.Orleans.WithoutResult dcb/tests/Sekiban.Dcb.WithResult.Tests dcb/tests/Sekiban.Dcb.WithoutResult.Tests dcb/tests/Sekiban.Dcb.Orleans.Tests dcb/tests/Sekiban.Dcb.Postgres.Tests dcb/tests/Sekiban.Dcb.WithResult.BinaryConsumer dcb/tests/Sekiban.Dcb.WithoutResult.BinaryConsumer docs/dcb_llm/03_aggregate_command_events.md docs/dcb_llm_ja/03_aggregate_command_events.md docs/dcb_llm/11_storage_providers.md docs/dcb_llm_ja/11_storage_providers.md

Target part: Attempt-local raw state-read evidence, emitted consistency-tag derivation, typed Orleans forwarding and between-attempt conflict recovery.

## In Scope

An explicit false-default derived mode, framework-owned attempt-local raw read ledger, emitted-tag-only total expected-position validation, additive General/Orleans surface parity, original-conflict-preserving actor invalidation between attempts, discriminating tests and EN/JA docs.

## Out Of Scope

Implicit read-set serializability, handler reinvocation/automatic retry, post-decision replacement evidence, changed PostgreSQL schema/transactions/enablement, new serialized protocol, unconditional/explicit legacy behavior changes, MV fixes, default template behavior, release/version changes.

### A. Explicit selection and ordering

Add an init-only opt-in selector on CommandExecutionOptions, `DeriveExpectedTagPositionsFromStateReads`. Default false means the existing behavior. Reject combining this mode with explicit ExpectedTagPositions or ConditionalAppend before invoking the handler. Do not offer a user-supplied mutable list or arbitrary post-handler callback as the evidence authority.

Resolve and validate the service identity and live ExpectedTagPosition capability before the handler. In derived mode require the service enforcement epoch before the handler even when the eventual outcome is no-event: this is explicit enforcement, not a best-effort mode. Unsupported capability or missing epoch must yield the existing typed failure without handler execution, reservation, ID allocation, publication or event/head mutation. Reads needed to check capability/epoch are permitted; no claim of zero I/O. Recheck existing service/transaction invariants at their existing boundaries; never mix evidence from a different service.

Invoke the handler exactly once per execution attempt. Do not automatically retry or re-invoke it. Select framework actor invalidation after a derived-mode durable `ExpectedTagPositionConflictException`, rather than promising that another ordinary cached read alone is fresh. After the failed write and existing reservation cleanup, mark the affected emitted tags in this service for re-catch-up through the existing `ITagConsistentActorCommon.NotifyEventWrittenAsync` mechanism (`GeneralTagConsistentActor.cs:256-261`). Await notification before returning the original typed conflict. Do not confirm reservations, publish events, fetch a replacement head, or retry the failed append. This changes only derived-mode conflict recovery; explicit/serialized legacy paths keep their current behavior.

A caller may then begin a bounded new attempt with a new ledger and one fresh handler invocation. The next actor access performs normal store catch-up; `GeneralTagStateActor` compares that head to its cached state cursor and rebuilds/catches up when different. This is recovery between attempts, never permission to replace evidence inside an attempt. Notification/accessor failure must preserve the original expected/observed conflict and be observable through a new additive, publicly readable `StateReadRecovery` property on `ExpectedTagPositionConflictException`, backed by public enum `StateReadRecoveryStatus` with `NotAttempted = 0`, `InvalidationCompleted = 1`, and `InvalidationIncomplete = 2`. Keep the existing constructor and expected/observed pairs unchanged; the property has a framework-internal setter and defaults to NotAttempted on legacy paths. Set it before returning the original conflict object: Completed only when every affected tag's invalidation call completed successfully, Incomplete if any accessor or notification failed. Complete the bounded per-tag attempts and preserve cleanup; do not use the existing swallowing Debug.WriteLine-only helper as outcome evidence. This is an additive observable surface, not an existing Release diagnostic facility.

InvalidationCompleted means only that actor heads were invalidated, not that catch-up has already completed or that retry will certainly succeed. On InvalidationIncomplete, callers must stop automatic bounded retries and resynchronise the actor/session through operational recovery; the original conflict still reports the original positions and remains the typed failure. Do not claim unconditional progress under notification failures or continued concurrent writes. External effects remain the caller's responsibility. No logger constructor/DI change or failure-path ExecutionResult.Metadata is required for this contract.

### B. Evidence belongs to the successful decision reads

Capture successful GetStateAsync results from the framework-owned raw read boundary before returning the state to the handler. Retain validated logical tag identity, service identity, projector identity/version, and raw last-applied SUID/empty evidence. Do not re-fetch a tag head after the decision to manufacture a newer expected position. Existing reads inherently needed to build GetStateAsync and existing reservation/PG reconciliation reads remain allowed; the prohibition is on replacing decision evidence with a post-decision head lookup.

Use an attempt-local observation ledger independent of the legacy reservation dictionary. Key it by normalized `GetTag()` string with ordinal comparison, unwrapping `ConsistencyTag.InnerTag`; never key by ITag object equality, since custom implementations need not override it. Successful repeated observations of the same normalized logical tag at the same position are compatible, including different projectors with the same position. Different positions, empty-versus-nonempty observations, malformed evidence, and successful evidence mixed with a failed state-read attempt on an emitted tag fail closed; never pick first/latest/max or discard the disagreeing observation. Bound stored evidence per distinct tag by one agreed position, provenance summary and an ambiguity/failure flag; do not accumulate an unbounded history of repeated reads.

Missing actor/accessor errors must not become certified empty evidence merely because an existing legacy context path synthesizes EmptyTagStatePayload. Both synthetic-empty branches leave the tag unread in the new ledger, so a later emit fails as emitted-but-unread; a prior actual successful read combined with that failure is ambiguous and also rejects. A successfully returned framework state with version zero and an empty raw cursor is the supported empty case; nonempty valid cursor is Exact. A positive version with missing cursor or mismatched logical tag identity is invalid evidence. Any additional legitimate empty-state shape found during implementation needs explicit design disposition and a discriminating test, not a permissive fallback. Preserve legacy mode behavior. Existence-only TrackObservedVersion entries remain exclusively in legacy reservation bookkeeping and cannot populate the new ledger, preserving G19 behavior.

TagExistsAsync and GetTagLatestSortableUniqueIdAsync alone do not qualify as payload decision reads in this mode. An explicit ConsistencyTag token must not override the ledger: if a supplied token disagrees with captured evidence, reject before reservation. Normalize the tag wrapper to the same logical identity used by emitted consistency tags, without relying on ITag object reference equality. Preserve the original reservation process after validation.

### C. Derive only the emitted consistency-tag set

After the single handler completes successfully and the existing event collection/deduplication step runs, derive exactly one AssertEmpty or Exact entry for each distinct emitted consistency tag. Never generate NoEnforcement in derived mode. Every such tag must have compatible successful state-read evidence; an unread emitted tag is a typed validation failure. Non-consistency tags do not gain expectation checks. Reads of tags absent from the emitted consistency-tag set do not add entries, do not cause set-equality failure, and do not gain a serializable read-set guarantee.

If no events are emitted, return the existing no-op ExecutionResult with zero writes/reservations/ID allocation/publication. Do not impose nonempty expectation-set validation from unused decision reads. Capability/epoch preflight from A still applies. A handler error or cancellation retains its existing failure semantics and does not become a no-op success.

Validate the frozen derived specification through the existing total validator before reservation, ID allocation or write. Submit unchanged event instances through the existing serialized expected-position writer, retaining reservation cleanup, event allocation rules, publication behavior and the PG multi-tag atomic transaction. A reservation may reject before PG; the new API does not promise that every competing call reaches the database. PG conflict evidence must be measured independently with a test barrier/fixture that permits both valid attempts to reach that seam.

### D. Public surfaces and compatibility

Both GeneralSekibanExecutor facades expose the opt-in through existing options overloads. Add typed IConditionalCommandExecutor forwarding and matching options overloads to both Orleans facades, preserving existing constructors/method signatures and guarded error behavior. Prove DI resolves an Orleans executor whose conditional interface reaches the same registered service/store, actor accessor, monotonic allocator, executed-user provider and publisher. Do not recommend constructing a second executor that silently changes those dependencies.

The init-only property and additional interface implementation are intended to be binary-additive. Extend `ExpectedTagPositionCompatibilityInventoryTests` and existing binary consumers to prove this, and include source-overload resolution cases using existing default/null arguments; do not introduce ambiguous existing calls. The selector stays a single explicit mode in this candidate; any discriminated-type alternative requires packet API review rather than an implicit read-set mode.

Extend the frozen exception inventory explicitly for the additional read-only public StateReadRecovery property and enum; retain its existing constructor, pairs and legacy default behavior. This bounded additive failure metadata is part of the new derived mode, not a change to explicit/serialized CAS semantics.

Existing explicit ExpectedTagPositions, unconditional APIs, single-event unique-key APIs and serialized V1/V2 contracts remain unchanged. Serialized commit callers have no live handler decision-read context, so no implicit derived mode or new serialized protocol is introduced here. No PostgreSQL schema, enablement policy, head bootstrap, writer fencing, batch transaction, event identity, release/version or default template CAS behavior changes.

## Acceptance Criteria

- AC1: **Dynamic A/B append and no-op:** the same command/opt-in invokes its handler once, reads A and conditionally B, submits precisely the emitted tags with their original positions, and takes the idempotent no-event branch with zero write/reserve/allocate/publish calls. Pair with a mutant that precomputes a fixed nonempty set or runs the handler twice. Use one handler invocation counter for AC1/AC6/AC7; an IActorObjectAccessor decorator counts reservation and NotifyEventWrittenAsync calls, a fake IEventPublisher counts publication, and independent PG command-row absence plus existing TagHeadProtocolHook phase counts distinguish allowed reads from mutation. These instrumentations do not turn fake actor tests into provider-backed CAS proof.

- AC2: **Captured H versus later H2:** use two separately discriminating schedules. For the store-refresh mutant, read state at H and advance the durable tag to H2 through an independent PG writer before derivation; the submitted expectation remains H and PG rejects it with observed H2. This alone does not kill an actor-refresh mutant because an out-of-band writer may leave the actor cached at H. For that mutant, pause after the single handler has returned and event collection has completed, immediately before freezing the derived specification; commit H2 through the same cluster executor and confirm its actor sees H2, then resume derivation and assert that the captured submitted specification is still Exact(H). Place this optional barrier and submission observation hook internally in CoreGeneralSekibanExecutor, exposed only to test assemblies through InternalsVisibleTo, with no public callback or runtime dependency. The same-cluster reservation may reject before PG, so inspect the frozen specification separately rather than require that schedule to reach PG. The independent-writer schedule uses existing internal TagHeadProtocolHook to reach the PG conflict seam, then checks exact typed pairs and absence of command rows using an independent connection; preserve the G40 intentional repair-only audit exception. Mutants replacing evidence from either store or actor must fail their respective schedule. The actor-refresh schedule runs in Sekiban.Dcb.Orleans.Tests using the real Orleans+PostgreSQL TwoClusterPostgresProductTests fixture shape; InMemory cannot execute this derived-mode handler because capability preflight rejects it. Add Core InternalsVisibleTo for Sekiban.Dcb.Orleans.Tests and Sekiban.Dcb.Postgres.Tests for the internal derivation barrier/submission observation, and Sekiban.Dcb.WithResult.Tests when internal ledger/inventory tests require it. These are new test-access declarations, not assumed existing ones.

- AC3: **Ambiguous reads:** same logical tag at H/H succeeds; H/H2, empty/H, failed-read-plus-success on an emitted tag, malformed positive-version/empty cursor, and cross-service/tag identity reject before reserve/allocate/write. Different-projector H/H is the near passing case; H/H2 is the rejecting case. Different tag wrapper objects with identical logical identity must not evade this rule. Malformed positive-version/empty-cursor evidence is injected through a fake ITagStateActorCommon behind a test IActorObjectAccessor decorator. Label this as framework evidence-validation coverage, not PostgreSQL CAS proof; retain realPG schedules for actual state/cursor conflicts.

- AC4: **Missing evidence:** emitted-but-unread tag, existence-only read, head-only lookup and a forged mismatching ConsistencyTag token reject. Successful framework empty read emits AssertEmpty; synthetic actor-error empty never qualifies. Extra read-only tag C does not become an emitted expectation.

- AC5: **Fail-closed ordering:** actual unsupported InMemory/SQLite/Cosmos/Dynamo descriptors and missing PG epoch stop before handler, including a handler that would emit nothing; kill the mutant that skips epoch preflight for no-op. Mutually exclusive option modes stop before handler; no fallback unconditional append. Instrument counts and distinguish allowed preflight reads from mutation.

- AC6: **Surface parity:** WithResult/WithoutResult and both Orleans typed entry forms (delegate and self-handler) exercise success, no-op and typed conflict/validation/unsupported paths. Real Orleans+PostgreSQL proves normal DI reaches the capability and uses state returned by the same invocation. Existing explicit/unconditional/serialized contract matrices remain green.

- AC7: **Atomicity/compatibility and recovery:** preserve G40 multi-event/multi-tag rollback and published public signature inventory/consumer proof; preserve reservations/cleanup/IDs/publication and G19 existence behavior outside the new mode. Real PG tests must not rely solely on actor reservations as proof of store CAS. Using TwoClusterPostgresProductTests or an independent durable writer, hold cluster A at cached H, commit H2 independently, and force A's first derived attempt to PG: it returns the original H/H2 conflict with no command rows. After awaited invalidation, a caller-owned second attempt reads payload and cursor at H2 and succeeds against a quiescent store. Kill the mutant omitting invalidation: it reads H again and conflicts. Verify no handler reinvocation, replacement expectation, or success publication occurs during the first attempt. Assert StateReadRecovery == InvalidationCompleted only after all affected calls succeed; injected accessor/notification failure on one of multiple tags must preserve the original conflict and expose InvalidationIncomplete in Release, with no false retry-progress guarantee. Assert legacy construction/explicit paths remain NotAttempted. Kill the mutant reporting Completed despite a failed notification; a Debug.WriteLine assertion is not proof.

- AC8: **Docs:** EN/JA dynamic A/B/no-op example with caller-owned bounded retry and one handler invocation per attempt; explain between-attempt actor head invalidation (not an event write), deferred re-catch-up, the observable StateReadRecovery result and notification-failure/continued-contention limitations from A. Explicit opt-in, epoch/drain prerequisites, unsupported providers, emitted-tag-only protection, no exactly-once external-effects claim, and legacy APIs. Write back to this intent and `postgres-tag-head-cas.md` at closeout.

- AC9: For derived-mode conflict recovery, attempt actor invalidation exactly once for every distinct affected emitted consistency tag in the same service, even when an earlier accessor/notification attempt fails. Set InvalidationCompleted only when all succeed; otherwise preserve the original conflict object/pairs and expose InvalidationIncomplete. Kill an early-return-on-first-failure mutant and a false-Completed mutant, asserting all per-tag call counts in Release. No in-attempt handler retry, replacement head read or event publication. This carries Fable N9 from the accepted design review.

## Verification

Execute real PostgreSQL and Orleans+PostgreSQL discriminating schedules on supported net9/net10. Record exact commands, counts and frozen expected/observed positions; reservations alone are not store-CAS evidence. Preserve existing explicit/unconditional/serialized matrices, G19 existence and G40 rollback/repair-only-audit boundaries. Extend the existing expected-tag ABI inventory and applicable binary consumers, including source overload cases. Use controlled internal test barriers, no public callback. Successful CI and fresh exact-head intent-cli independent review are mandatory before individual merge/closeout. Run git diff --check.

## Related Links

- https://github.com/J-Tech-Japan/Sekiban/issues/1193
- intents/sekiban/design/command-decision-read-cas-1193.md
- intents/sekiban/design/postgres-tag-head-cas.md
- intents/sekiban/intent-tree/00-map.md

## Knowledge Maintenance

Update the accepted decision-read design, PostgreSQL CAS design, intent map and the four EN/JA documentation paths above in the same closeout. Record actual API, actor invalidation outcome, both mutation schedules and all-tag failure behavior. No separate ADR or diagram is required; the accepted design document is the decision record.

## Guide Reachability (G645)

No intent-cli role-facing workflow surface is introduced. Developer product API guidance belongs in the named EN/JA documentation.

## Base Branch Policy

Policy: direct-main
Expected PR base branch: main

Publish/promote an implementation issue and begin implementation only after G56/G57/G58 individual closeouts. G65 is the next serial slot ahead of deferred G62/G64/G63; numeric identifiers do not reorder this operator instruction.

tomohisa:
Closes #1197

## Summary

Adds the explicit `CommandExecutionOptions.DeriveExpectedTagPositionsFromStateReads` protocol. A single handler invocation records successful raw `GetStateAsync` evidence in a bounded ordinal ledger, derives only the emitted consistency-tag `AssertEmpty`/`Exact` expectations, and submits the existing expected-position write unchanged. Missing, failed, malformed, conflicting, or synthetic-empty evidence fails closed before reservation or write.

The executor preflights service identity, provider capability, and the service-scoped enablement epoch before the handler, including no-event attempts. The derived mode is mutually exclusive with explicit expected positions and conditional append. Durable derived conflicts preserve the original typed exception and perform one awaited invalidation attempt per affected tag; `StateReadRecovery` reports complete versus incomplete invalidation without retrying or reinvoking the handler. Both Orleans facades forward the additive options surface through `IConditionalCommandExecutor`.

## Provider and compatibility evidence

- `DerivedExpectedTagPositionTests` against Testcontainers PostgreSQL: 4 passed on net9.0 and 4 passed on net10.0.
  - successful PostgreSQL cursor is frozen after one handler/read;
  - an independent durable store refresh rejects the frozen H read and invalidates both affected tags;
  - injected failure on one tag still attempts both invalidations and reports `InvalidationIncomplete`;
  - a same-composition actor refresh cannot replace the already-frozen successful read.
- WithResult expected-position compatibility inventory: 3/3 passed on net9.0 and 3/3 on net10.0.
- WithoutResult expected-position compatibility inventory: 1/1 passed on net9.0 and 1/1 on net10.0.
- WithResult and WithoutResult Orleans projects build successfully on net9.0 and net10.0; both existing binary consumers run successfully against the published compatibility package.
- `git diff --check` is clean. Restore/build output contains existing package-vulnerability warnings only; no compile or test errors.

No schema, serialized protocol, release/version, unconditional behavior, explicit expected-position behavior, or automatic retry was changed.

### DIFF
--- a/dcb/src/Sekiban.Dcb.Core/Actors/CoreGeneralSekibanExecutor.cs
+++ b/dcb/src/Sekiban.Dcb.Core/Actors/CoreGeneralSekibanExecutor.cs
@@ -40,6 +40,15 @@ public class CoreGeneralSekibanExecutor
     internal Func<Guid> ConditionalEventIdFactory { get; set; } = Guid.CreateVersion7;
     internal Func<string>? ConditionalSortableIdFactory { get; set; }
 
+    /// <summary>
+    ///     Test-only barrier immediately before the attempt-local state-read evidence is frozen. The custom argument
+    ///     type keeps this seam out of the public API and out of the provider seam inventory.
+    /// </summary>
+    internal Func<DerivedExpectedTagPositionFreezeContext, Task>? BeforeDerivedExpectedTagPositionFreeze { get; set; }
+
+    /// <summary>Test-only observation of the exact specification submitted after the freeze barrier.</summary>
+    internal Action<ExpectedTagPositionSpecification>? DerivedExpectedTagPositionFrozen { get; set; }
+
     private const string DefaultExecutedUser = "GeneralSekibanExecutor";
     private const string SerializedExecutedUser = "SerializedSekibanExecutor";
     private readonly IExecutedUserProvider? _executedUserProvider;
@@ -139,7 +148,7 @@ public class CoreGeneralSekibanExecutor
         TCommand command,
         Func<TCommand, ICoreCommandContext, Task<ResultBox<EventOrNone>>> handlerFunc,
         CancellationToken cancellationToken = default) where TCommand : ICommand =>
-        ExecuteAsyncCore(command, handlerFunc, null, cancellationToken);
+        ExecuteAsyncCore(command, handlerFunc, null, false, cancellationToken);
 
     /// <summary>
     ///     Shared ordinary-batch pipeline. The legacy public call enters with <paramref name="expectedTagPositions" />
@@ -150,6 +159,7 @@ public class CoreGeneralSekibanExecutor
         TCommand command,
         Func<TCommand, ICoreCommandContext, Task<ResultBox<EventOrNone>>> handlerFunc,
         ExpectedTagPositionSpecification? expectedTagPositions,
+        bool deriveExpectedTagPositionsFromStateReads,
         CancellationToken cancellationToken) where TCommand : ICommand
     {
         var stopwatch = Stopwatch.StartNew();
@@ -166,10 +176,11 @@ public class CoreGeneralSekibanExecutor
             // Expected-position is optional, but never best-effort. Validate its discriminated shape and the live
             // descriptor before the handler can allocate ids, reserve a tag, or reach any provider write method.
             IExpectedTagPositionEventStore? expectedPositionStore = null;
-            if (expectedTagPositions is not null)
+            string? currentServiceId = null;
+            if (expectedTagPositions is not null || deriveExpectedTagPositionsFromStateReads)
             {
-                var currentServiceId = ServiceIdValidator.NormalizeAndValidate(_serviceIdProvider.GetCurrentServiceId());
-                expectedTagPositions.ValidateEntryShapes(currentServiceId);
+                currentServiceId = ServiceIdValidator.NormalizeAndValidate(_serviceIdProvider.GetCurrentServiceId());
+                expectedTagPositions?.ValidateEntryShapes(currentServiceId);
 
                 var capability = Sekiban.Dcb.Capabilities.SekibanDcbCapabilityResolver.DescribeWriteConditions(
                     _eventStore, "event store");
@@ -183,7 +194,7 @@ public class CoreGeneralSekibanExecutor
                 }
 
                 expectedPositionStore = resolvedExpectedPositionStore;
-                if (expectedTagPositions.RequiresEnforcement)
+                if (deriveExpectedTagPositionsFromStateReads || expectedTagPositions?.RequiresEnforcement == true)
                 {
                     var enabled = await expectedPositionStore
                         .EnsureExpectedTagPositionEnforcementEnabledAsync(cancellationToken);
@@ -195,7 +206,9 @@ public class CoreGeneralSekibanExecutor
             }
 
             // Step 1: Create command context
-            var commandContext = new CoreCommandContext(_actorAccessor, _domainTypes);
+            var commandContext = currentServiceId is null
+                ? new CoreCommandContext(_actorAccessor, _domainTypes)
+                : new CoreCommandContext(_actorAccessor, _domainTypes, currentServiceId);
 
             // Step 2: Execute handler function with context
             var handlerResult = await handlerFunc(command, commandContext);
@@ -250,10 +263,25 @@ public class CoreGeneralSekibanExecutor
             // Step 3.1: Validate all tags
             TagValidator.ValidateTagsAndThrow(allTags);
 
+            if (deriveExpectedTagPositionsFromStateReads)
+            {
+                if (BeforeDerivedExpectedTagPositionFreeze is not null)
+                {
+                    await BeforeDerivedExpectedTagPositionFreeze(
+                        new DerivedExpectedTagPositionFreezeContext(commandContext, collectedEvents));
+                }
+
+                expectedTagPositions = DeriveExpectedTagPositions(
+                    commandContext,
+                    allTags,
+                    currentServiceId ?? throw new InvalidOperationException("Derived mode did not resolve a service id."));
+                DerivedExpectedTagPositionFrozen?.Invoke(expectedTagPositions);
+            }
+
             if (expectedTagPositions is not null)
             {
                 expectedTagPositions.ValidateFor(
-                    ServiceIdValidator.NormalizeAndValidate(_serviceIdProvider.GetCurrentServiceId()),
+                    currentServiceId ?? ServiceIdValidator.NormalizeAndValidate(_serviceIdProvider.GetCurrentServiceId()),
                     allTags.Where(tag => tag.IsConsistencyTag()).Select(tag => tag.GetTag()));
             }
 
@@ -355,7 +383,14 @@ public class CoreGeneralSekibanExecutor
                     if (!writeResult.IsSuccess)
                     {
                         await TagReservationHelper.CancelReservationsAsync(_actorAccessor, reservations);
-                        return ResultBox.Error<ExecutionResult>(writeResult.GetException());
+                        var writeException = writeResult.GetException();
+                        if (deriveExpectedTagPositionsFromStateReads &&
+                            writeException is ExpectedTagPositionConflictException conflict)
+                        {
+                            await RecoverDerivedStateReadConflictAsync(conflict, currentServiceId!);
+                        }
+
+                        return ResultBox.Error<ExecutionResult>(writeException);
                     }
 
                     // The expected-position store receives a lossless serialization of the exact Event instances above;
@@ -408,10 +443,18 @@ public class CoreGeneralSekibanExecutor
                         },
                         firstEvent.SortableUniqueIdValue));
             }
-            catch (Exception)
+            catch (Exception ex)
             {
                 // If anything fails after reservations, cancel them
                 await TagReservationHelper.CancelReservationsAsync(_actorAccessor, reservations);
+
+                if (deriveExpectedTagPositionsFromStateReads &&
+                    ex is ExpectedTagPositionConflictException conflict &&
+                    conflict.StateReadRecovery == StateReadRecoveryStatus.NotAttempted)
+                {
+                    await RecoverDerivedStateReadConflictAsync(conflict, currentServiceId!);
+                }
+
                 throw;
             }
         }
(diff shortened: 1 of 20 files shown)


tomohisa:
Withdrawing this optional feature for now at the operator request. The source request #1193 was withdrawn, and we will prioritize the existing bug fixes. Branch and implementation commits are retained for possible future evaluation; this is not merged or released.

Issue #1185, opened 2026-09-05
DCB Orleans publisher: support large-event delivery and handle permanent message-size rejection


joemphilips:
## Summary

We need a supported DCB event-delivery policy when a persisted event exceeds the configured Orleans stream backend's message limit. In a local PostgreSQL + Orleans Azure Queue integration run, event publication repeatedly received HTTP 413 with `MaxLimit` 65536. `OrleansEventPublisher` retries the same full event indefinitely.

This is a capability and failure-handling report. The Azure Queue size limit itself is not a Sekiban defect. We would like guidance on the intended library-level solution before adding an application-specific event publisher or changing event semantics to fit a transport.

## Versions and source

- Observed with Sekiban DCB 10.17.0, package source commit `cca3c8d64c5a5a013b8ce4e5197c160b26423864`, using local Azurite and PostgreSQL.
- The same publisher implementation is present at the inspected 10.20.0 commit `b7f2386c41f90712ac65633efaa0f2b03b64c882`. We have not repeated this run with 10.20.0 binaries.
- [10.17 OrleansEventPublisher](https://github.com/J-Tech-Japan/Sekiban/blob/cca3c8d64c5a5a013b8ce4e5197c160b26423864/dcb/src/Sekiban.Dcb.Orleans.Core/OrleansEventPublisher.cs)
- [Inspected 10.20 OrleansEventPublisher](https://github.com/J-Tech-Japan/Sekiban/blob/b7f2386c41f90712ac65633efaa0f2b03b64c882/dcb/src/Sekiban.Dcb.Orleans.Core/OrleansEventPublisher.cs)

## Observed behavior

The integration archive contains 66 failed queue sends with HTTP 413. The backend response reports a 65,536-byte limit. The application uses `AddAzureQueueStreams` for the DCB event stream and the standard `OrleansEventPublisher`.

The affected workload emits immutable events with large collections. Their storage representation is valid for the event store, but full-event stream delivery exceeds the selected transport limit. We have not measured a general maximum safe event payload; serialization and transport encoding overhead also matter.

## Source-level behavior

`PublishAsync` converts each event to `SerializableEvent`, enqueues it in an unbounded `Channel`, and returns. `ProcessQueueAsync` calls `stream.OnNextAsync(item.Event)` with the full event.

[...]

PR #1208, opened 2026-09-10
SEK-G67 Add DynamoDB event-item size gate

### LINKED ISSUE #1207: SEK-G67 Add opt-in DynamoDB event-item size measurement before persistence
## Goal

Provide an opt-in DynamoDB base event-item measurement that the existing Executor gate uses to reject oversized events before persistence. This is an event-item guarantee, not a complete DynamoDB-write or Queue-delivery guarantee.

## Why This Slice Exists Now

G66 supplied the gate but no provider item encoder. DynamoDB has a concrete 409600-byte item limit and constructs AttributeValue maps before writes, making a truthful provider capability feasible. PostgreSQL physical-row estimates and unobservable Azure adapter envelopes are not suitable substitutes.

## Current Observed State

Typed, serialized and conditional provider paths separately construct DynamoEvent and call ToAttributeValues. They vary in payload conversion, and the mapper injects UtcNow; service/options determine keys. Event items emit strings and a list of strings, with optional metadata. Separate tag items and transaction/batch grouping have separate constraints.

G66 landed as f180a56a. Its existing IExecutorSizeMeasurement and typed strict/non-strict capability contract is the integration boundary. No required interface change is needed. Read current main before coding. Prior isolated DynamoDB Local2.6.1 accepted exact409600 and refused409601 using generated maps; this is emulator characterization, not live AWS proof.

## Target Repo / Path / Part

Repository: J-Tech-Japan/Sekiban.

Paths: dcb/src/Sekiban.Dcb.DynamoDB, affected dcb/tests, docs/dcb_llm and docs/dcb_llm_ja, required CI wiring under .github/workflows.

Part: provider mapping/measurement registration and test oracle.

## In Scope

Shared provider event mapper, accurate item admission count, additive opt-in registration, exact write-path parity, meaningful executor boundary tests, isolated emulator verification and EN/JA docs.

## Out Of Scope

Complete provider-request admission, separate tag-item gate, transaction/batch byte splitting, key-length validation, other provider measurement, Azure Queue wire envelope, permanent retry/recovery changes, new transport, schema/data migration, release/version bump,1169 and withdrawn1193.

Implement the eight acceptance criteria below using the published issue alone. Count the actual DynamoDB base event item through its shared provider mapper and reject an oversized whole executor operation before any durable write. Preserve legacy unconfigured behavior and provider error identity. Clearly retain tag/request/queue limitations. Child workers must not read or mutate parent host metadata.

## Acceptance Criteria

- AC1: Add a provider-owned opt-in IExecutorSizeMeasurement implementation and registration for a specifically named DynamoDB event-item StorageItem scope. Reuse G66 contracts, with no new Core capability interfaces. Count base-item admission bytes under documented AWS rules, not JSON request framing, Base64 wire bytes, billing rounding, GSI storage copies or the 100-byte storage overhead. Enforce the 409600-byte service ceiling in the provider-owned registration entry point: reject a configured quota above 409600 and default to 409600 when none is given. A smaller application quota may be selected. A caller that hand-constructs ExecutorSizePolicy bypasses that entry point; such a policy is the caller-owned quota and carries no service-ceiling claim. Legacy unregistered behavior stays unchanged.

- AC2: Extract or reuse one provider-owned canonical event mapper across typed WriteEventsAsync, WriteSerializableEventsAsync and conditional TryWriteConditionalClaimAsync. Measurement must use the same serializer/payload conversion, service identity, WriteShardCount/GSI keys, event identity, optional metadata and tag list as the writer. Serialized/conditional mapping reads supplied payload bytes without introducing an additional domain binding requirement. Preserve G66 capability and exact provider in-doubt semantics.

- AC3: Return an exact count of emitted S attributes and L-of-S tags under the AWS formula: UTF8 attribute names/values and list/element overhead. Unsupported emitted types or unprovable service/mapper parity yield Unavailable, never a guessed Exact. Timestamp parity is size parity: share timestamp preparation where feasible or explicitly prove the fixed-width UTC O-format count across all writer paths; normalize only the timestamp value in a map test while independently asserting its format and byte length. Do not normalize keys, payload or other variable fields.

- AC4: Reuse G66 pre-write evaluation across all five Executor entrances. Oversized last event rejects the entire operation before durable event/tag mutations or publisher enqueue; preserve reservation cleanup on the three reservation-bearing entries and none on conditional entries. A configured strict unavailable capability rejects before writes; non-strict emits named unvalidated diagnostics. Measurements for other providers or Queue envelopes must not silently become available. This capability covers base event items only, not separate tag-table items or actual provider transaction/batch totals.

- AC5: Independent expected-count and actual outgoing AttributeValue-map tests must cover typed, serialized and conditional provider paths; UTF8 payload/tags/names, metadata absent/present, service identity and shard options, timestamp contribution and list overhead. Exercise item sizes409599/409600/409601: at or under passes the size gate, above yields typed size failure before SDK dispatch. Last-event oversized proves zero writes/enqueues. Include an executable logical-payload-count mutant or equivalent contrasting implementation that fails the item-boundary oracle.

- AC6: Add reproducible isolated pinned DynamoDB Local integration using synthetic credentials/data; compare production-generated item maps with the independent byte oracle and demonstrate emulator acceptance at409600 and refusal above. Label this emulator evidence, not live AWS proof. Tests must also observe zero SDK write dispatch for gate rejection; a post-write service error is not a passing gate test. Existing recording-client tests alone are not emulator proof. No production account is required. Integrate the meaningful verification into the existing required CI, without skipping/suppressing or weakening assertions. The emulator integration must exercise the real AmazonDynamoDBClient against the pinned local endpoint; an AWS CLI or raw item-file substitute is characterization evidence only and does not satisfy this criterion. The local CookieContainer/GetDomainName runtime failure is not grounds for skipping or replacing CI integration; if it reproduces on CI, report a red gate without waiver.

- AC7: Preserve existing public DynamoDB and Executor/facade constructor and DI signatures. New registration is additive and service/options propagation is tested; use focused reflection inventory for changed public surfaces and existing legacy tests. Run relevant DynamoDB, serialized/conditional and size-gate tests on supported net9/net10 plus actual required CI. No schema/data migration, package/provider upgrade or version bump. New test-only container packages are permitted; existing workflow inclusion may satisfy CI wiring without a workflow edit.

- AC8: Update EN/JA storage and command docs with opt-in configuration, actual attribute-byte accounting, hard ceiling versus smaller policy quota, capability errors and examples. Explicitly exclude tag items, key-length limits, transaction aggregate4MB, BatchWriteItem request size, GSI storage footprint, other providers and Azure Queue envelope from this event-item guarantee. Explain MaxBytesPerOperation sums measured copies; Destination fan-out counts each destination copy, unlike a database footprint or queue batch guarantee. Source1185 remains open for remaining provider-operation and permanent-failure/recovery work; no release, new transport or splitting. Document the hand-constructed policy bypass of provider registration: it has no provider service-ceiling claim. A passing event item does not establish that a separately mapped tag item near the limit fits.

## Verification

Use independent attribute-count and captured-production-map oracles plus an isolated pinned DynamoDB Local boundary test. Test all provider mapping paths and Executor rejection entrances. Preserve no-write/no-enqueue evidence and add a logical-payload mutant discriminator. Run relevant net9/net10 tests, actual required CI and git diff --check. No mock or emulator result is represented as live AWS proof.

## Related Links

- https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Constraints.html
- https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/CapacityUnitCalculations.html

## Knowledge Maintenance

EN/JA command and storage docs required. Host Architect/Orchestrator record landed scope, exact-head review, CI and residual1185 work after individual merge. ADR/diagram declined because this integrates the existing capability; closeout writeback required.

## Guide Reachability (G645)

No intent-cli role-facing surface added; no_role_facing_surface=true. Product documentation is required above.

## Base Branch Policy

Policy: direct-main. Expected PR base branch: main. Independent exact-head intent-cli review and actual green CI precede individual merge and full closeout. Source1185 must not be closed as fully fixed by this slice.

tomohisa:
Closes #1207\n\n## Scope\n- Adds opt-in DynamoDB base event-item measurement through the existing executor size gate.\n- Uses one provider-owned mapper for typed, serialized, and conditional event writes.\n- Counts UTF-8 attribute names/values, binary lengths, and DynamoDB list/map overhead; the provider registration enforces the 409600-byte event-item ceiling.\n- Preserves legacy unregistered writes and excludes tag items, GSI projection/billing overhead, request envelopes, transaction/BatchWrite limits, and other transports.\n\n## Proof\n- Independent captured-map oracle and typed/serialized/conditional parity tests.\n- Boundary tests for 409599/409600/409601 and executor no-write rejection.\n- Real AmazonDynamoDBClient against pinned DynamoDB Local in both CI TFM jobs; the integration proof accepts exact 400 KiB, observes Local rejection above it, and proves gate rejection leaves the event table unchanged.\n- EN/JA storage-provider documentation and pinned CI service setup are included.

### DIFF
--- a/.github/workflows/run_test_dcb.yml
+++ b/.github/workflows/run_test_dcb.yml
@@ -48,6 +48,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       SEKIBAN_TAG_STREAM_CONTROLLED_CEILING: '1'
+      SEKIBAN_DYNAMODB_LOCAL_ENDPOINT: http://127.0.0.1:18000
     concurrency:
       group: dcb-tests-net9-${{ github.ref }}
       cancel-in-progress: false
@@ -80,6 +81,21 @@ jobs:
         run: |
           dotnet build -c Release -p:GeneratePackageOnBuild=false --verbosity minimal -m:1
 
+      - name: Start pinned DynamoDB Local for G67
+        run: |
+          docker run --rm -d --name sekiban-g67-dynamodb-local -p 18000:8000 \
+            amazon/dynamodb-local:2.6.1@sha256:1856c05cc66a0e49dc1099e483ad2851477eeebe2135250ac11a1d1227db54b1 \
+            -jar DynamoDBLocal.jar -inMemory -sharedDb
+          for attempt in $(seq 1 30); do
+            http_code="$(curl --silent --output /dev/null --write-out '%{http_code}' http://127.0.0.1:18000/ || true)"
+            if [ "$http_code" != "000" ]; then
+              exit 0
+            fi
+            sleep 1
+          done
+          docker logs sekiban-g67-dynamodb-local
+          exit 1
+
       - name: dotnet test DCB projects sequentially (net9.0)
         run: |
           dotnet test dcb/tests/Sekiban.Dcb.Orleans.Tests/Sekiban.Dcb.Orleans.Tests.csproj -c Release --no-build -f net9.0 --verbosity normal --collect:"XPlat Code Coverage" -- RunConfiguration.MaxCpuCount=1
@@ -90,6 +106,10 @@ jobs:
           dotnet test dcb/tests/Sekiban.Dcb.MaterializedView.Postgres.Tests/Sekiban.Dcb.MaterializedView.Postgres.Tests.csproj -c Release --no-build -f net9.0 --verbosity normal --collect:"XPlat Code Coverage" -- RunConfiguration.MaxCpuCount=1
           dotnet test dcb/tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests.csproj -c Release --no-build -f net9.0 --verbosity normal --collect:"XPlat Code Coverage" -- RunConfiguration.MaxCpuCount=1
           dotnet test dcb/tests/Sekiban.Dcb.BlobStorage.AzureStorage.Unit/Sekiban.Dcb.BlobStorage.AzureStorage.Unit.csproj -c Release --no-build -f net9.0 --verbosity normal --collect:"XPlat Code Coverage" -- RunConfiguration.MaxCpuCount=1
+
+      - name: Stop pinned DynamoDB Local
+        if: always()
+        run: docker rm -f sekiban-g67-dynamodb-local || true
 
       - name: Verify 10.14.1 materialized-view binary consumer without rebuild
         run: |
@@ -106,6 +126,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       SEKIBAN_TAG_STREAM_CONTROLLED_CEILING: '1'
+      SEKIBAN_DYNAMODB_LOCAL_ENDPOINT: http://127.0.0.1:18000
     concurrency:
       group: dcb-tests-net10-${{ github.ref }}
       cancel-in-progress: false
@@ -138,6 +159,21 @@ jobs:
         run: |
           dotnet build -c Release -p:GeneratePackageOnBuild=false --verbosity minimal -m:1
 
+      - name: Start pinned DynamoDB Local for G67
+        run: |
+          docker run --rm -d --name sekiban-g67-dynamodb-local -p 18000:8000 \
+            amazon/dynamodb-local:2.6.1@sha256:1856c05cc66a0e49dc1099e483ad2851477eeebe2135250ac11a1d1227db54b1 \
+            -jar DynamoDBLocal.jar -inMemory -sharedDb
+          for attempt in $(seq 1 30); do
+            http_code="$(curl --silent --output /dev/null --write-out '%{http_code}' http://127.0.0.1:18000/ || true)"
+            if [ "$http_code" != "000" ]; then
+              exit 0
+            fi
+            sleep 1
+          done
+          docker logs sekiban-g67-dynamodb-local
+          exit 1
+
       - name: dotnet test DCB projects sequentially (net10.0)
         run: |
           dotnet test dcb/tests/Sekiban.Dcb.Orleans.Tests/Sekiban.Dcb.Orleans.Tests.csproj -c Release --no-build -f net10.0 --verbosity normal --collect:"XPlat Code Coverage" -- RunConfiguration.MaxCpuCount=1
@@ -148,3 +184,7 @@ jobs:
           dotnet test dcb/tests/Sekiban.Dcb.MaterializedView.Postgres.Tests/Sekiban.Dcb.MaterializedView.Postgres.Tests.csproj -c Release --no-build -f net10.0 --verbosity normal --collect:"XPlat Code Coverage" -- RunConfiguration.MaxCpuCount=1
           dotnet test dcb/tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests/Sekiban.Dcb.MaterializedView.MultiProvider.Tests.csproj -c Release --no-build -f net10.0 --verbosity normal --collect:"XPlat Code Coverage" -- RunConfiguration.MaxCpuCount=1
           dotnet test dcb/tests/Sekiban.Dcb.BlobStorage.AzureStorage.Unit/Sekiban.Dcb.BlobStorage.AzureStorage.Unit.csproj -c Release --no-build -f net10.0 --verbosity normal --collect:"XPlat Code Coverage" -- RunConfiguration.MaxCpuCount=1
+
+      - name: Stop pinned DynamoDB Local
+        if: always()
+        run: docker rm -f sekiban-g67-dynamodb-local || true
--- a/dcb/src/Sekiban.Dcb.DynamoDB/DynamoDbEventItemMapper.cs
+++ b/dcb/src/Sekiban.Dcb.DynamoDB/DynamoDbEventItemMapper.cs
@@ -1,0 +1,105 @@
+using System.Security.Cryptography;
+using System.Text;
+using Sekiban.Dcb.Events;
+using Sekiban.Dcb.DynamoDB.Models;
+
+namespace Sekiban.Dcb.DynamoDB;
+
+/// <summary>
+/// The one provider-owned event-item mapper used by typed, serialized, and conditional writes.
+/// Keeping the mapper at the write boundary makes the measured item the item that the store emits.
+/// </summary>
+internal static class DynamoDbEventItemMapper
+{
+    public static DynamoDbEventWriteItem FromEvent(
+        Event @event,
+        string serializedPayload,
+        string serviceId,
+        int writeShardCount,
+        DateTimeOffset timestamp)
+    {
+        ArgumentNullException.ThrowIfNull(@event);
+        ArgumentNullException.ThrowIfNull(serializedPayload);
+        ArgumentNullException.ThrowIfNull(serviceId);
+
+        var timestampText = timestamp.ToString("O");
+        var dynamoEvent = DynamoEvent.FromEvent(
+            @event,
+            serializedPayload,
+            GetGsiPartitionKey(@event.SortableUniqueIdValue, serviceId, writeShardCount),
+            serviceId);
+        dynamoEvent.Timestamp = timestampText;
+
+        var tags = @event.Tags.Select(tagString =>
+        {
+            var tagGroup = tagString.Contains(':', StringComparison.Ordinal)
+                ? tagString.Split(':')[0]
+                : tagString;
+            var tag = DynamoTag.FromEventTag(
+                serviceId,
+                tagString,
+                BuildStoredTagGroup(serviceId, tagGroup),
+                @event.SortableUniqueIdValue,
+                @event.Id,
+                @event.EventType);
+            tag.CreatedAt = timestampText;
+            return tag;
+        }).ToList();
+
+        return new DynamoDbEventWriteItem(@event, dynamoEvent, tags);
+    }
+
+    public static DynamoDbEventWriteItem FromSerializableEvent(
+        SerializableEvent serializedEvent,
+        string serviceId,
+        int writeShardCount,
+        Guid? eventId = null,
+        DateTimeOffset? timestamp = null)
+    {
+        ArgumentNullException.ThrowIfNull(serializedEvent);
+
+        var helperEvent = new Event(
+            PlaceholderPayload.Instance,
+            serializedEvent.SortableUniqueIdValue,
+            serializedEvent.EventPayloadName,
+            eventId ?? serializedEvent.Id,
+            serializedEvent.EventMetadata,
+            serializedEvent.Tags);
+
+        return FromEvent(
+            helperEvent,
+            Encoding.UTF8.GetString(serializedEvent.Payload),
+            serviceId,
+            writeShardCount,
+            timestamp ?? DateTimeOffset.UtcNow);
+    }
+
+    private static string GetGsiPartitionKey(string sortableUniqueId, string serviceId, int writeShardCount)
+    {
+        if (writeShardCount <= 1)
+            return BuildEventsGsiPartitionKey(serviceId);
+
+        var hash = SHA256.HashData(Encoding.UTF8.GetBytes(sortableUniqueId));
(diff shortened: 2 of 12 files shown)


tomohisa (commented):
Independent exact-head review at `5f756512cd044c199e5f35f9d11113349328c2cf` against standalone issue #1207. Read-only; no source, label or merge action taken.

**Verdict: request-update.** No functional defect found — the mechanism is correct and the guarantee is honestly bounded. Five explicit AC clauses are not yet delivered.

## CI (settled, inspected in the job logs)

`dcbTestsNet9` (job 102692756999) and `dcbTestsNet10` (job 102692757233) both **success**; PostgreSQL packaged-consumer and template validation also green. The pinned emulator (`amazon/dynamodb-local@sha256:1856c05c…`) pulled and ran on both TFMs and was cleaned up. All eight new tests ran and passed on net9.0 and net10.0 with no skips, including `DynamoDbLocalEventItemTests` at ~1 s — so the real `AmazonDynamoDBClient` genuinely reached the emulator on Linux CI, confirming that the recorded `CookieContainer`/`GetDomainName` failure was a developer-host limitation only.

## Verified correct

- **Mapper extraction is real (AC2).** All three routes now build through `DynamoDbEventItemMapper`; 97 lines of triplicated mapping removed. I compared the removed and new tag mapping field-by-field — `pk`, `sk`, stored tag group, `eventType` and the conditional `deterministicId` are all preserved. Event and tag rows now share one captured timestamp instead of two independent `UtcNow` calls (same `"O"` format, size-neutral).
- **Serializer/identity parity is exact, verified at source.** `ToSerializableEvent` sets `Payload = UTF8(SerializeEventPayload(...))` and `EventPayloadName = EventType`, and the typed executor route builds prepared events with exactly that, so the measured `payload`/`eventType` bytes are identical to what the writer emits. Two subtleties also check out: `ServiceIdValidator` lowercases but the `^[a-z0-9-]{1,64}$` pattern keeps it ASCII, so byte count is unchanged; and the conditional route's `deterministicId` differs in value from `serializable.Id` but GUID and key widths are fixed, so size parity holds.
- **Shard derivation is byte-identical** to the store's `ComputeShardIndex`, so `gsi1pk` is unchanged from `f180a56a` — no data-layout regression.
- **Ceiling/default/bypass (AC1)** implemented and tested; the DI registration resolves `IOptions<DynamoDbEventStoreOptions>` through a factory, so it is registration-order independent, and all three `AddSekibanDcbDynamoDb*` paths do register it.
- **Byte accounting (AC3)** matches AWS: UTF-8 names/values, raw `B` length, `3 + Σ(1 + element)` for `L`/`M`, excluding the 100-byte storage overhead, billing rounding, GSI copies and request framing.
- **The emulator proof is the strong kind (AC6):** the test reads the actual stored item back with a consistent `GetItemAsync`, re-sizes the payload to exactly 409600, and puts that *production-shaped* item — accepted; one byte more is refused by the service. That is production-map parity against a real implementation, not a self-consistency check.
- Scope is clean: nine files, all inside `target_path`, no Core change, no existing public signature touched, no migration or version bump.

## Findings

**F1 (AC5) — the mutant is missing.** AC5 requires "an executable logical-payload-count mutant or equivalent contrasting implementation that fails the item-boundary oracle"; `review-context.md` names the same discriminator. Nothing in either test file does this. `MeasurementMatchesAnIndependentUtf8AndListOracle…` compares two counts over the *same captured map*, which proves the calculator, not that a wrong representation would be caught. This is the clause that makes the capability demonstrably non-ceremonial — the reason DynamoDB was chosen over PostgreSQL. Suggested: for an event whose mapped item is exactly 409600, assert a payload-only count is *not* equal, and/or that a gate built on it admits an item the DynamoDB oracle rejects.

**F2 (AC4) — strict/non-strict capability behaviour is untested.** The `Unavailable` paths exist (non-`StorageItem` representation; the mapping-failure catch) but nothing exercises them: no test that a strict policy throws `ExecutorSizeCapabilityException` before any write, none that a non-strict policy records the named diagnostic and proceeds. The representation guard is what prevents this capability from silently answering for a Queue/Destination policy, and it has zero coverage.

**F3 (AC5/AC6) — three related gaps.**
1. Both rejection tests commit a **single** event. Despite the name `…RejectsAnOversizedLastEventBeforeTheStore`, there is no earlier event, so the all-or-nothing property (a good first event is not written when a later one is oversized) is unproven — the case that distinguishes a whole-operation gate from a per-event one.
2. The 409599/409600/409601 theory asserts measurement equality only; no test runs the **gate** at 409600 (pass) and 409601 (typed failure), which is how AC5 words it.
3. AC6 asks tests to "observe zero SDK write dispatch"; the local test asserts an unchanged table count and the unit test an empty in-memory store — both effect assertions. The unit test file already has a capturing `IAmazonDynamoDB` (`capture.EventItems`), so asserting it recorded nothing after a gate rejection would satisfy this directly.

**F4 (AC8) — command docs missing.** AC8 says "storage **and command** docs" and the packet declares four doc targets, but only the two `11_storage_providers.md` files changed; both `03_aggregate_command_events.md` (EN and JA) are untouched — and the command doc is where a reader first meets a typed gate rejection. Also absent: any configuration example or exception names (`ExecutorSizeLimitExceededException`, `ExecutorSizeCapabilityException`), and AC8's Destination fan-out sentence ("counts each destination copy, unlike a database footprint or queue batch guarantee") — the docs cover the "sums measured items" half but not the contrast that stops a reader treating the operation quota as a storage footprint. The EN and JA sections that were added are faithful to each other and accurate.

**F5 (AC7) — no reflection inventory.** `DynamoDbEventItemSizeMeasurementTests.cs:1` imports `using System.Reflection;` but nothing in the file uses reflection: there is no inventory of the new public surface (`DynamoDbEventItemSizeMeasurement`, `DynamoDbExecutorSizeGateExtensions`) and no freeze over existing DynamoDB signatures. `dcb/tests/Sekiban.Dcb.TestSupport` has the precedent. Relatedly, no test covers the DI path resolving `IOptions<DynamoDbEventStoreOptions>`, nor a `WriteShardCount > 1` configuration where the `gsi1pk` suffix changes the measured byte count — the one setup where a propagation bug would silently undercount.

**N1 (non-blocking, AC2)** — `DynamoDbEventStore.GetGsiPartitionKey(string, string)` and its `ComputeShardIndex` now have zero call sites, duplicating the mapper's private derivation. They are byte-identical today so there is no risk now, but AC2 asks for one canonical mapper precisely to prevent drift. `BuildEventsGsiPartitionKey` and `BuildStoredTagGroup` are still live on read paths and must stay.

With F1–F5 closed I expect to approve at the next exact head. Full artifact: `sek-g67-pr1208-5f756512-exact-opus5-review-20260910.md`.

tomohisa (commented):
Independent exact-head re-review at `0f70b7754d971914e2c879b016a4c4a342727e6e` against issue #1207 and the prior `5f756512` review. Read-only; no source, label, merge or release action taken. This verdict is for `0f70b775` only and is not transferred from the earlier head.

**Verdict: APPROVE.** F1–F5 are all closed with executable evidence, no prohibited production scope was touched, and CI is settled green on both TFMs with every new test verified to have actually run.

## Delta — evidence only

Six files, `467 insertions(+), 5 deletions(-)`: two test files and the four declared doc targets. `DynamoDbEventItemMapper.cs`, `DynamoDbEventItemSizeMeasurement.cs`, `DynamoDbExecutorSizeGateExtensions.cs`, `DynamoDbEventStore.cs` and `run_test_dcb.yml` are byte-identical to `5f756512`, so the mapper extraction, serializer/identity parity, ceiling enforcement, AWS byte formula and pinned-emulator CI wiring I verified there carry forward unchanged. **N1 was correctly left alone** — it was explicitly non-blocking and a production refactor would have been out of scope for an evidence repair.

## Closures

**F1 — closed.** `LogicalPayloadOnlyMutant_FailsTheMappedItemBoundaryOracle` builds an event whose mapped item is exactly 409601 and asserts the mapped oracle rejects it while the payload-only count stays under 409600 and strictly below the mapped count. Because the mutant value derives from the same event, the test genuinely fails if the measurement is ever switched to logical payload bytes. AC5 permits "mutant **or equivalent contrasting implementation**", and this proves the substantive claim — the capability is now demonstrably non-ceremonial.

**F2 — closed.** `NonStorageRepresentation_IsUnavailable` covers **both** `LogicalSerializedEventUtf8` and `Destination` (the guard that stops this capability silently answering for a Queue/Destination policy). `StrictUnavailableDynamoDbCapability_RejectsBeforeAnyStoreWrite` asserts `ExecutorSizeCapabilityException` with scope/representation/reason **and** an empty store; `NonStrictUnavailable…` asserts success, a single named diagnostic, and that the event was persisted.

**F3 — closed.** `ProviderMeasurementRejectsOversizedLastEventWithoutAnyDynamoDbWriteDispatch` drives a **real `DynamoDbEventStore`** through the recording proxy with a genuine two-event operation (small first, oversized last) and asserts `EventIndex == 1`, `WriteDispatches == 0`, and empty `EventItems`/`TagItems`. `WriteDispatches` increments in both the `TransactWriteItemsAsync` and `BatchWriteItemAsync` branches — the store's only dispatch paths — so this is a true zero-dispatch observation, not the effect assertion I flagged, and the absent good first event is the all-or-nothing proof that was missing. `ExecutorGate_UsesTheDynamoDbCeilingAtExactBoundary` pins 409600 pass / 409601 typed failure with `MeasuredBytes == 409601` and nothing persisted. One honest note: it drives the evaluator through a `FixedMeasurement` stub, so the end-to-end boundary claim is **composed** — measurement exactness at those sizes is proven separately by `MeasurementMatchesTheDynamoDbItemBoundary` and by the emulator test. I accept that; constructing 400 KB payloads inside an evaluator test would add no information.

**F4 — closed.** All four declared doc targets updated. Both `03_aggregate_command_events.md` files gained a SEK-G67 section — the command-side doc is where a reader first meets a typed gate rejection. The configuration example (`AddSekibanDcbDynamoDb(..., WriteShardCount = 2)` + `AddSekibanDcbDynamoDbEventItemSizeGate()`), both exception names, and the Destination fan-out contrast now appear in all four files in EN and JA, with the existing exclusion list retained.

**F5 — closed.** The stray `using System.Reflection;` is gone. `PublicDynamoDbSizeGateAndExistingProviderSurfacesRemainFrozen` pins by exact parameter-type sequence both the additive surface and the existing `DynamoDbEventStore` / `DynamoDbEventStoreFactory` / `SekibanDcbDynamoDbExtensions` signatures, and asserts `MaximumItemBytes` is a literal `const` with the expected value. `DiDynamoOptionsFlowIntoTheGateMeasurement_AndWriteShardsChangeMeasuredBytes` resolves the gate from a real container after `WriteShardCount = 2` and asserts the measurement reports **exactly two bytes more** — the `#0`/`#1` suffix on `gsi1pk`. That is precisely the configuration where a propagation failure would have silently undercounted.

## CI at this head (settled, job logs inspected)

`dcbTestsNet9` (job 102699456010) and `dcbTestsNet10` (job 102699455748) both **success**; PostgreSQL packaged-consumer and template validation green. **All 18 DynamoDB size-gate tests ran and passed on both net9.0 and net10.0**, zero failures, zero skips among them — including `DynamoDbLocalEventItemTests` against the pinned emulator (888 ms net9 / 806 ms net10, so the real `AmazonDynamoDBClient` again reached DynamoDB Local over HTTP). The only skip in either run is the pre-existing, unrelated `SimpleOrleansCommandQueryTests.Should_Update_Aggregate_Multiple_Times`. This discharges the repair report's own caveat that the emulator test could not be rerun locally.

## Scope

Nine files versus `main`, all inside `target_path`; no Core change, no new Core interface, no schema/data migration, no package upgrade, no version bump, no release. The guarantee stays correctly bounded — tag items, key length, GSI/billing, transaction and BatchWrite envelopes, the 4 MiB aggregate, other providers and the Azure Queue envelope remain explicitly excluded, and #1185 is left open for the remaining provider-operation and permanent-failure/recovery work.

N1 (dead duplicated `GetGsiPartitionKey`/`ComputeShardIndex` in the store) remains open by design — a small optional tidy-up for a future packet, not a merge condition.

Full artifact: `sek-g67-pr1208-0f70b775-exact-opus5-rereview-20260910.md`.

PR #1210, opened 2026-09-10
SEK-G68 DynamoDB item and operation size gates

### LINKED ISSUE #1209: SEK-G68 Add opt-in DynamoDB maximum-item and operation byte budgets
# SEK-G68 Add opt-in DynamoDB maximum-item and operation byte budgets

Source: #1185. Dependency: SEK-G67 / PR1208.

# G68 accepted design and implementation code plan

Source1185; operator accepted registered backend limits checked before persistence, safe bug fixes and serial intent-cli delivery. Architect accepts two additive opt-in DynamoDB scopes from independent Opus5 correction review. G67 fully closed; baseline merge3ed403cec4f7b158395205abe149a6ca61e27bf3. This is accepted design and draft packet, not published or implementation WIP. Independent packet-readiness pass is required before publication.

## Code decisions

- Add public sealed DynamoDbMaxWrittenItemSizeMeasurement : IExecutorSizeMeasurement in the provider, Scope=dynamodb-max-written-item, MaximumItemBytes=409600. Measure max of the shared mapper event item and every mapped tag item for this event. Register MaxBytesPerEvent only, default409600 with smaller quota allowed and larger quota rejected. MaxBytesPerOperation is not a sum-of-all-items meaning on this max scope and must not be offered by the helper. Preserve G67 dynamodb-event-item unchanged.
- Add public sealed DynamoDbWriteOperationSizeMeasurement : IExecutorSizeMeasurement, Scope=dynamodb-write-operation, MaximumTransactionBytes=4194304. Measure event-item bytes plus the production event-Put condition-expression contribution plus all mapped tag-item bytes. Derive condition bytes from one internal provider constant shared by both transaction writer sites and the measurement; never duplicate a magic byte count. The measured quantity is transaction-payload contribution characterized against pinned DynamoDB Local, not stored-item bytes alone. Registration sets MaxBytesPerOperation only, default4194304, smaller quota allowed, larger rejected. This is an opt-in conservative application budget over the whole Executor operation, not per-transaction exact certification. Existing grouping/atomicity and fallback behavior remain unchanged. A caller hand-building a generic policy owns its quota and receives no helper ceiling claim.
- Both constructors accept DynamoDbEventStoreOptions? as G67 does. Reuse DynamoDbEventItemMapper.FromSerializableEvent and DynamoDbItemSizeCalculator; avoid divergent mapper/formula copies. Use context.ServiceId, same payload bytes/type/metadata, same WriteShardCount/options. Timestamp comparison may normalize only timestamp value with independent fixed UTC O-format33-byte evidence for event/tag rows. Unsupported representation/type/identity gives Unavailable. Preserve binding and exact conditional in-doubt behavior.
- Add AddDynamoDbMaxWrittenItemPolicy and AddDynamoDbWriteOperationPolicy extensions to ExecutorSizeGateOptions, plus AddSekibanDcbDynamoDbMaxWrittenItemSizeGate and AddSekibanDcbDynamoDbWriteOperationSizeGate on IServiceCollection following G67 IOptions registration. No new Core interface or changes to existing public signatures. Helpers expose only the relevant quota; direct generic policy behavior remains Core-owned. Both new scopes may coexist with G67.
- Reuse all five Executor admission entries before any durable mutation/enqueue. Max scope overlimit has IsOperationLimit=false; sum scope true. Existing strict capability exception and non-strict named diagnostic behavior remain.

## Evidence and tests

Independent correction report supplies characterization only; commit reproducible tests. Pin the existing DynamoDB Local image and use real AmazonDynamoDBClient on supported net9/net10 CI. Shared mapper output must match actual outgoing event/tag maps in typed, serialized and conditional routes, not just a synthetic reconstructed shape. Use independent expected byte count.

1. Legal pk/sk event-type expansion case: event item fits, one tag exceeds409600, total remains below4194304. G67 and sum scopes admit; max-item scope rejects with typed failure and zero SDK writes/enqueues. Test409599/409600/409601 largest-item boundary twice, with asserted event-largest and tag-largest orientations, using production mapping, not approximate constants.
2. Aggregate4194303/4194304/4194305 INCLUDING production event-Put condition contributions with individually legal items: evaluator boundary and real SDK Local acceptance/rejection; quote-heavy and plain payloads prove aggregate stored-byte accounting. Include conditional route with many legal tag rows and large EventPayloadName if aggregate is reachable. Do not assume conditional transactions cannot exceed4MiB: the same large eventType is repeated in every tag row.
3. Discriminators: event-only counting fails largest-tag oracle; max-only counting fails aggregate oracle. Last-event overlimit proves whole-operation zero dispatch; test strict/non-strict and representation guard separately for both scopes.
4. Coexistence, helper quota validation/defaults, DI service/options/shard propagation and focused reflection compatibility. Preserve reservation cleanup and serialized binding semantics with relevant existing tests.
5. Prove documented conservatism using an operation above4MiB that existing count grouping would split into individually legal transactions. Admission rejects solely because caller selected whole-operation budget; no grouping change.
6. Capture an actual SDK BatchWriteItem body containing25 production-shaped legal items with escaped content and assert wire bytes exceed16MiB although item-byte arithmetic is below it. This is an exclusion discriminator, not a supported wire guarantee or exact escape-ratio contract.

## Exclusions and later candidates

EN/JA command and provider docs must distinguish per-event maximum, whole-operation sum, stored vs wire bytes, helper ceilings and caller quota. Exclude key length, operation count, BatchWriteItem wire envelope, GSI/billing footprint, other providers and Queue envelopes. Both scopes cover only event/tag Put items currently emitted by the mapper, not future outbox/head/other items. Source1185 stays open; no release or data migration.

MaxBatchWriteItems validation is a separate bug candidate:0 causes a non-progressing chunk loop; values above25 violate service count. Investigate timing/constructor/direct options mutation and validation path before a separate packet; do not slip a default runtime behavior change into this opt-in slice.

Cosmos measurement remains design work. Graceful shutdown and classification-only observability remain separate. Recovery alternatives are not ready for operator selection: a finite safe window alone is no proof against arbitrarily late commits; resolver purity is not guaranteed by an interface signature; subscriber-selective replay over shared streams needs an actual routing seam. Validate these facts before proposing checkpoint/outbox contracts. No new durable state, terminal retry stop, transport, splitting,1169 or1193 work.

## Packet readiness B1 and notes incorporated

B1 accepted: Local production-condition probes exposed undercount in earlier item-only aggregate characterization. Use the exact production condition literal at both writer sites and measurement, with UTF8 byte derivation. Capture actual outgoing condition expression and compare to the shared constant. Item-only aggregate mutant must admit an operation that corrected measurement rejects; actual Local rejection payload size must match the corrected measurement. Include positive inclusive boundary through the production-shaped request, conditional multi-tag path, and ClientRequestToken non-contribution. Do not claim these probes are live AWS validation. The budget also conservatively includes the per-event condition contribution when existing fallback uses BatchWriteItem; it is never actual BatchWrite wire certification.

N1: document each registered provider scope maps independently; advise registering only needed scopes. No memoization introduced in this packet because service/options/timestamp identity and lifetime would need separate validation.
N2: both event-largest and tag-largest boundary triples mandatory. N3: wire-exclusion discriminator intentionally allocates roughly70MB; serialize it with other heavy tests, assert threshold only, never exact escaping ratios. N4: separate batch-count bug remains deferred.

## Acceptance criteria

AC1: Implement the two exact StorageItem measurements, max(event and each tag item) and sum(event item plus its production event-Put condition-expression contribution plus every tag item), with names/constructors/constants and helper quota semantics in the accepted code plan. Preserve G67 scope and legacy behavior. No Core interface change. Share an internal condition constant with both transaction writer sites and derive UTF8 bytes, not a duplicated magic count. Call this transaction-payload contribution.

AC2: Reuse canonical mapper/calculator with actual serializer/service/options/identity/timestamp parity across typed, serialized and conditional writes. Unsupported/unprovable input returns Unavailable. Preserve serialized binding and exact in-doubt propagation.

AC3: Add policy and DI helpers with409600 max-item and4194304 operation defaults/ceilings, lower caller budgets, coexistence and correct IOptions propagation. Existing public constructors/DI remain compatible. Direct generic policy does not imply helper guarantees.

AC4: Reuse five Executor prewrite entries and strict/non-strict behavior. Largest-tag and aggregate excess yield typed errors with correct IsOperationLimit; last-event failure causes zero mutations/enqueues and preserves reservation cleanup.

AC5: Commit independent count and actual outgoing map tests on all3provider routes; largest-item409599/409600/409601 and aggregate4194303/4194304/4194305; event-vs-tag-vs-sum discriminators, legal keys, escaped/UTF8/meta/service/shard content and reachable conditional aggregate. Demonstrate operation-budget conservatism without changing grouping. Both event-largest and tag-largest boundary triples are mandatory. Aggregate boundary includes production condition contributions; actual requests carry those conditions. Include item-only undercount mutant, Local reported-size equality on rejection and ClientRequestToken non-contribution.

AC6: Use real AmazonDynamoDBClient and existing pinned Local in required net9/net10 CI for inclusive item/aggregate boundary oracle. No raw CLI substitute, skip, waiver or live AWS credentials. Assert zero SDK dispatch for gate rejection independently. Keep characterizations reproducible.

AC7: Capture real SDK25-item BatchWriteItem body showing stored-byte count cannot certify wire limit; do not pin incidental exact escape ratios. Focused public-surface reflection, helper/DI and existing size/conditional regressions plus required CI pass. Wire-exclusion capture has deliberate roughly70MB allocation; avoid concurrency with other heavy tests and assert only threshold, not exact escape ratios.

AC8: Update EN/JA command and storage docs with runnable opt-in setup, scopes, quota/error semantics and precise exclusions from accepted code plan. Operation sum is a conservative application budget, never exact transaction-fit certification. Source1185 stays open; no release, migration, newtransport, grouping change, shutdown/retry behavior change or MaxBatchWriteItems validation in this packet. Explain operation contribution includes shared production event-Put condition bytes and is conservatively charged on BatchWrite fallback too. Document independent mapping cost per scope and advise only needed registrations; no memoization in this slice.

## Goal

Reject oversized mapped event/tag items and configured whole-operation transaction-payload contribution before persistence through additive opt-in DynamoDB policies.

## Why This Slice Exists Now

G67 covers base event items. Reviewed production-shaped probes demonstrate remaining tag-item and transaction-contribution admission gaps.

## Current Observed State

G67 is merged and fully closed. G68 is draft, awaiting independent readiness closure; no implementation or release.

G67 merge3ed403cec4f7b158395205abe149a6ca61e27bf3, shared mapper/calculator and five Executor admission entries. Start from current main.

## Target Repo / Path / Part

J-Tech-Japan/Sekiban; dcb/src/Sekiban.Dcb.DynamoDB, dcb/tests, docs/dcb_llm, docs/dcb_llm_ja, existing .github/workflows test wiring if needed. Provider measurements and additive registration helpers.

## In Scope

Two scopes, shared condition constant, production mapping parity, admission tests and EN/JA docs as specified in the code plan and AC1–AC8.

## Out Of Scope

Core interface changes, transaction regrouping, BatchWrite wire certification, batch-count validation, Cosmos implementation, shutdown/retry changes, durable recovery state, release, migration, new transport, splitting,1169 and1193.

## Verification

AC1–AC8 above are mandatory, including both TFMs, actual SDK/pinned Local boundary tests, exact-head Opus5 review and required CI. Never weaken tests or substitute raw CLI characterization for SDK integration.

## Related Links

- https://github.com/J-Tech-Japan/Sekiban/pull/1208
- intents/sekiban/design/sek-g68-accepted-code-plan-2026-09-10.md
- intents/sekiban/design/sek-g68-packet-readiness-review-2026-09-10.md

## Base Branch Policy

Direct main: PR base main, begin from current main. Merge individually only after exact-head review and CI, complete host closeout before next implementation.

## Accepted F4 amendment — 2026-09-10

Architect accepts the independently reviewed two-sided conflict rejection design. This amendment supersedes the earlier blanket no-Core-change boundary ONLY for dcb/src/Sekiban.Dcb.Core/SizeGates/SekibanDcbExecutorSizeGateExtensions.cs. No new public signature/interface or other Core change. F1-F3 repaired atbb8aa400; F4 remains required before approval/merge.

Code: introduce internal ExecutorSizeGateRegistrationMarker in that existing Core file, accessible to DynamoDB through the existing InternalsVisibleTo grant. Marker identity is its dedicated ServiceType, with CoreConvenience vs ProviderComposable mechanism; do not infer identity from unrelated descriptors. Both Core overloads and AddDynamoDbGateConfiguration consult it. A second different mechanism throws InvalidOperationException naming the conflict and supported setup, at registration before mutating IServiceCollection. Same-mechanism repetitions preserve existing behavior: Core remains last-wins, provider helpers accumulate. No manual descriptor bypass guarantee.

Important validation-order refinement: the supplied review pseudocode places a mutating Declare before Core configure/Validate. That would leave a marker after an invalid first call, violating the clean failure behavior. Use a read-only compatibility check after null checks and before callbacks, then existing configure/Validate, and only declare the marker immediately before adding the successfully validated singleton. Provider calls compatibility check/declaration before its other service additions. Keep guard state in IServiceCollection only, not global. Test invalid Core registration on an empty collection leaves its descriptor list unchanged and a subsequent provider registration succeeds. Rejected mixed calls must also leave the collection unchanged.

Both Core overloads retain eager validation and original exception types, supplied options instance reference identity, and no-registration ungated behavior. The guard rejects all12 combinations of two Core overloads, two orders and three provider helpers. Verify actionable message and unchanged descriptor count; existing three-provider accumulation/DI proof, Core instance identity, invalid policy timing, repeated same-mechanism behavior and no new public surface. Both-sidedness is demonstrated by tests that fail if either side omits its guard.

Supported migration: a single Core callback may combine logical and provider policy extensions only when caller passes the same DynamoDbEventStoreOptions instance used by the store. Prove same-shard2 measurement parity with provider DI registration, plus an omitted-options negative comparison showing the expected undercount. Update all4 EN/JA command/storage docs with both supported setup examples, mixed-registration error, explicit store-options responsibility and manual AddSingleton bypass limitation. Note G67 hand-built mismatched measurement contexts now yield Unavailable in the PR description.

AC9: Close F4 by the above internal two-sided registration guard, validation-failure atomicity and full registration/ABI/migration tests. Existing F1-F3 and allAC1-8 continue to apply except the explicit single Core-file target exception. Same PR1210, fresh CI on both TFMs, independent exact-head Opus5 rereview after all repairs. No release, schema migration, unrelated batch-count fix, transport,1169 or1193.

tomohisa:
Closes #1209

## Summary
- Adds opt-in `dynamodb-max-written-item` and `dynamodb-write-operation` StorageItem measurements using the production mapper/calculator.
- Shares the production event-Put condition expression and UTF-8 contribution across both transaction writer sites and operation accounting.
- Adds additive policy/DI helpers, preserves G67 coexistence and all existing executor admission routes, and documents EN/JA scope, strictness, stored-vs-wire limits, and exclusions.
- Adds provider-shaped typed, serialized, conditional, BatchWrite exclusion, zero-dispatch, boundary, conservative aggregate, compatibility, and DI proofs.

## Validation
- `dotnet build .../Sekiban.Dcb.WithResult.Tests.csproj -c Release --no-restore -f net9.0 /p:UseSharedCompilation=false`: succeeded, 0 errors.
- Same build for `net10.0`: succeeded, 0 errors.
- Focused measurement/surface/event-item tests: net9 39/39 passed; net10 39/39 passed.
- Pinned DynamoDB Local `amazon/dynamodb-local:2.6.1@sha256:1856c05cc66a0e49dc1099e483ad2851477eeebe2135250ac11a1d1227db54b1`: net10 G67/G68 Local tests 4/4 passed, including 4 MiB under/over and persisted-map parity.
- The bounded net9 Local attempt is recorded as an environment limitation: macOS .NET/AWS SDK initialization failed before a request with `System.TypeInitializationException` / `CookieContainer` inner `InvalidOperationException: GetDomainName: -1`; no product assertion was reached.
- `git diff --check`: passed.

The operation budget remains a conservative application contribution, not wire-size certification; BatchWrite 25-item wire evidence is an explicit exclusion discriminator. No MaxBatchWriteItems validation or other transport/provider scope is included.

## Final review and compatibility note

Exact head c55b7c02f5403fb9059bc060ba7ef7f53d22fe45 passed independent Opus5 re-review with F1-F4 and AC1-AC9 closed. Required CI passed on net9/net10, including all46 G68-suite cases per framework and real SDK/pinned DynamoDB Local evidence.

G67 `dynamodb-event-item` now returns `Unavailable` for hand-built measurement contexts whose `Event` and `SerializedEvent` disagree on id, event type, sortable-unique id or tags (strict mode raises `ExecutorSizeCapabilityException`). Production context-construction paths preserve this identity. Mixing Core convenience registration with provider-specific DI helpers now throws at registration instead of silently discarding policies; use one supported mechanism and preserve the same store options.

### DIFF
--- a/dcb/src/Sekiban.Dcb.Core/SizeGates/SekibanDcbExecutorSizeGateExtensions.cs
+++ b/dcb/src/Sekiban.Dcb.Core/SizeGates/SekibanDcbExecutorSizeGateExtensions.cs
@@ -4,6 +4,8 @@ namespace Sekiban.Dcb.SizeGates;
 
 public static class SekibanDcbExecutorSizeGateExtensions
 {
+    private const string CoreConvenienceMechanism = ExecutorSizeGateRegistrationMarker.CoreConvenience;
+
     /// <summary>
     /// Registers one opt-in executor size policy set. No service is added to the container when this method is not
     /// called, so existing construction remains ungated.
@@ -15,9 +17,11 @@ public static class SekibanDcbExecutorSizeGateExtensions
         ArgumentNullException.ThrowIfNull(services);
         ArgumentNullException.ThrowIfNull(configure);
 
+        ExecutorSizeGateRegistrationMarker.EnsureCompatible(services, CoreConvenienceMechanism);
         var options = new ExecutorSizeGateOptions();
         configure(options);
         options.Validate();
+        ExecutorSizeGateRegistrationMarker.Declare(services, CoreConvenienceMechanism);
         services.AddSingleton(options);
         return services;
     }
@@ -28,8 +32,51 @@ public static class SekibanDcbExecutorSizeGateExtensions
     {
         ArgumentNullException.ThrowIfNull(services);
         ArgumentNullException.ThrowIfNull(options);
+        ExecutorSizeGateRegistrationMarker.EnsureCompatible(services, CoreConvenienceMechanism);
         options.Validate();
+        ExecutorSizeGateRegistrationMarker.Declare(services, CoreConvenienceMechanism);
         services.AddSingleton(options);
         return services;
     }
+}
+
+/// <summary>
+/// Internal registration identity shared by the Core convenience and provider-composable gate helpers.
+/// The marker is held in the caller's service collection so separate registrations cannot silently replace one another.
+/// </summary>
+internal sealed class ExecutorSizeGateRegistrationMarker
+{
+    internal const string CoreConvenience = "Core convenience gate registration";
+    internal const string ProviderComposable = "provider-composable gate registration";
+
+    private ExecutorSizeGateRegistrationMarker(string mechanism)
+    {
+        Mechanism = mechanism;
+    }
+
+    internal string Mechanism { get; }
+
+    internal static void EnsureCompatible(IServiceCollection services, string mechanism)
+    {
+        var existing = Find(services);
+        if (existing is null || existing.Mechanism == mechanism)
+            return;
+
+        throw new InvalidOperationException(
+            $"'{existing.Mechanism}' and '{mechanism}' cannot both configure the executor size gate. " +
+            "Configure every policy through one mechanism: either a single " +
+            "AddSekibanDcbExecutorSizeGate callback that calls the provider policy extensions, " +
+            "or only the AddSekibanDcbDynamoDb*SizeGate helpers.");
+    }
+
+    internal static void Declare(IServiceCollection services, string mechanism)
+    {
+        EnsureCompatible(services, mechanism);
+        if (Find(services) is null)
+            services.AddSingleton(new ExecutorSizeGateRegistrationMarker(mechanism));
+    }
+
+    private static ExecutorSizeGateRegistrationMarker? Find(IServiceCollection services) =>
+        services.FirstOrDefault(descriptor => descriptor.ServiceType == typeof(ExecutorSizeGateRegistrationMarker))
+            ?.ImplementationInstance as ExecutorSizeGateRegistrationMarker;
 }
--- a/dcb/src/Sekiban.Dcb.DynamoDB/DynamoDbEventItemSizeMeasurement.cs
+++ b/dcb/src/Sekiban.Dcb.DynamoDB/DynamoDbEventItemSizeMeasurement.cs
@@ -36,11 +36,7 @@ public sealed class DynamoDbEventItemSizeMeasurement : IExecutorSizeMeasurement
 
         try
         {
-            var mapped = DynamoDbEventItemMapper.FromSerializableEvent(
-                context.SerializedEvent,
-                context.ServiceId,
-                _writeShardCount,
-                timestamp: DateTimeOffset.UtcNow);
+            var mapped = DynamoDbMeasurementContextMapper.Map(context, _writeShardCount);
             return ExecutorSizeMeasurementResult.Exact(
                 DynamoDbItemSizeCalculator.Calculate(mapped.DynamoEvent.ToAttributeValues()));
         }
@@ -49,6 +45,147 @@ public sealed class DynamoDbEventItemSizeMeasurement : IExecutorSizeMeasurement
             return ExecutorSizeMeasurementResult.Unavailable(
                 $"DynamoDB event-item mapping failed with {ex.GetType().Name}");
         }
+    }
+}
+
+/// <summary>
+/// Measures the largest mapped DynamoDB item emitted for one event, including every tag item.
+/// </summary>
+public sealed class DynamoDbMaxWrittenItemSizeMeasurement : IExecutorSizeMeasurement
+{
+    /// <summary>The DynamoDB service item ceiling: 400 KiB.</summary>
+    public const long MaximumItemBytes = DynamoDbEventItemSizeMeasurement.MaximumItemBytes;
+
+    /// <summary>The canonical provider scope for the largest written item.</summary>
+    public const string Scope = "dynamodb-max-written-item";
+
+    private readonly int _writeShardCount;
+
+    /// <summary>Creates a measurement using the provider's configured write-shard count when supplied.</summary>
+    public DynamoDbMaxWrittenItemSizeMeasurement(DynamoDbEventStoreOptions? options = null)
+    {
+        _writeShardCount = Math.Max(1, options?.WriteShardCount ?? 1);
+    }
+
+    /// <summary>Measures the largest event or tag item mapped for the serialized executor boundary.</summary>
+    public ExecutorSizeMeasurementResult Measure(ExecutorSizeMeasurementContext context)
+    {
+        ArgumentNullException.ThrowIfNull(context);
+
+        if (context.Representation != ExecutorSizeRepresentation.StorageItem)
+        {
+            return ExecutorSizeMeasurementResult.Unavailable(
+                "DynamoDB maximum-written-item measurement supports only the StorageItem representation");
+        }
+
+        try
+        {
+            var mapped = DynamoDbMeasurementContextMapper.Map(context, _writeShardCount);
+            var largest = DynamoDbItemSizeCalculator.Calculate(mapped.DynamoEvent.ToAttributeValues());
+            foreach (var tag in mapped.DynamoTags)
+            {
+                largest = Math.Max(largest, DynamoDbItemSizeCalculator.Calculate(tag.ToAttributeValues()));
+            }
+
+            return ExecutorSizeMeasurementResult.Exact(largest);
+        }
+        catch (Exception ex) when (ex is ArgumentException or FormatException or InvalidOperationException or OverflowException)
+        {
+            return ExecutorSizeMeasurementResult.Unavailable(
+                $"DynamoDB maximum-written-item mapping failed with {ex.GetType().Name}");
+        }
+    }
+}
+
+/// <summary>
+/// Measures the provider's conservative transaction-payload contribution for one executor event.
+/// </summary>
+public sealed class DynamoDbWriteOperationSizeMeasurement : IExecutorSizeMeasurement
+{
+    /// <summary>The DynamoDB transaction payload ceiling used by this opt-in application budget.</summary>
+    public const long MaximumTransactionBytes = 4L * 1024L * 1024L;
+
+    /// <summary>The canonical provider scope for the whole write operation contribution.</summary>
+    public const string Scope = "dynamodb-write-operation";
+
+    private readonly int _writeShardCount;
+
+    /// <summary>Creates a measurement using the provider's configured write-shard count when supplied.</summary>
+    public DynamoDbWriteOperationSizeMeasurement(DynamoDbEventStoreOptions? options = null)
+    {
+        _writeShardCount = Math.Max(1, options?.WriteShardCount ?? 1);
+    }
+
+    /// <summary>
+    /// Measures event and tag item bytes plus the production event-Put condition contribution. The result is a
+    /// conservative application budget and is not an exact transaction request-wire certification.
+    /// </summary>
+    public ExecutorSizeMeasurementResult Measure(ExecutorSizeMeasurementContext context)
+    {
+        ArgumentNullException.ThrowIfNull(context);
+
+        if (context.Representation != ExecutorSizeRepresentation.StorageItem)
+        {
+            return ExecutorSizeMeasurementResult.Unavailable(
+                "DynamoDB write-operation measurement supports only the StorageItem representation");
+        }
+
+        try
+        {
+            var mapped = DynamoDbMeasurementContextMapper.Map(context, _writeShardCount);
+            return ExecutorSizeMeasurementResult.Exact(
+                DynamoDbTransactionSizeAccounting.CalculateContribution(mapped));
+        }
+        catch (Exception ex) when (ex is ArgumentException or FormatException or InvalidOperationException or OverflowException)
+        {
+            return ExecutorSizeMeasurementResult.Unavailable(
+                $"DynamoDB write-operation mapping failed with {ex.GetType().Name}");
+        }
+    }
+}
+
+internal static class DynamoDbMeasurementContextMapper
+{
+    public static DynamoDbEventWriteItem Map(ExecutorSizeMeasurementContext context, int writeShardCount)
+    {
+        ArgumentNullException.ThrowIfNull(context.Event);
+        ArgumentNullException.ThrowIfNull(context.SerializedEvent);
+
+        if (context.Event.Id != context.SerializedEvent.Id ||
+            !string.Equals(context.Event.EventType, context.SerializedEvent.EventPayloadName, StringComparison.Ordinal) ||
+            !string.Equals(
+                context.Event.SortableUniqueIdValue,
+                context.SerializedEvent.SortableUniqueIdValue,
+                StringComparison.Ordinal) ||
+            !context.Event.Tags.SequenceEqual(context.SerializedEvent.Tags, StringComparer.Ordinal))
+        {
+            throw new InvalidOperationException(
+                "DynamoDB measurement requires matching event and serialized-event identity and tags");
+        }
+
+        return DynamoDbEventItemMapper.FromSerializableEvent(
+            context.SerializedEvent,
+            context.ServiceId,
+            writeShardCount,
+            timestamp: DateTimeOffset.UtcNow);
+    }
+}
+
+internal static class DynamoDbTransactionSizeAccounting
+{
+    /// <summary>Production event-Put guard shared by conditional and ordinary transaction writers.</summary>
+    internal const string EventPutConditionExpression = "attribute_not_exists(pk)";
+
+    internal static long CalculateContribution(DynamoDbEventWriteItem mapped)
+    {
+        var bytes = DynamoDbItemSizeCalculator.Calculate(mapped.DynamoEvent.ToAttributeValues());
+        bytes = checked(bytes + Encoding.UTF8.GetByteCount(EventPutConditionExpression));
+        foreach (var tag in mapped.DynamoTags)
+        {
+            bytes = checked(bytes + DynamoDbItemSizeCalculator.Calculate(tag.ToAttributeValues()));
+        }
+
+        return bytes;
     }
 }
 
--- a/dcb/src/Sekiban.Dcb.DynamoDB/DynamoDbEventStore.cs
+++ b/dcb/src/Sekiban.Dcb.DynamoDB/DynamoDbEventStore.cs
@@ -75,7 +75,7 @@ public partial class DynamoDbEventStore : IHotEventStore, IStorageDurabilityDesc
                 {
                     TableName = _context.EventsTableName,
                     Item = dynEvent.ToAttributeValues(),
-                    ConditionExpression = "attribute_not_exists(pk)"
+                    ConditionExpression = DynamoDbTransactionSizeAccounting.EventPutConditionExpression
                 }
             }
         };
@@ -902,7 +902,7 @@ public partial class DynamoDbEventStore : IHotEventStore, IStorageDurabilityDesc
                     {
                         TableName = _context.EventsTableName,
                         Item = item.DynamoEvent.ToAttributeValues(),
-                        ConditionExpression = "attribute_not_exists(pk)"
+                        ConditionExpression = DynamoDbTransactionSizeAccounting.EventPutConditionExpression
                     }
                 });
                 eventIdsForToken.Add(item.DynamoEvent.EventId);
--- a/dcb/src/Sekiban.Dcb.DynamoDB/DynamoDbExecutorSizeGateExtensions.cs
+++ b/dcb/src/Sekiban.Dcb.DynamoDB/DynamoDbExecutorSizeGateExtensions.cs
@@ -1,4 +1,5 @@
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Options;
 using Sekiban.Dcb.SizeGates;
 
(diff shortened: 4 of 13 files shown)


tomohisa (commented):
Independent Opus 5 exact-head review at `bddf4b119d07af3d8f5ebe4eeb9f55652df626e3` against issue #1209 (packet `cf95a1f`), in a clean detached checkout. Read-only; no source, host, label, merge or release action. Head re-checked after the review: still `bddf4b11`.

**Verdict: request-update.** No functional defect in the shipped measurements — the B1 condition-contribution correction is implemented exactly as agreed and CI is green on both TFMs with every new test executed. Two acceptance clauses added by the `cf95a1f` revision are not met, plus two medium findings.

## CI (settled, job logs inspected)

`dcbTestsNet9` (102830587156) and `dcbTestsNet10` (102830586984) both **success**; PostgreSQL consumer, template validation and SonarCloud green. All 27 G68 tests ran and passed on both net9.0 and net10.0 with zero failures, including the three pinned-emulator tests — so the real `AmazonDynamoDBClient` reached DynamoDB Local on Linux CI. No skip, no waiver, no workflow change needed.

## Verified correct

- **B1 implemented as agreed.** `DynamoDbTransactionSizeAccounting.EventPutConditionExpression` is one internal constant now consumed by **both** writer sites (`DynamoDbEventStore.cs:78` conditional claim, `:905` transaction writer), and the measurement derives bytes with `Encoding.UTF8.GetByteCount(...)` rather than a literal `24`. This matches the accounting I measured against the service: payload size = Σ item bytes + Σ `ConditionExpression` bytes, with `ClientRequestToken` and `TableName` contributing nothing.
- **The max scope correctly stays free of condition bytes.** I verified on the pinned emulator that the 400 KiB per-item boundary is unaffected by an attached condition (409,600 accepted / 409,601 rejected with the production condition present), so charging 24 there would have been wrong — and it is not charged.
- **The new identity guard is safe.** `DynamoDbMeasurementContextMapper.Map` requires `Event`/`SerializedEvent` agreement on id, type, SUID and tags. Every production construction site (typed `ToSerializableEvent`, both serialized `ToEvent`) satisfies all four, so no production path can regress.
- **Registration shape is right.** Moving to `AddOptions` + `IConfigureOptions` + `TryAddSingleton` makes multiple DynamoDB helpers *accumulate* instead of shadowing; helper quotas are correctly asymmetric with both ceilings enforced and defaulted.
- **Route parity, strictness, zero dispatch, conditional aggregate** are all covered against actual captured requests, including asserting the production condition literal on the event `Put` and `null` on every tag `Put`. Docs (EN+JA, both files each) are honest: conservative BatchWrite charge stated as *not* wire certification, per-scope mapping cost documented, no false wire guarantee anywhere.

## Findings

**F1 (blocking, AC5 "legal keys") — the tag-largest boundary uses an illegal partition key.** `CreateTagLargestBoundary` reaches the target by inflating the **tag string**, which is embedded in `DynamoTag.Pk = "SERVICE#{serviceId}#TAG#{tagString}"`. Reproducing the construction against the production shapes:

| target | eventType | tagString | tag item | event item | `pk` bytes |
|---|---|---|---|---|---|
| 409,599 | 1 | 204,654 | 409,599 | 205,029 | **204,679** |
| 409,600 | 2 | 204,654 | 409,600 | 205,030 | **204,679** |
| 409,601 | 1 | 204,655 | 409,601 | 205,030 | **204,680** |

DynamoDB caps a partition key at 2,048 bytes; I confirmed on the pinned emulator that such a row is rejected with `"Hash primary key values must be under 2048 bytes…"` *before* item-size admission. So all three cases exercise a shape the service can never accept, and the tag-largest orientation — added precisely to protect the reachable hazard — never demonstrates it.

*Fix:* build it the reachable way — hold the tag string at ~2,000 bytes (`pk` ≈ 2,016) and binary-search **`EventPayloadName`** instead. Measured band: `eventType` 405,333 → tag row exactly 409,600 (event item 407,680); `eventType` 405,433 → tag row 409,700 rejected for item size while the event item still passes the G67 gate. `ConditionalRouteWithManyLegalTagRowsCountsLargeEventTypeInTheAggregate` already uses exactly this construction.
*Discriminating test:* assert inside the construction that `pk` ≤ 2,048 and `sk` ≤ 1,024 — the current shape fails immediately, the `eventType`-based shape passes; optionally round-trip 409,600 (accepted) and 409,700 (rejected for item size, not key length) through pinned Local.

**F2 (blocking, AC5/AC6) — the emulator aggregate omits the inclusive point and never pins the formula to the service.** `PinnedLocalCharacterizesUnderAndOverFourMiBOperationContribution` uses only 4,194,303 and 4,194,305. AC5 requires the 4,194,303/**4,194,304**/4,194,305 triple with actual requests carrying the conditions, and AC6 an *inclusive* boundary oracle; the unit theory covers 4,194,304 but the service-facing one does not. The over-limit branch also asserts only that an exception occurred, never comparing the service's reported `Payload Size: <n>` with the measured contribution — the AC5 clause that pins our formula to DynamoDB's own accounting.

*Failure scenario:* a regression that drops the condition term again survives this suite entirely — 4,194,303 stays accepted and 4,194,305 stays rejected even after a 24·N shift. The two missing assertions are exactly the ones that would fail.
*Fix:* add `[InlineData(4_194_304, true)]`, and on rejection parse the reported payload size and assert equality with `events.Sum(Measure(...))`. I measured that a contribution of exactly 4,194,304 is accepted by the service, so the case is satisfiable.

**F3 (medium, AC5) — no item-only undercount mutant.** AC5 requires one; `grep` for `mutant` returns nothing. Existing assertions would fail under an item-only implementation, but nothing demonstrates the failure mode (an item-only budget admitting an operation the service rejects). *Fix:* a test-project `ItemOnlyOperationMeasurement` registered at 4,194,304 that admits an operation the production measurement rejects, and which the emulator then refuses.

**F4 (medium, AC3) — Core gate helper + DynamoDB helpers silently drop policies.** `AddDynamoDbGateConfiguration` uses `TryAddSingleton`, while Core's `AddSekibanDcbExecutorSizeGate` uses `AddSingleton`. Core first → `TryAdd` is a no-op and the DynamoDB policies never reach the executor; DynamoDB first → Core's `AddSingleton` wins the resolve, same result. No error either way, so the executor runs a gate silently missing the scope the user registered. Not introduced by this PR, but AC3 asks for coexistence and the PR's coexistence test only covers the three DynamoDB helpers. *Fix:* detect a foreign pre-existing `ExecutorSizeGateOptions` descriptor in `AddDynamoDbGateConfiguration` and throw naming the conflict (or document prominently) — silence is not an acceptable third option. *Test:* both orders must either yield both scopes or throw.

*Note for the PR description:* the new identity guard also applies to G67's `dynamodb-event-item`, whose behaviour on a hand-built **mismatched** context changes from `Exact` to `Unavailable` (strict: `ExecutorSizeCapabilityException`). Unreachable from production, but it is a behaviour change to a published scope.

With F1–F4 closed I expect to approve at the next exact head. Full artifact: `sek-g68-pr1210-bddf4b11-exact-review-20260910.md`.

tomohisa (commented):
Independent Opus 5 exact-head re-review at `c55b7c02f5403fb9059bc060ba7ef7f53d22fe45` against the **updated** issue #1209 (AC1–AC9 with the accepted F4 single-Core-file amendment), in the prepared clean detached checkout. Read-only; no source, host, label, merge or release action. Head re-checked after the review: still `c55b7c02`. This verdict is bound to that head.

**Verdict: APPROVE.** F1–F4 are all closed with executable evidence, AC1–AC9 are satisfied, and CI is settled green on both TFMs with every repaired test verified to have run. One minor non-blocking item remains (a PR-description note), detailed at the end.

## Structure

`bb8aa400` is tests only (F1–F3, +251/−32). `c55b7c02` (F4) touches two production files — the single permitted Core file (+47) and the provider helper (+6) — plus one new test file and all four EN/JA docs. No other production file changed, so everything verified at `bddf4b11` carries forward unchanged.

## Closures

**F1 — closed.** `CreateTagLargestBoundary` now holds the tag at `"Tag:" + 1_900` characters and varies the **event type** instead, giving `pk` = 1,934 bytes — inside the 2,048-byte cap. The requested assertion is present and applied to *every* captured item: `Assert.InRange(GetByteCount(pk), 1, 2_048)` and `Assert.InRange(GetByteCount(sk), 1, 1_024)`. The old 204,679-byte partition key is gone and the shape now demonstrates the hazard that is actually reachable.

**F2 — closed.** `[InlineData(4_194_304, true)]` added, so the inclusive point is asserted against the service and not only in the unit theory. The rejection branch now does `Assert.Equal(expectedContribution, ReadReportedPayloadSize(exception))`, which walks the inner-exception chain and parses `Payload Size: <n>`. That pins the formula to DynamoDB's own accounting — a systematic error of the B1 class can no longer pass.

**F3 — closed, stronger than requested.** `PinnedLocalItemOnlyUndercountMutantAdmitsButServiceRejects` uses a real `ItemOnlyOperationMeasurement` and proves the whole failure mode: the item-only contribution is `production − 24 × eventCount` and sits *inside* the 4 MiB budget; an executor gated by the mutant **admits** the operation; the real provider then rejects it with a non-`ExecutorSizeLimitExceededException` whose reported payload size equals the production contribution; zero rows land in either table; and the production policy rejects the same operation at the gate with `MeasuredBytes == productionContribution`.

**F4 — closed, matching the amendment including the validation-order refinement.** `ExecutorSizeGateRegistrationMarker` is internal, lives in the single permitted Core file, identifies itself by its own dedicated `ServiceType`, and is consulted by both Core overloads and by `AddDynamoDbGateConfiguration`. The ordering is exactly as refined: `EnsureCompatible` (read-only) after null checks → `configure` → `Validate()` → `Declare` (first mutation) → `AddSingleton`. So a rejected mixed call and a failing `Validate()` both leave the collection untouched.

The 19 registration cases cover: the 12-case matrix (both Core overloads × both orders × three provider helpers) asserting `Assert.Equal(before, services.ToArray())` **and** `Assert.False(callbackInvoked)` — the latter proving the check precedes `configure`; invalid-Core atomicity for both overloads with a subsequent provider registration succeeding; last-wins plus `Assert.Same` instance identity; provider accumulation and order; ungated-with-no-marker; and the ABI check that the marker is not public. The matrix is itself the both-sidedness mutant: dropping Core's check fails the provider-first cases, dropping the provider's fails the Core-first cases.

**Migration shard parity closed too:** the Core-callback path with `WriteShardCount = 2` measures identically to the provider DI path, and omitting `storeOptions` measures **exactly 2 bytes fewer** — the negative half that makes the documentation requirement load-bearing.

## CI at this head (job logs inspected)

Run **34472828859** confirmed bound to `c55b7c02`. `dcbTestsNet9` (102856500554) and `dcbTestsNet10` (102856500814) both **success**; PostgreSQL consumer, template validation and SonarCloud green. **46 G68-suite tests passed on each TFM with zero failures** — 19 registration cases, 4 pinned-emulator tests (three-point inclusive boundary plus the item-only mutant), 23 measurement tests including the three legal-key tag-largest cases. The real `AmazonDynamoDBClient` again reached DynamoDB Local on Linux CI (the mutant test alone runs ~520 ms), so the local macOS SDK limitation was neither invoked nor needed. No skip, suppression or waiver anywhere.

## Scope and docs

The only Core change is the explicitly permitted `SizeGates/SekibanDcbExecutorSizeGateExtensions.cs`; no new public signature or interface; no schema/data migration, package upgrade, version bump, grouping change, `MaxBatchWriteItems` validation, transport, release or `#1169`/`#1193` contact. All four EN/JA docs carry both supported setups with a runnable example, the same-`storeOptions` responsibility and its undercount warning, the mixed-registration exception quoted with its actual mechanism names and its pre-mutation guarantee, the repeated-call semantics, and the manual `AddSingleton<ExecutorSizeGateOptions>` bypass limitation.

## Remaining minor item (non-blocking)

The amendment asks that the PR description note the G67 behaviour change; the body currently mentions neither "Unavailable" nor "mismatch". `DynamoDbMeasurementContextMapper.Map` applies its identity guard to all three scopes, so a **hand-built** context whose `Event` and `SerializedEvent` disagree on id, event type, sortable-unique id or tags now yields `Unavailable` (strict: `ExecutorSizeCapabilityException`) where G67 previously returned `Exact`. I re-checked every production construction site — none can produce such a context — so this is a description obligation, not a code issue. Suggested line: "G67 `dynamodb-event-item` now returns `Unavailable` for hand-built measurement contexts whose `Event` and `SerializedEvent` disagree on id, event type, sortable-unique id or tags; no production path can construct such a context." I did not edit the description myself.

Full artifact: `sek-g68-pr1210-c55b7c02-exact-rereview-20260910.md`.

PR #1212, opened 2026-09-10
Snapshot DynamoDB batch-write size per operation

### LINKED ISSUE #1211: SEK-G69 Validate and snapshot DynamoDB batch-write chunk size
# SEK-G69 Validate and snapshot DynamoDB batch-write chunk size

## Goal

Reject invalid DynamoDB batch-write chunk configuration before write dispatch and preserve one validated chunk size throughout an operation, including rollback.

## Why This Slice Exists Now

G68 is fully closed. Its design investigation found this separate mutable-option bug; operator approves safe bug fixes and Design/Review code planning before implementation. Corrected independent Opus5 plan is accepted.

## Current Observed State

At merged0832d10e, the private chunk helper reads mutable MaxBatchWriteItems in both loop increment and Take. Zero/negative makes an empty request rejected by the service;26 may create an oversized request. Mid-operation changes can alter later chunks and rollback. No unconditional production-hang or static26 orphan claim is made.

G68 PR1210 merged0832d10e and fully settled. Existing default100 transaction actions causes101-action event with100 tags to use batch fallback. Exact methods, code decisions and corrected test basis are embedded below; no host-file access is needed by Builder.

## Target Repo / Path / Part

J-Tech-Japan/Sekiban. Production: dcb/src/Sekiban.Dcb.DynamoDB/DynamoDbEventStore.cs only; tests under dcb/tests; docs/dcb_llm/11_storage_providers.md and docs/dcb_llm_ja/11_storage_providers.md.

## In Scope

Private ResolveBatchWriteChunkSize reads one local, validates1..25, returns local. WriteEventsWithBatchAsync calls it before list building and sends it to all event/tag/rollback ExecuteBatchWriteWithRetryAsync calls. Helper gains required int parameter; both loop reads use it. All public and non-batch semantics remain. Test mock records realistic request/state behavior and deterministic callbacks.

## Out Of Scope

MaxBatchGetItems/read-path sibling (including its rereading validator) deferred separately; no Core change, global constructor validation, clamping, byte-size policy change, transactional regrouping, retry policy or lifecycle change. No release, migration, provider/package upgrade, new transport,1169 or1193.

## Verification

All acceptance criteria below are required. Supported net9/net10 focused tests and CI, independent Opus5 exact-head review. Do not conflate mock control-flow proof with live service recovery. No additional emulator infrastructure needed for this bounded change.

## Related Links

- https://github.com/J-Tech-Japan/Sekiban/pull/1210
- intents/sekiban/design/dynamodb-batch-write-corrected-code-plan-2026-09-10.md
- intents/sekiban/design/dynamodb-batch-count-design-ruling-2026-09-10.md

## Base Branch Policy

Direct main. Start from current main after G68; one PR, fresh exact-head review and CI, individual merge then full host closeout.

## Acceptance Criteria

AC1: At entry to WriteEventsWithBatchAsync, call new private ResolveBatchWriteChunkSize. Read MaxBatchWriteItems exactly once into a local, validate1..25, throw ArgumentOutOfRangeException with ParamName MaxBatchWriteItems and captured ActualValue, return that local. No clamping. First-statement placement and single read are reviewed code-shape requirements, not fabricated timing tests.

AC2: Add required int chunkSize parameter to private ExecuteBatchWriteWithRetryAsync; pass the same snapshot to all three event/tag/rollback call sites. Loop increment and Take use only the parameter. No runtime rereads of MaxBatchWriteItems after entry within the operation, and no default parameter. Each later operation captures its own fresh value.

AC3: Preserve public getter/setter/default25, constructor/DI/API, initialization order, existing retry/rollback eligibility and original error behavior. Typed and serialized routes return failed ResultBox carrying the argument exception. Validation applies only to batch fallback; normal transaction and conditional operations remain unaffected by an unused invalid option. EnsureTablesAsync may still create tables first when AutoCreateTables is true; zero write-dispatch does not mean zero DDL.

AC4: Force real public fallback with default MaxTransactionItems100 and an event with100 distinct legal tags. On typed and serialized writes,0/-1/26 yields named argument exception with captured value and zero BatchWrite/TransactWrite event/tag mutation. Valid25/1 produces event[1] and tags[25,25,25,25]/[1x100]. Postconstruction and DI-resolved option mutation before call are covered.

AC5: Use deterministic recording client callbacks for in-flight mutation25->0 and25->26 during first event dispatch. With snapshot, later tag batches keep original chunk size; injected tag failure triggers legal rollback delete with same snapshot and propagates the original tag exception. Assert applied event state returns empty in the mock. Add multi-event (>25 event writes) case with one100-tag event to prove later EVENT chunks also keep snapshot, and rollback deletes partition with the original size. Next operation after changing25->10 uses10. Label mock state/control-flow evidence honestly, never live durable-orphaning proof.

AC6: Test double rejects empty and >25 actual per-table write lists with AmazonDynamoDBException (guard its behavior), rather than accepting impossible service requests. Baseline invalid0/negative typically dispatches an empty request then receives provider error, not a proven hang. Static configured26 does not imply every actual request is oversized; no fictitious static26 orphan proof or tagless outer-placement discriminator. Isolate stricter mock behavior from unrelated suites if necessary. Existing measured SDK/Local behavior is characterization; no new mandatory emulator test suite.

AC7: Run meaningful focused tests on net9/net10 and required CI. Verify no public surface change with focused reflection/legacy tests; unused-option transaction and conditional controls succeed. Single-read validation inspected directly; no timing stress test pretending to prove atomicity. Include event/tag/rollback snapshot discriminators and original-exception assertion.

AC8: Update EN/JA storage provider docs explaining1..25, batch-only validation timing and ResultBox error, per-operation snapshot, next-operation updates, table-init caveat and unchanged byte/wire guarantees. No Core/read-path fix, MaxBatchGetItems change, provider/package upgrade, migration, release, transport, retry stop,1169 or1193. Source1185 remains open for residual delivery work.

## Readiness notes incorporated

N1: Keep existing BatchWriteSdkBodyCanExceedSixteenMiBWhileTwentyFiveStoredItemsRemainLegal passing unchanged as a default25 fallback control; its single25-item request is compatible with stricter mock validation. N2: Only MaxBatchWriteItems is snapshotted; existing MaxRetryAttempts/backoff and other option reads remain unchanged. N3: Configured values above25 are rejected because they can produce invalid requests, even when a particular smaller request would fit. This does not weaken configured-value validation.

tomohisa:
Closes #1211

## Summary
- Validate MaxBatchWriteItems once at the start of each batch-write fallback operation.
- Reuse the captured 1..25 chunk size for event puts, tag puts, and rollback deletes without changing transaction or conditional paths.
- Add guarded mock coverage for invalid values, mutation during dispatch, rollback, next-operation reread, and service-invalid request shapes.
- Document timing, per-operation snapshot behavior, and table-initialization caveat in English and Japanese.

## Validation
- DynamoDbBatchWriteSnapshotTests: 10/10 on net9.0 and 10/10 on net10.0.
- Existing DynamoDB event-item, write-operation, and conditional-append suites: 52/52 on each TFM.
- DynamoDB provider builds: net9.0 and net10.0 succeeded with 0 warnings and 0 errors.
- git diff --check: clean.

### DIFF
--- a/dcb/src/Sekiban.Dcb.DynamoDB/DynamoDbEventStore.cs
+++ b/dcb/src/Sekiban.Dcb.DynamoDB/DynamoDbEventStore.cs
@@ -931,6 +931,7 @@ public partial class DynamoDbEventStore : IHotEventStore, IStorageDurabilityDesc
 
     private async Task WriteEventsWithBatchAsync(List<DynamoDbEventWriteItem> eventItems)
     {
+        var chunkSize = ResolveBatchWriteChunkSize();
         var eventWrites = eventItems
             .Select(item => new WriteRequest
             {
@@ -946,11 +947,11 @@ public partial class DynamoDbEventStore : IHotEventStore, IStorageDurabilityDesc
             })
             .ToList();
 
-        await ExecuteBatchWriteWithRetryAsync(_context.EventsTableName, eventWrites).ConfigureAwait(false);
+        await ExecuteBatchWriteWithRetryAsync(_context.EventsTableName, eventWrites, chunkSize).ConfigureAwait(false);
 
         try
         {
-            await ExecuteBatchWriteWithRetryAsync(_context.TagsTableName, tagWrites).ConfigureAwait(false);
+            await ExecuteBatchWriteWithRetryAsync(_context.TagsTableName, tagWrites, chunkSize).ConfigureAwait(false);
         }
         catch
         {
@@ -963,18 +964,35 @@ public partial class DynamoDbEventStore : IHotEventStore, IStorageDurabilityDesc
                     })
                     .ToList();
 
-                await ExecuteBatchWriteWithRetryAsync(_context.EventsTableName, deleteRequests).ConfigureAwait(false);
+                await ExecuteBatchWriteWithRetryAsync(_context.EventsTableName, deleteRequests, chunkSize).ConfigureAwait(false);
             }
 
             throw;
         }
     }
 
-    private async Task ExecuteBatchWriteWithRetryAsync(string tableName, List<WriteRequest> requests)
+    private int ResolveBatchWriteChunkSize()
     {
-        for (var i = 0; i < requests.Count; i += _options.MaxBatchWriteItems)
+        var chunkSize = _options.MaxBatchWriteItems;
+        if (chunkSize is < 1 or > 25)
         {
-            var batch = requests.Skip(i).Take(_options.MaxBatchWriteItems).ToList();
+            throw new ArgumentOutOfRangeException(
+                nameof(DynamoDbEventStoreOptions.MaxBatchWriteItems),
+                chunkSize,
+                "DynamoDB batch writes require between 1 and 25 items per request.");
+        }
+
+        return chunkSize;
+    }
+
+    private async Task ExecuteBatchWriteWithRetryAsync(
+        string tableName,
+        List<WriteRequest> requests,
+        int chunkSize)
+    {
+        for (var i = 0; i < requests.Count; i += chunkSize)
+        {
+            var batch = requests.Skip(i).Take(chunkSize).ToList();
             var pending = new Dictionary<string, List<WriteRequest>>
             {
                 [tableName] = batch
--- a/dcb/tests/Sekiban.Dcb.WithResult.Tests/DynamoDbBatchWriteSnapshotTests.cs
+++ b/dcb/tests/Sekiban.Dcb.WithResult.Tests/DynamoDbBatchWriteSnapshotTests.cs
@@ -2,11 +2,14 @@ using Amazon.DynamoDBv2;
 using Amazon.DynamoDBv2.Model;
 using Dcb.Domain;
 using Dcb.Domain.Student;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Sekiban.Dcb.Common;
 using Sekiban.Dcb.DynamoDB;
+using Sekiban.Dcb.Domains;
 using Sekiban.Dcb.Events;
 using Sekiban.Dcb.ServiceId;
+using Sekiban.Dcb.Storage;
 using Xunit;
 
 namespace Sekiban.Dcb.Tests;
@@ -43,7 +46,7 @@ public sealed class DynamoDbBatchWriteSnapshotTests
     [InlineData(25)]
     public async Task ValidBatchWriteSizes_AreUsedForEventAndTagRequests(int chunkSize)
     {
-        var (store, options, client) = NewStore(maxTransactionItems: 1);
+        var (store, options, client) = NewStore(maxTransactionItems: 100);
         options.MaxBatchWriteItems = chunkSize;
 
         var result = await store.WriteSerializableEventsAsync([CreateSerializedEvent(
@@ -75,6 +78,25 @@ public sealed class DynamoDbBatchWriteSnapshotTests
         Assert.Equal([25, 1], RequestSizes(client, options.EventsTableName));
         Assert.Equal([25, 25, 25, 25, 25], RequestSizes(client, options.TagsTableName));
         Assert.Equal(0, options.MaxBatchWriteItems);
+    }
+
+    [Fact]
+    public async Task MidOperationOptionMutationTo26_DoesNotSilentlyDropTheLastEvent()
+    {
+        var (store, options, client) = NewStore(maxTransactionItems: 100);
+        options.MaxBatchWriteItems = 25;
+        client.AfterBatchWrite = (tableName, dispatchNumber) =>
+        {
+            if (dispatchNumber == 1 && tableName == options.EventsTableName)
+                options.MaxBatchWriteItems = 26;
+        };
+
+        var result = await store.WriteSerializableEventsAsync(CreateMultiEventBatch());
+
+        Assert.True(result.IsSuccess);
+        Assert.Equal([25, 1], RequestSizes(client, options.EventsTableName));
+        Assert.Equal([25, 25, 25, 25, 25], RequestSizes(client, options.TagsTableName));
+        Assert.Equal(26, client.PersistedEventKeys.Count);
     }
 
     [Fact]
@@ -119,7 +141,7 @@ public sealed class DynamoDbBatchWriteSnapshotTests
     }
 
     [Fact]
-    public async Task InvalidBatchOptionDoesNotAffectTransactionFallbackOrConditionalPaths()
+    public async Task InvalidBatchOptionDoesNotAffectTransactionOrConditionalPaths()
     {
         var (store, options, client) = NewStore(maxTransactionItems: 100);
         options.MaxBatchWriteItems = 0;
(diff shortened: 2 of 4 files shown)


tomohisa (commented):
Independent Opus 5 exact-head review at `41a47aadbff37c51644e6634624e51201d96d2f4` against issue #1211 (AC1–AC8), in the prepared clean detached checkout. Read-only; no source, host, label, merge or release action. Head re-checked after the review: still `41a47aad`.

**Verdict: request-update.** The production fix is exactly right and CI is green on both TFMs with all ten new tests verified to have run. Three explicit acceptance clauses are not delivered.

## CI (job logs inspected)

Run **34482936990** confirmed bound to `41a47aad`. `dcbTestsNet9` (102889977909) and `dcbTestsNet10` (102889978276) both **success**; PostgreSQL consumer, template validation and SonarCloud green. All ten `DynamoDbBatchWriteSnapshotTests` cases passed on both net9.0 and net10.0 with zero failures, and the 21 legacy `DynamoDbSizeGateSurfaceTests`/`DynamoDbSizeGateRegistrationTests` cases remain green.

## Verified correct

- **AC1 single read.** `ResolveBatchWriteChunkSize` reads `_options.MaxBatchWriteItems` once into `chunkSize` and uses that local for the condition, for `ActualValue` and for the return — the ruling's C1 correction implemented literally, and notably better than the in-repo precedent `ValidateTaggedStreamBatchSize()`, which still re-reads three times and is correctly out of scope. First-statement placement, `ParamName` `"MaxBatchWriteItems"`, no clamping.
- **AC2.** Required `int chunkSize` (no default) on `ExecuteBatchWriteWithRetryAsync`, passed at all three call sites (`:950` events, `:954` tags, `:967` rollback), both loop reads on the parameter, and `_options.MaxRetryAttempts` deliberately still read per attempt — the freeze is scoped to the one option.
- **AC5's strongest cases.** `MidOperationOptionMutation_…` uses **default** `MaxTransactionItems = 100` with a 26-event batch whose first event carries 100 tags, mutates to `0` during the first event dispatch, and asserts events `[25, 1]` and tags `[25,25,25,25,25]` — the `[25, 1]` proves a later **event** chunk keeps the snapshot, which my own plan could not reach. `TagFailure_…` asserts `[25, 1, 25, 1]` (write plus rollback partitions), `Assert.Same(failure, …)` for the original exception, and `Assert.Empty(client.PersistedEventKeys)`. `NextOperationReads…` proves the per-operation reread with `[10, 10, 6]`. `GuardedClient…` guards the double itself.
- **AC3/AC8.** Public surface, init order, retry/rollback eligibility and the `ResultBox` conversion unchanged; EN and JA docs state 1..25, batch-only timing, the named exception, the snapshot reused for event/tag/rollback, next-operation observation, unchanged `MaxBatchGetItems`/`MaxRetryAttempts`/transaction/conditional, and the `AutoCreateTables` DDL caveat.
- No fabricated hang, static-26 orphan, tag-less discriminator or new emulator suite. ABI is covered by the green legacy surface tests, which AC7 explicitly permits.

## Findings

**F1 (blocking, AC5) — the `25 → 26` in-flight mutation variant is missing, and it covers a different read than `25 → 0`.** AC5 requires both; both mutation tests use `0`. AC2 names two reads and the variants pin different ones against the pre-fix shape:

- `25 → 0` exercises the `Take` read: `i += 0` and `Take(0)` produce an empty request the guarded double rejects — failure mode is a provider exception.
- `25 → 26` exercises the **increment** read: with 26 event writes `i = 0 + 26 = 26`, `26 < 26` is false, the loop exits, **the 26th event is never dispatched, no exception is raised and the operation reports success** — a silent lost write.

A mutant restoring the property read in the increment only therefore survives the entire current suite. *Fix:* a sibling test mutating to `26` on the first event dispatch, asserting the same `[25, 1]` partition **and** that all 26 events were dispatched (`PersistedEventKeys.Count == 26`). Under the pre-fix increment read it observes `[25]`, 25 persisted events and success.

**F2 (blocking, AC7) — no conditional control despite the test name.** `InvalidBatchOptionDoesNotAffectTransactionFallbackOrConditionalPaths` never calls `AppendIfUniqueAsync`, so the conditional half of its own name and of AC7's "transaction and conditional controls" is untested. *Fix:* add the `AppendIfUniqueAsync` control with `MaxBatchWriteItems = 0`, asserting success, one transaction request and zero batch requests. A mutant hoisting `ResolveBatchWriteChunkSize()` into the route methods or a shared helper fails it while every existing test passes.

**F3 (minor-blocking, AC4) — DI-resolved option mutation not covered.** Tests construct with `Options.Create(options)`; none goes through `AddSekibanDcbDynamoDb` + `BuildServiceProvider` and mutates the resolved `IOptions<>.Value`. G68's DI test already proves the registration yields that instance, so residual risk is small, but the clause is explicit and the test is cheap. A mutant snapshotting in the store constructor passes everything today and fails this.

**N1 (note, AC4 conformance).** `ValidBatchWriteSizes_…` sets `maxTransactionItems: 1` although its 100-tag event already forces the fallback at the default 100 (`1 + 100 = 101`). Dropping the override matches AC4's "default `MaxTransactionItems` 100" and strengthens the point that the fallback is reachable without unusual configuration. (`NextOperationReads…` and the invalid-size tests legitimately need it.)

With F1–F3 closed I expect to approve at the next exact head. Full artifact: `sek-g69-pr1212-41a47aad-exact-review-20260910.md`.

tomohisa (commented):
Independent Opus 5 exact-head re-review at `d8568cdd1f0f50da0a85baf452b2f1b1ef2304a9`, in a fresh clean detached checkout at that head. Read-only; no Git, label, merge, release or host mutation. Head re-checked after the review: still `d8568cdd`. This verdict is bound to `d8568cdd` and is **not** inherited from the `41a47aad` request-update verdict.

**Verdict: APPROVE.** F1–F3 and N1 are all closed with executable evidence, AC1–AC8 re-derived at this head are satisfied, and CI is green on both TFMs with all twelve tests verified to have run.

## Delta

`41a47aad..d8568cdd` is one test file, `+75 / −5`. The `dcb/src` and `docs` diffs are **empty**, so the production fix and EN/JA documentation carry forward byte-identical.

## AC1–AC3 re-derived independently

Grepping the whole of `DynamoDbEventStore.cs` at this head, `MaxBatchWriteItems` appears at exactly two lines: `:976` (the single value read into `chunkSize`) and `:980` (a `nameof` on the options **type**, which reads nothing). The resolver is the first statement of `WriteEventsWithBatchAsync` (`:934`) and the snapshot reaches all three call sites — `:950` events, `:954` tags, `:967` rollback — via a required parameter with no default. `_options.MaxRetryAttempts` is still read per attempt at `:839` and `:1001`, so retry behaviour is untouched and the freeze is scoped to the one option; the `MaxBatchGetItems` read path is unchanged and stays correctly out of scope.

## Closures

**F1 — closed.** `MidOperationOptionMutationTo26_DoesNotSilentlyDropTheLastEvent` uses **default `MaxTransactionItems = 100`**, mutates 25 → 26 on the first event dispatch, and asserts `[25, 1]` events, `[25,25,25,25,25]` tags **and `PersistedEventKeys.Count == 26`**. That pins the increment read the `25 → 0` case cannot reach: pre-fix, `i += 26` makes `26 < 26` false, the loop exits after the first chunk and the operation returns success with only 25 events written, so the test observes `[25]` and a count of 25 and fails. The two mutation tests now cover both reads AC2 names.

The count assertion is meaningful only because the double was corrected in the same commit — `PersistedEventKeys` now records puts and deletes for the **events table only** (`EventsTableName` wired in `NewStore` and in the DI test), where it previously also accumulated tag puts. I checked this does not weaken `TagFailure_…`: the double's per-table size guard and the injected `BatchWriteFailure` both throw **before** any key is recorded, so `Assert.Empty(PersistedEventKeys)` was sound before and is now unambiguous.

**F2 — closed.** Renamed `InvalidBatchOptionDoesNotAffectTransactionOrConditionalPaths` (dropping the inaccurate "Fallback") and now actually calls `AppendIfUniqueAsync` under `MaxBatchWriteItems = 0`, asserting success, zero batch requests and one transaction request, with request state cleared between the two halves so each assertion is about its own call.

**F3 — closed.** `DiResolvedOptionsMutationBeforeBatchFallbackIsValidatedWithoutDispatch` builds a real container via `AddSekibanDcbDynamoDb`, resolves both `IOptions<DynamoDbEventStoreOptions>.Value` and `DynamoDbEventStore` from DI (registered through `AddEventStoreAliases<DynamoDbEventStore>()`), mutates the resolved options to `0` after construction, then issues a 100-tag event — forcing the fallback at the default 100 — and asserts the typed exception with `ParamName`/`ActualValue` plus zero batch and zero transaction dispatch. A mutant snapshotting in the constructor passes everything else and fails this.

**N1 — closed.** `ValidBatchWriteSizes_…` now uses `maxTransactionItems: 100`, relying on its 100-tag event (`1 + 100 = 101`) to force the fallback under default configuration, exactly as AC4 words it.

## CI at this head (job logs inspected)

Runs **34485702250** (DCB), **34485702199** (packaged consumer) and **34485702239** (template) all confirmed bound to `d8568cdd` and successful; both SonarCloud checks green. **All twelve `DynamoDbBatchWriteSnapshotTests` cases passed on both net9.0 and net10.0 with zero failures**, including the three new ones. The 21 legacy `DynamoDbSizeGateSurfaceTests`/`DynamoDbSizeGateRegistrationTests` cases remain green on each framework, which is how AC7's ABI clause is satisfied — the legacy public surfaces are still frozen.

## Scope

Four files; no Core change, no read-path/`MaxBatchGetItems` expansion, no package upgrade, migration, version bump, transport, release or `#1169`/`#1193` contact. No fabricated hang, static-26 orphan, tag-less discriminator or new emulator suite; mock results are used as control-flow evidence only.

Full artifact: `sek-g69-pr1212-d8568cdd-exact-opus5-rereview-20260910.md`.

Issue #1217, opened 2026-09-10
VerifyAndExecute: fresh native projectors cannot obtain own-table bindings at apply time


tomohisa:
# Problem

A fresh native materialized-view projector that stores its own `MvTable` handles in `InitializeAsync` cannot obtain those handles when started in `VerifyAndExecute`. Schema verification can succeed, but the subsequent first event application dereferences an uninitialized table handle.

## Verified versions and evidence boundary

- NuGet 10.19.0 and 10.20.0: isolated runs using the actual `NativeMvApplyHost` / `NativeMvApplyHostFactory` reproduce the failure with synthetic events for PostgreSQL and SQL Server provider shapes.
- The two runs loaded distinct versioned assemblies (not the same binary relabeled).
- These isolated host-level runs record SQL statements without executing SQL. They are not evidence of a completed real-database executor regression. That regression is being added downstream.
- No production event payloads, credentials, infrastructure identifiers, or raw service logs are included here.

## Failure mechanism

1. A projector registers and retains its table handles only in `InitializeAsync`, then uses their physical names when generating apply statements.
2. The executor's verification branch calls the declarative schema contract and intentionally bypasses projector initialization, preserving its no-DDL/no-arbitrary-initialization-effects boundary.
3. The native host receives `IMvTableBindings tables` in `ApplyEventAsync`, but does not pass these own-view bindings to `NativeMvApplyContextAdapter`.
4. Consequently a fresh projector has neither initialized handles nor an apply-context API providing its own resolved table bindings.

Source anchors, fixed tag:

- [Native host and adapter](https://github.com/J-Tech-Japan/Sekiban/blob/dcb-v10.20.0/dcb/src/Sekiban.Dcb.MaterializedView/MvApplyHostSupport.cs)
- [Executor verification / initialization boundary](https://github.com/J-Tech-Japan/Sekiban/blob/dcb-v10.20.0/dcb/src/Sekiban.Dcb.MaterializedView/MvExecutorBase.cs)

## Minimal reproduction sequence

Use a synthetic projector with a table property initially unassigned, assignment through the init context's `RegisterTable`, declarative schema requirements, and an apply statement referencing that property's physical name.

1. Create a **new** projector and actual native host/factory.
2. Call `GetSchemaContract` with the correct provider/options bindings, without calling host/projector initialization.
3. Apply one synthetic serialized event through the actual native host.
4. Observe `NullReferenceException` at the table property's physical-name access.
[...]

PR #1219, opened 2026-09-10
feat: add Cosmos event-document size admission

### LINKED ISSUE #1218: SEK-G72 Cosmos serialized-item size admission
# SEK-G72 Cosmos serialized-item size admission

## Goal
Add an opt-in pre-persistence Cosmos item-document application quota using a truthful certified upper bound for the provider-owned default serializer.

## Why This Slice Exists Now
User accepted #1185 storage/delivery admission at Executor before durable writes. G66-G71 are merged and fully settled. Architect and Opus5 resolved code-plan findings including timestamp width, injected serializer trust, disposed fast path and bulk-resource materialization. D1 quota2,000,000 and D2 provider refactor accepted within that scope; release remains withheld.

## Current Observed State
Cosmos has three construction sites; each uses write-time DateTime.UtcNow. JSON UTC fractions vary by up to8 bytes; Local-kind can be wider. Measurement context contains no storage-plan token. Client creation is currently coupled to database initialization; context owns clients from connection/config constructors but adopts external clients for Aspire. Production sets AllowBulkExecution=true. Thus remapping with a fresh timestamp is not Exact, and custom serializer repeatability cannot be assumed.

Current merged main at least a69e1a3d3e7481f48a33515729a2ee5effb1c8a6, Cosmos SDK3.57.1. G66 evaluator already supports CertifiedBound/Unavailable and five entrances; G68 ProviderComposable marker/IVT support provider helpers. Use latest main and renew CI/review after rebase. No Core change needed.

## Target Repo / Path / Part
J-Tech-Japan/Sekiban. dcb/src/Sekiban.Dcb.CosmosDb/CosmosDbContext.cs, CosmosDbEventStore.cs, Models/CosmosEvent.cs; new CosmosEventDocumentMapper.cs, CosmosEventDocumentSizeMeasurement.cs, CosmosExecutorSizeGateExtensions.cs; focused tests under dcb/tests/Sekiban.Dcb.WithResult.Tests; docs/dcb_llm/11_storage_providers.md and docs/dcb_llm_ja/11_storage_providers.md. Preserve public CosmosEvent.FromEvent as a thin delegation to the same construction core, retaining its existing null/serviceId argument checks. Builder uses this standalone issue, not host state.

## In Scope
Provider-only behavior-preserving mapper refactor and context lifecycle split; opt-in StorageItem/cosmos-event-document measurement and quota helpers. The FromParts construction core accepts already-serialized payload text, original documentId, string tags, metadata and UTC timestamp. Typed writes pass the existing SerializeEventPayload result without serializing it again. FromSerializableEvent is a thin UTF-8 payload adapter for the other routes and measurement. Preserve actual ID/pk formatting and all argument behavior. Synthetic timestamp is for measurement only.

Use a private monitor around disposed check, client creation, serializer use/stream disposal and Dispose(bool). An internal EnsureClientCreated wrapper takes the monitor and calls a private no-lock helper; both existing-client return and creation follow disposed check. Internal TryMeasureSupportedDocument(object,out long,out string?) holds the monitor through serialization, checks provider ownership/settings, and returns false with named reason for unsupported cases. InitializeAsync releases the monitor before async database initialization. Preserve protected Dispose(bool) hook and injected ownership. No global caches. Internal creation-count instrumentation is explicitly allowed for concurrency verification. Serializer exceptions are translated to existing Unavailable semantics without swallowing cancellation incorrectly.

CosmosEventDocumentSizeMeasurement has internal context-bound constructor, public scope/default constants and Measure. Standard DI helper resolves the context lazily from the same service provider, missing context produces Unavailable at measurement. Explicit-context policy helper places same-context responsibility on the caller, as a stated limitation. Do not falsely prove identity merely from a parameter. G68 marker compatibility and eager quota validation must hold in both registration orders.

The bound covers event-item JSON only. Stable payload/metadata/tags/serviceId are copied without mutation; route-specific GUID differences must have equal serialized width proven against actual mapper. Timestamp sentinel is new DateTime(2000,1,1,0,0,0,DateTimeKind.Utc).AddTicks(1234567). Supported serialization returns finite seekable streams; any defensive unsupported stream returns Unavailable and is disposed.

## Out Of Scope
Exact byte identity through new Core token, timestamp changed to event-time, arbitrary/injected serializer certification, data migration, new durable state, transport envelope limits, aggregate-operation/tag/claim document limits, release,1169,1193.2,000,000 is an application quota equal to the decimal reading of the2MB documentation label; do not call it below both MB interpretations or certify service acceptance. New provider APIs are explicitly allowed; existing public signatures remain.

## Verification
Use named discriminators: maximum-width sentinel vs whole-second under-bound; actual route document comparison vs independent fourth mapper; existing-client dispose then measure vs disposed fast-path bypass; concurrent first use vs unsynchronized creation; oversized last event vs partial durable dispatch; injected stateful serializer vs assumed repeatability; G68 registration mixing versus silent policy replacement. Run focused/legacy tests bothTFMs and applicable required CI. Record exact commands/counts and independently review final head. SDK reflection probes are disposable evidence only, no production reflection. Point POST evidence is point-only; bounded SDK MaterializeResourceAsync probe recorded293 input/output byte-identical for3.57.1, but bulk envelope is unmeasured. Keep actual bulk setting; no hanging full dispatcher test requirement. Docs and tests must honor scope limitations. Every review finding proposes a correction and discriminating test.

## Related Links
- https://github.com/J-Tech-Japan/Sekiban/pull/1216
- https://learn.microsoft.com/en-us/azure/cosmos-db/concepts-limits
- https://learn.microsoft.com/en-us/azure/cosmos-db/transactional-batch

## Base Branch Policy
Direct main, serial individual merge, rebase next onto merged main with fresh CI and exact-head Opus5.

## Review Corrections Incorporated
F1/F2: one FromParts construction core; typed path does not call ToSerializableEvent; public FromEvent delegates with unchanged validation. Count actual typed payload serialization calls with a feasible existing serializer/payload instrumentation seam; equality alone does not detect duplicate serialization. No new public test API. Review must verify the test distinguishes a second serialization and preserves failures of the first.
F3: supported over-quota tests use Cosmos measurement plus an Executor recording store/publisher, as an admission test only. Do not pretend a foreign recording store proves actual Cosmos context identity. Assert no write invocation, not only an empty store after an attempted write. Injected CosmosClient is a separate Unavailable control. Three-route mapper/output tests remain distinct from this Executor gate control.
N1: eager Cosmos quota validation is deliberate at helper invocation, before service/policy mutations. N2: reject Local for bound safety and Unspecified for the UTC contract.

## Acceptance Criteria
- AC1: Extract one internal CosmosEventDocumentMapper.FromParts(string serviceId,Guid documentId,string sortableUniqueId,string eventType,string serializedPayload,IReadOnlyList<string> tags,EventMetadata metadata,DateTime utcTimestamp) construction core and a thin FromSerializableEvent adapter. Typed writes use their existing serializedPayload directly, without ToSerializableEvent or another payload serialization; serialized and conditional writes use the adapter. Preserve public CosmosEvent.FromEvent signature, argument validation and DateTime.UtcNow behavior by delegating its construction to FromParts. All paths retain original IDs/pk,payload/metadata/tags,CreateItemAsync and Conflict classification. Non-UTC rejection belongs in the core. No schema or stored timestamp semantics change; verify one typed payload serialization, not merely document equality.
- AC2: CosmosDbContext uses one private monitor for client creation, measurement serializer use/stream disposal and Dispose(bool). Check disposed before every return/creation, including existing-client paths; no volatile bypass. InitializeAsync performs database work outside this monitor. Concurrent first use creates one client; disposal never resurrects it or disposes an injected client. A declared internal creation-count seam is allowed, no public test API. State explicitly that general in-flight database I/O shutdown is not solved.
- AC3: Context-owned internal TryMeasureSupportedDocument serializes only on the provider-owned client with the supported CamelCase default serializer options. Injected/unproven clients, disposed contexts, missing DI registration and unsupported representation return named Unavailable. Pure client creation performs no database/container initialization. Streams disposed on all paths; nonseekable/serialization failures are defensive Unavailable paths, not a new generic counting facility. No arbitrary serializer trust or public ownership bool.
- AC4: CosmosEventDocumentSizeMeasurement implements IExecutorSizeMeasurement for StorageItem scope cosmos-event-document and returns CertifiedBound, never Exact. Measure a mapped document using fixed measurement-only UTC sentinel 2000-01-01T00:00:00Z plus1234567 ticks; writers never store the sentinel. For supported immutable inputs, prior bound covers actual serialized item bytes with0..8-byte timestamp slack. Assert against0/1/3/7 fractional UTC times, year boundaries and small ASCII/large CJK with null/escaping metadata; mapper rejects Local and Unspecified. Do not claim timestamp fixture alone proves arbitrary serializer behavior. Local rejection protects the byte bound; Unspecified is rejected for the explicit UTC contract, not because it exceeds the sentinel width.
- AC5: Add provider APIs AddCosmosEventDocumentPolicy(ExecutorSizeGateOptions,CosmosDbContext,long? maxBytesPerEvent,ExecutorSizeStrictness) and AddSekibanDcbCosmosEventDocumentSizeGate(IServiceCollection,long? maxBytesPerEvent,ExecutorSizeStrictness), default Strict. Default and helper maximum are2,000,000 bytes; reject nonpositive or larger quota and permit smaller positive quota. Preserve G68 ProviderComposable registration coexistence and Core-convenience mixing rejection without partial registration on failure. DI resolves the same singleton context used by the standard store registration. Explicit-context helper documents caller identity responsibility; no false automatic mismatch detection. Cosmos quota arguments are deliberately validated at helper invocation before changing the service collection or policy list; do not defer validation to service resolution. Test invalid2,000,001 and nonpositive values leave registration unchanged. This timing is explicit, not an assumed sibling-provider guarantee.
- AC6: Test over-quota admission at Executor level using a provider-owned supported Cosmos measurement context plus a recording/in-memory event store and recording publisher. On all five existing gate entrances, an oversized last event causes ExecutorSizeLimitExceededException with IsCertifiedUpperBound=true before any durable-store dispatch or publisher enqueue; assert invocation counts as well as empty retained events. The recording store is an admission control, not proof of Cosmos transport interception or store-context identity. Separately test an injected CosmosClient under strict yields ExecutorSizeCapabilityException with zero writes; non-strict retains existing named unvalidated diagnostic. Do not inject a client into the supported over-quota fixture, require a supported-context transport recorder or add a production HTTP interception hook. Preserve existing bypass/upstream-binding/conflict/in-doubt semantics.
- AC7: Lifecycle tests include successful first measurement followed by disposal then Unavailable with creation count1; fresh-disposed and repeated no-resurrection controls; concurrent first-use instance/count; injected ownership; pure measurement no database/container work. Document deliberate option capture at first client use (measurement or write), earlier than previous database initialization if mutable options change. Default supported serialization settings are fixed by existing constructors; do not add unreachable owned-custom-serializer positive tests or reflection-only application safety proofs.
- AC8: Record actual parity evidence for all three writers and the shared mapper. Offline point item POST capture is labelled point-only. Pinned SDK3.57.1 materialization evidence shows serializer bytes become ItemBatchOperation.ResourceBody unchanged; any new probe must be bounded with exact config/commands/output. Reflection permitted only in disposable pinned-SDK characterization, not production or sole application safety tests. Production AllowBulkExecution remains true; no transport envelope certification or Direct/Gateway test mislabelled as a bulk toggle.
- AC9: EN/JA11_storage_providers documents opt-in scope, supported owned vs injected Unavailable, strict/non-strict, application quota2,000,000 and smaller quotas, CertifiedBound slack, explicit-context caller duty and earlier options capture. No service safety-margin/inclusive admission guarantee, stored-footprint/server-added-property guarantee, point/bulk envelope guarantee, or tag/claim/aggregate-operation coverage. Focused tests and legacy/ABI/DI controls pass on net9/net10; required applicable CI and exact-head Opus5 approval precede individual merge/full closeout.
- AC10: Confine shipped changes to Cosmos provider, focused tests and EN/JA storage docs; new provider API additions named in AC5 are allowed, existing signatures and Core ABI remain unchanged. No release, data migration, Core storage-plan contract, stream-write conversion, retry stop/outbox, event splitting/reference transport,1169/1193 or unrelated cleanup. Temporary verification mutations are allowed only as restored disposable evidence, absent from final diff.

## Reviewed Counting Seam
Use a test-local CountingEventTypes implementing IEventTypes and delegating to the existing domain EventTypes, supplied through CosmosDbEventStore existing constructor. Count SerializeEventPayload calls: N typed events require N calls; a duplicate ToSerializableEvent mutation yields2N. A first-call throwing serializer must retain original error channel and count1. Do not conflate domain payload serialization with Cosmos document serialization. This specifies the existing instrumentation requirement and adds no production API.

tomohisa:
Closes #1218

## Summary
- add the shared Cosmos event-document mapper and provider-owned serializer boundary
- add certified Cosmos StorageItem measurement, strict/non-strict capability handling, eager validation, and all-five-entry pre-write gate coverage
- harden CosmosDbContext lazy client/serializer lifecycle without database initialization during measurement
- document the bounded Cosmos-only scope in EN/JA storage guides

## Verification

```text
NUGET_HTTP_CACHE_PATH=/private/tmp/sek-g72-http-cache NUGET_PACKAGES=/private/tmp/sek-g72-nuget dotnet test dcb/tests/Sekiban.Dcb.WithResult.Tests/Sekiban.Dcb.WithResult.Tests.csproj -c Release -f net9.0 --no-restore --filter "FullyQualifiedName~CosmosEventDocumentSizeGateTests|FullyQualifiedName~CosmosExclusiveAfterPositionTests|FullyQualifiedName~CosmosCrashWindowE2ETests|FullyQualifiedName~ExecutorSizeGateCompatibilityTests"
Passed: 34, Failed: 0, Skipped: 0

NUGET_HTTP_CACHE_PATH=/private/tmp/sek-g72-http-cache NUGET_PACKAGES=/private/tmp/sek-g72-nuget dotnet test dcb/tests/Sekiban.Dcb.WithResult.Tests/Sekiban.Dcb.WithResult.Tests.csproj -c Release -f net10.0 --no-restore --filter "FullyQualifiedName~CosmosEventDocumentSizeGateTests|FullyQualifiedName~CosmosExclusiveAfterPositionTests|FullyQualifiedName~CosmosCrashWindowE2ETests|FullyQualifiedName~ExecutorSizeGateCompatibilityTests"
Passed: 34, Failed: 0, Skipped: 0

NUGET_HTTP_CACHE_PATH=/private/tmp/sek-g72-http-cache NUGET_PACKAGES=/private/tmp/sek-g72-nuget dotnet build dcb/src/Sekiban.Dcb.CosmosDb/Sekiban.Dcb.CosmosDb.csproj -c Release -f net9.0 --no-restore
Build succeeded. 0 Warning(s), 0 Error(s)

NUGET_HTTP_CACHE_PATH=/private/tmp/sek-g72-http-cache NUGET_PACKAGES=/private/tmp/sek-g72-nuget dotnet build dcb/src/Sekiban.Dcb.CosmosDb/Sekiban.Dcb.CosmosDb.csproj -c Release -f net10.0 --no-restore
Build succeeded. 0 Warning(s), 0 Error(s)
```

`git diff --check` passed. The bounded local tests exercise the provider-owned `Microsoft.Azure.Cosmos.CosmosClient` serializer path but do not initialize a live Cosmos database; live service acceptance remains CI/environment evidence.


tomohisa (commented):
Exact-head Opus5 review of `014995e8ad0b63b0df1b9ac51b7e4f4052c1c8e8` against #1218 AC1–AC10 (head rechecked immediately before posting: unchanged; CI on this head fully green).

## Verdict: request-update (blocking)

The size-admission feature itself is well built and satisfies most of AC1–AC10. It is blocked by one defect introduced by the supporting refactor, not by the gate.

### B1 — BLOCKING: `CosmosProviderDefaultSerializer.FromStream` leaves the input stream open, so every deserialization on a provider-owned client throws

`CosmosProviderDefaultSerializer.cs:30-36` builds its `StreamReader` with `leaveOpen: true` and never disposes the incoming stream. The Cosmos SDK wraps every user-supplied `CosmosSerializer` in `CosmosJsonSerializerWrapper`, which requires the user serializer to dispose the response stream. `CosmosDbContext.EnsureClientCreatedLocked` installs this serializer for **every** connection-string context — the production path — replacing `SerializerOptions = new CosmosSerializationOptions { PropertyNamingPolicy = CamelCase }`.

Measured offline against the **actually compiled** type from this branch's `Sekiban.Dcb.CosmosDb.dll`, installed on a real `CosmosClient` (`AllowBulkExecution = true`, fake local connection string, no network, no live service):

```
REAL type loaded: Sekiban.Dcb.CosmosDb.CosmosProviderDefaultSerializer
REAL FromStream THREW:          InvalidOperationException: Json Serializer left an open stream.
REAL CreateItemResponse THREW:  InvalidOperationException: Json Serializer left an open stream.
```

Control, same probe with the configuration this PR replaces: `CONTROL (SerializerOptions CamelCase) FromStream OK`.

Blast radius:
- All 33 typed Cosmos calls in the provider (`ReadItemAsync<…>`, `GetItemQueryIterator<…>`, `DeleteItemAsync<…>` over `CosmosEvent` / `CosmosTag` / `CosmosMultiProjectionState` / `long` / `string` / `dynamic`) deserialize through this boundary.
- `CosmosResponseFactoryCore.CreateItemResponse<T>` materialises eagerly, so `CreateItemAsync(cosmosEvent, …)` at `CosmosDbEventStore.cs:72`, `:808`, `:1505` writes the item to the service and then throws `InvalidOperationException` — a write that succeeds on the server and fails to the caller — which also defeats the existing `catch (CosmosException ex) when (ex.StatusCode == HttpStatusCode.Conflict)` classification, since the thrown type is not `CosmosException`.

Why CI does not see it: no test in the repo constructs a real `CosmosClient` round trip. `Cosmos/InMemoryCosmosClient.cs` is always *injected* (`_ownsCosmosClient == false`), so the gate correctly reports Unavailable and `CosmosProviderDefaultSerializer` is only ever exercised through `ToStream`. `FromStream` has zero coverage.

Fix — mirror the SDK's own `CosmosJsonDotNetSerializer.FromStream`:

```csharp
public override T FromStream<T>(Stream stream)
{
    ArgumentNullException.ThrowIfNull(stream);
    using (stream)
    {
        if (typeof(Stream).IsAssignableFrom(typeof(T)))
        {
            return (T)(object)stream;
        }

        using var reader = new StreamReader(stream, Encoding.UTF8, detectEncodingFromByteOrderMarks: true);
        using var jsonReader = new JsonTextReader(reader);
        return JsonSerializer.Create(Settings).Deserialize<T>(jsonReader)!;
    }
}
```

Discriminating test: `FromStream_DisposesTheInputStream` — after `FromStream<CosmosEvent>(ms)`, assert `ms.CanRead == false`. `leaveOpen: true` is exactly the mutant it kills. Add a `CosmosEvent` round-trip through the client's serializer boundary as the behavioural companion.

### B2 — BLOCKING (docs/behaviour): the serializer swap is unconditional and changes non-UTC `DateTime` rendering

EN/JA docs say the feature "does not change legacy writes when the gate is not registered", but the serializer replacement happens for every connection-string context regardless of gate registration. Measured old-config vs new-serializer output for the same POCO:

| `DateTimeKind` | previous config | this PR |
|---|---|---|
| `Utc` | `"2020-01-01T00:00:00.1234567Z"` (100 B) | identical (100 B) |
| `Local` | `"2020-01-01T00:00:00.1234567-08:00"` (105 B) | `"2020-01-01T08:00:00.1234567Z"` (100 B) |
| `Unspecified` | `"2020-01-01T00:00:00.1234567"` (99 B) | `"2020-01-01T00:00:00.1234567Z"` (100 B) |

Property naming is unaffected (all shipped models use explicit already-camelCase `[JsonProperty]` names, `_etag` included). Every current `DateTime` writer in the provider is UTC (`CosmosEvent.Timestamp`, `CosmosTag.CreatedAt` via `SpecifyKind(..., Utc)`, `CosmosMultiProjectionState.CreatedAt/UpdatedAt`), so **stored bytes are unchanged today** — but `DateTimeZoneHandling.Utc` also now coerces read-back values to UTC kind instead of Json.NET's `RoundtripKind`, and AC1 asks for no stored timestamp semantics change.

Fix: correct the docs sentence, state the `DateTimeZoneHandling.Utc` contract on both directions, and add a parity test asserting `ToStream` output for `CosmosEvent` / `CosmosTag` / `CosmosMultiProjectionState` is byte-identical to the previous `CosmosSerializationOptions.CamelCase` configuration for UTC inputs.

### F3 — test strength: the CertifiedBound slack oracle cannot fail

In `CertifiedBound_CoversProviderSerializationForFractionAndYearBoundaries` (`:151-200`), `bound` is derived from the event built by `NewEvent`, whose metadata is `("cause","correlation","user")`, while the comparison loop builds documents with `EventMetadata(null!, null!, null!)` — roughly 14 bytes shorter regardless of the timestamp. And `MeasurementTimestampUtc` already has 7 non-zero fractional digits, so the sentinel is already the widest rendering. `Assert.True(slack >= 0)` therefore passes for any `TimestampSlackBytes >= -14`; setting it to `0` or deleting it does not fail the test. (The bound itself is sound — only its oracle is inert.)

Secondary: `[InlineData(0/1/3/7)]` are **ticks**, not fractional digits; Json.NET trims trailing zeros, so 1, 3 and 7 all render 7 digits and AC4's 0/1/3/7 widths are never exercised.

Fix: use `serialized.EventMetadata` in the loop; add `Assert.Equal(bound - 8, maxObservedLength)` to pin both sentinel maximality and the slack constant; use ticks `0`, `1_000_000` (`.1`), `1_230_000` (`.123`), `1_234_567` (`.1234567`).

### F4 — minor: AC2's required statement that general in-flight database I/O shutdown is not solved appears nowhere in the shipped diff (code XML docs or EN/JA pages).

### F5 — minor: `FromParts` adds `ThrowIfNullOrWhiteSpace` on `sortableUniqueId` and `eventType`, which the public `CosmosEvent.FromEvent` did not previously perform. The three original guards are correctly preserved and fire in the original order, so this is a strengthening — but it is an undeclared public behaviour change. Either declare it deliberate with an `Assert.Throws` test on the public entry, or confine those two guards to the internal callers.

### F6 — minor: AC6 asks for a recording publisher and a zero publisher-enqueue assertion; the tests supply recording stores only.

## Verified correct (no change requested)

- **AC1**: `FromParts` is the single core; both adapters are thin; all three write sites (`:64` conditional, `:801` typed, `:1499` serialized) route through it; `Pk`/`Id` formatting, payload/metadata/tags, `CreateItemAsync` and Conflict classification unchanged; `claimEvent with { Id = deterministicId }` reproduces the previous conditional document without mutating the caller's record; non-UTC rejection lives in the core with both `Local` and `Unspecified` asserted.
- **Counting seam**: `CountingEventTypes` supplied through the existing store constructor; typed write asserts `SerializeCalls == 1` (a duplicate `ToSerializableEvent` mutant gives 2) and the failure test asserts the first-call throw surfaces once with an empty container. No new production API.
- **AC2**: one `_lifecycleLock` monitor over creation, measurement, container-cache lookups and `Dispose(bool)`; `EnsureClientCreatedLocked` checks `_disposed` before the existing-client fast return (L1 fixed); `InitializeAsync` releases the monitor before `CreateDatabaseIfNotExistsAsync` and re-checks before assigning `_database`; injected clients never disposed; 16 concurrent measures → `ClientCreationCount == 1`; dispose → named Unavailable, count stays 1.
- **AC3**: named Unavailable reasons for disposed / injected / missing-registration / wrong-representation; monitor held across serialization and stream disposal; `CanSeek` guard; streams disposed on all paths; `OperationCanceledException` excluded from the catch; pure creation does no database or container work.
- **AC4**: `CertifiedBound` only, never `Exact` (`Bytes` asserted null); specified sentinel, never written; identity/tag mismatch → named Unavailable; `checked(...)` on the addition.
- **AC5**: default and maximum `2_000_000`; `0` and `2_000_001` throw before any service/policy mutation (`Assert.Empty`); G68 marker mixing rejected in both orders with descriptor count unchanged; DI resolves the same singleton (`Assert.Same`); missing context yields a policy that reports Unavailable rather than failing option resolution.
- **AC6**: all five entrances yield `ExecutorSizeLimitExceededException` with zero write/conditional/expected-position invocations (the store throws if reached, so this is an invocation assertion, not "empty store"); `IsCertifiedUpperBound == true` asserted; injected client Strict → `ExecutorSizeCapabilityException` with zero writes, NonStrict → success with a named diagnostic and one write.
- **AC9/AC10**: EN and JA both carry scope, owned-vs-injected, strict/non-strict, quota, slack, caller duty, earlier options capture and the exclusions, with no safety-margin claim; diff confined to Cosmos provider, focused tests and EN/JA docs; no leftover verification mutations.

Live Cosmos acceptance is treated as CI/environment evidence only; none was available to me and none was attempted. All probes were disposable and offline, no production reflection is proposed, and no source, label or PR state was mutated by this review.

tomohisa (commented):
Fresh exact-head re-review of `02bab945d50b79e11e6e52a07e95f77d92d72606` against #1218 AC1–AC10, from a new detached checkout with nothing inherited from the previous head. Head and CI rechecked immediately before posting: unchanged, and CI is now fully green (`dcbTestsNet9`, `dcbTestsNet10`, templates, PostgreSQL consumers, SonarCloud all SUCCESS).

## Verdict: request-update — narrow, and non-blocking in effect

**All five previously blocking items are closed and verified by measurement.** Three small items remain; none breaks current behaviour and each is a one-to-few-line change.

### Closed

**B1 — response-stream disposal.** `using (stream)` replaces `leaveOpen: true`. Re-ran the same probe against the **newly compiled** assembly for this head (mvid `b9bbbfb1-05d5-47a6-9f19-e2791953d280`) on a real `CosmosClient` (`AllowBulkExecution = true`, offline):

```
[1] SerializerCore.FromStream<Poco> OK name=a inputCanRead=False
[2] CreateItemResponse OK resource=a
[6] ToStream CanSeek=True Length=40
```

Both paths that threw `Json Serializer left an open stream.` at `014995e8` now pass through `CosmosJsonSerializerWrapper`, so `CreateItemAsync` no longer writes-then-throws and the `CosmosException`/`Conflict` classification is intact. Minor accuracy nit: `ProviderSerializer_ThroughCosmosClientOptions_DisposesResponseStream` does not actually traverse the wrapper — `client.ClientOptions.Serializer` returns the raw user serializer (the test's own `Assert.Same` proves it) — so the `CosmosClient` construction is decorative. The `Assert.False(stream.CanRead)` oracle is still exactly the wrapper's condition and kills the mutant, so this is a naming issue, not a defect.

**B2 — UTC-only parity and docs.** The misleading "does not change legacy writes" sentence is gone; EN/JA now state the serializer applies whether or not the gate is registered, scope byte-identity to the current three UTC writers, and declare `DateTimeZoneHandling.Utc` deliberate with no non-UTC guarantee. I independently verified the parity test's control is byte-faithful to the **actual** SDK serializer using a POCO exercising every differing setting: `replica == real SDK : True`. `Newtonsoft.Json 13.0.4` default `MaxDepth` is 64, matching the SDK's explicit value, so that is not a delta either.

**F3 — CertifiedBound oracle.** Metadata now matches the bound and the inline data genuinely render 0/1/3/7 fractional digits. Mutation-verified in a disposable checkout (all restored):

| mutation | focused suite |
|---|---|
| `TimestampSlackBytes` 8 → 0 | Failed 4/18 |
| sentinel `AddTicks(1_234_567)` → `AddTicks(0)` | Passed 18/18 |
| both | Failed 4/18 |
| slack → −10, bound assertion neutralised so only the loop can catch it | Failed 4/4 |

Row 2 is correct, not a gap — a whole-second sentinel with 8 bytes of slack is still an exactly tight bound. Row 4 shows the loop alone now catches an undersized slack, which the old null-metadata cushion masked.

**F4** — the in-flight-shutdown limitation is stated in `Dispose(bool)`'s XML doc and on both doc pages. **F5** — `FromEvent` routes to the new private `FromPartsCore`, bypassing the whitespace checks, with a compatibility test asserting an `Event` with empty `SortableUniqueIdValue`/`EventType` still produces a document; the three original guards remain in order, and `FromPartsCore` is still the single construction site holding the non-UTC rejection.

### R1 — the provider serializer is not name-equivalent to the SDK serializer

`DefaultContractResolver { NamingStrategy = new CamelCaseNamingStrategy() }` defaults to `OverrideSpecifiedNames = false` and `ProcessDictionaryKeys = false`; the SDK's `CamelCasePropertyNamesContractResolver` sets both `true`. Measured:

```
REAL SDK : … "someUpper":"u" … "map":{"keyOne":"v"} …
PROVIDER : … "SomeUpper":"u" … "map":{"KeyOne":"v"} …
```

No shipped model is affected — all `[JsonProperty]` names are already camelCase (`_etag` is unchanged either way) and none has a dictionary, which is exactly why the parity test passes. But the parity test covers only today's three models, so the first model added with a non-camelCase `[JsonProperty]` name or a `Dictionary` property silently diverges from every document written before this PR, with no failing test.

Fix (measured safe for all three current models):

```csharp
NamingStrategy = new CamelCaseNamingStrategy(processDictionaryKeys: true, overrideSpecifiedNames: true)
```

Discriminating test: extend the parity fixture with a POCO carrying `[JsonProperty("SomeUpper")]` and `Dictionary<string,string> { ["KeyOne"] = "v" }`, compared against `PreviousSdkCamelCaseSerializerBytes`. Fails today, passes with the fix.

### R2 — `FromStream` still lacks the SDK's `Stream` passthrough (minor)

`CosmosSerializerCore` does not special-case `T = Stream`; it delegates, and the SDK's own serializer short-circuits it. Measured at this head vs the replaced configuration:

```
[3] SerializerCore.FromStream<Stream>          THREW: JsonSerializationException: Could not create an instance of type System.IO.Stream …
[3-CONTROL old SDK config] FromStream<Stream>  OK -> MemoryStream
```

Nothing in-repo asks for `T = Stream`, but `CosmosDbContext` exposes `GetEventsContainerAsync` / `GetTagsContainerAsync` / `GetMultiProjectionStatesContainerAsync` publicly (`:171`, `:177`, `:183`), so a consumer can reach it. Fix: `if (typeof(Stream).IsAssignableFrom(typeof(T))) { return (T)(object)stream; }` as the first statement inside `using (stream)`. Test: `Assert.Same(stream, serializer.FromStream<Stream>(stream))`.

### R3 — the publisher zero-enqueue assertion cannot fail

`RecordingEventPublisher` is correctly wired (the 5-arg call binds to the size-gate constructor at `GeneralSekibanExecutor.cs:50`, whose fifth parameter is `IEventPublisher?`), but `Assert.Empty(publisher.PublishedEvents)` passes whether the gate suppressed the publish or the publisher was never connected. Mutation-proven — removing `publisher` from the executor construction leaves the test green:

```
### publisher NOT wired to the executor (mutant):
Passed!  - Failed: 0, Passed: 1, Skipped: 0, Total: 1
```

So F6's closure is nominal. Fix: add a positive control using the same executor construction — an under-quota commit that succeeds and asserts `PublishedEvents` is non-empty with matching event ids. With that control present the mutant fails.

### R4 — cosmetic

`Assert.Equal(maxObservedLength, bound - minimumSlack)` is a tautology (`bound − min(bound − len) ≡ max(len)`) and asserts nothing. Its neighbour `Assert.Equal(expectedMaxStream.Length, maxObservedLength)` *is* meaningful — it pins timestamp-width independence across the year 1/2000/9999 boundaries. Suggest deleting the tautology or replacing it with `Assert.True(maxObservedLength <= bound - 8)`.

### R5 — observation only

`CosmosEvent.FromEvent` now throws `ArgumentNullException` for a null `ev.Tags` where base `a69e1a3d` produced a document with `Tags = null`. The throw is inherent to `FromPartsCore`'s shared `tags` guard, so the explicit `ThrowIfNull(ev.Tags)` only fixes the parameter name. Recorded so the AC1 "argument validation preserved" claim is exact.

## Re-derived and still holding

`CosmosDbEventStore.cs`, `CosmosEventDocumentSizeMeasurement.cs`, `CosmosExecutorSizeGateExtensions.cs` and `Models/CosmosEvent.cs` are byte-identical to `014995e8`; re-read at this head, AC1 (single core, three writers, one typed serialization, ID/pk/`CreateItemAsync`/Conflict preserved, `Local` and `Unspecified` rejected), the `CountingEventTypes` seam with its first-call-throws control, AC2 (single monitor, disposed-before-every-return, database work outside the monitor, concurrent count 1, no resurrection, injected client never disposed), AC3 (named Unavailable set, `CanSeek` guard, streams disposed, no database work on pure creation), AC4 (`CertifiedBound` only, sentinel never written, `checked`), AC5 (quota bounds, eager validation with no partial registration in both G68 marker orders, same singleton), AC6 (five entrances, `IsCertifiedUpperBound`, zero store invocations, strict/non-strict injected controls), AC8 and AC10 all hold.

Focused class locally: `Passed! - Failed: 0, Passed: 18, Total: 18` (net9, Release). All probes were disposable and offline; every mutation was restored and the checkout's `git status --porcelain` is empty. No source, label, PR-state, merge, release or #1217 mutation was performed.

tomohisa (commented):
Final exact-head review of `f15731c1264d206c122980b33e31895ea0e82032` against #1218 AC1–AC10, from a fresh detached checkout with nothing inherited from the earlier heads. Head and CI rechecked after all checks settled and immediately before posting: head unchanged, OPEN/MERGEABLE, and CI **all green** (`dcbTestsNet9`, `dcbTestsNet10`, templates, PostgreSQL consumers, SonarCloud, SonarCloud Code Analysis all SUCCESS).

## Verdict: approve

R1, R2 and R3 are closed, and each is proven by a mutation the shipped tests kill. B1, B2, F3, F4 and F5 all still hold. Production change at this head is 8 lines in one file; everything else in the PR is byte-identical to `02bab945`.

### R1 — name parity with the actual SDK serializer

`CamelCaseNamingStrategy(processDictionaryKeys: true, overrideSpecifiedNames: true)` now matches the two flags `CamelCasePropertyNamesContractResolver` sets. Verified against the **real** SDK serializer via `CosmosClient.ClientContext.SerializerCore` (not the test's replica), using the newly compiled assembly for this head (mvid `f12703d7-f10b-4c22-a14a-9114fef11487`), offline:

```
SDK : …"someUpper":"u"…"map":{"keyOne":"v"}…
PROV: …"someUpper":"u"…"map":{"keyOne":"v"}…
R1 identical: True
```

Both divergences measured at `02bab945` are gone. I separately checked the case where `overrideSpecifiedNames: true` could plausibly have caused damage — the shipped document shape with already-camelCase `[JsonProperty]` names plus `_etag`: identical to the SDK, `_etag` preserved.

Mutation-verified: `overrideSpecifiedNames` → `false` fails 1/21; `processDictionaryKeys` → `false` fails 1/21.

### R2 — SDK-compatible `Stream` passthrough, including disposal semantics

The passthrough sits as the first statement inside `using (stream)` — the same placement as the SDK's own `CosmosJsonDotNetSerializer`, so both return a disposed stream. Measured at the real `CosmosSerializerCore` boundary:

```
[PROVIDER(head)]          FromStream<Stream> -> MemoryStream  same=True  canRead=False
[CONTROL(old SDK config)] FromStream<Stream> -> MemoryStream  same=True  canRead=False
```

Exact parity on instance identity, type and `CanRead`, which also keeps `CosmosJsonSerializerWrapper`'s "left an open stream" check satisfied. Public reachability is unchanged (`GetEventsContainerAsync` `:171`, `GetTagsContainerAsync` `:177`, `GetMultiProjectionStatesContainerAsync` `:183`). Mutation-verified: removing the passthrough fails 1/21.

### R3 — discriminating positive publisher control paired with the five zero-enqueue paths

`ProviderOwnedCertifiedBound_AllowsUnderQuotaAndPublishesOnce` uses the *same* construction shape as the rejection test (`new GeneralSekibanExecutor(store, accessor, domain, options, publisher)` → the size-gate constructor at `GeneralSekibanExecutor.cs:50`, fifth parameter `IEventPublisher?`), with a provider-owned Cosmos measurement at the full `2_000_000` quota, and asserts success plus `Assert.Single(publisher.PublishedEvents)` and the payload type. Mutation-verified: **dropping `publisher` from this control fails 1/21** — exactly the discrimination missing at `02bab945`.

Recorded for accuracy, not as a requested change: the `Assert.Empty` inside the rejection test is still non-discriminating in isolation (I measured that dropping `publisher` from that test alone leaves the suite green, since an unwired publisher is also empty). That is the normal limitation of a negative assertion, and it is precisely what the new positive control exists to anchor. Nothing further is needed.

### Earlier fixes re-confirmed

**B1** — re-probed on this head's assembly: `SerializerCore.FromStream<Poco> OK … inputCanRead=False` and `CreateItemResponse OK`, so no write-then-throw and the `CosmosException`/`Conflict` classification is intact. Mutation: restoring `leaveOpen: true` fails 2/21.

**B2** — EN/JA docs unchanged and still accurate; with R1 fixed the parity is broader than the docs claim, so the wording is conservative rather than wrong. **F3** — the tautology is replaced by `Assert.True(maxObservedLength <= bound - 8)`, which pins that the sentinel is genuinely the maximum-width rendering. Mutation table at this head: `TimestampSlackBytes` 8→0 fails 4/21, and sentinel `AddTicks(1_234_567)`→`AddTicks(0)` now fails 3/21 — that second mutant **survived** at `02bab945`, so this is a real strengthening. **F4** and **F5** unchanged and holding.

### ACs re-derived

`CosmosDbContext.cs`, `CosmosDbEventStore.cs`, `CosmosEventDocumentMapper.cs`, `CosmosEventDocumentSizeMeasurement.cs`, `CosmosExecutorSizeGateExtensions.cs`, `Models/CosmosEvent.cs` and both doc pages are byte-identical to `02bab945`. Re-read at this head: AC1 (single `FromPartsCore`, three routes, one typed serialization, ID/pk/`CreateItemAsync`/Conflict preserved, `Local` and `Unspecified` rejected, public `FromEvent` preserved), the `CountingEventTypes` seam with its first-call-throws control, AC2 (single monitor, disposed-before-every-return, database work outside the monitor, concurrent count 1, no resurrection, injected client never disposed, in-flight limitation stated), AC3 (named Unavailable set, `CanSeek` guard, streams disposed, no database work on pure creation), AC4 (`CertifiedBound` only, sentinel never written, `checked`, 0/1/3/7 widths and year boundaries), AC5 (quota bounds, eager validation with no partial registration in both G68 marker orders, same singleton), AC6 (five entrances, `IsCertifiedUpperBound`, zero store invocations and zero enqueue, injected strict/non-strict controls), AC8, AC9 and AC10 all hold.

### Test evidence

Focused class locally, Release: **net9.0 21/21**, **net10.0 21/21**. Mutation battery of 7 mutants, all restored: every one of R1×2, R2, B1, F3×2 and the R3 positive control is killed; the single deliberate non-kill is documented under R3. A full local `Sekiban.Dcb.WithResult.Tests` net9 run (at the immediately preceding head) was 1375 passed / 6 failed, all six being `DynamoDbLocal*` pinned-emulator tests that need a running DynamoDB Local endpoint this machine does not have — unrelated to Cosmos and to this PR; required CI covers them and is green.

All probes were disposable, offline and reflection-only against the pinned SDK; no production reflection is introduced. Every mutation was restored and the checkout's `git status --porcelain` is empty apart from build output. Live Cosmos acceptance was neither available nor attempted and is treated as CI/environment evidence only. No source, label, PR-state, merge, release or #1217 mutation was performed by this review.

tomohisa (commented):
Exact-head re-review of `ad7153179af265e5294402cb1e235b228955f8bc` after the accepted A1/A2 repair (Option 1), from a fresh detached checkout with nothing inherited. Requested CI run **34536047574** at this head is `completed/success` with `dcbTestsNet9` and `dcbTestsNet10` both success; full rollup rechecked immediately before posting is all green (templates, PostgreSQL consumers, SonarCloud ×2). Head unchanged, OPEN/MERGEABLE.

## Verdict: request-update — scoped to test coverage. The production repair itself is correct.

Disposition items 1, 2, 4 and 6 are implemented faithfully and I verified each by measurement. Item 3 is implemented by half — the missing-capability simulation seam exists, the *effective serializer observation* accessor does not — and as a direct consequence item 5's Local/Unspecified public-container discriminator points at a test-local client rather than the context's own.

### F1 — no test observes the context's effective serializer, so the A2 regression is undetectable

Every serializer test builds its own client through the private helper `NewSdkClient()` (test `:723`), and `SdkSerializerBytes` (`:745`) constructs one per call. No test asks the `CosmosDbContext` what serializer its owned client actually got. The Local/Unspecified theory therefore reduces to "the real SDK serializer agrees with the Json.NET replica control" — a fidelity check of the control, not a compatibility check of the provider.

Mutation-measured in a disposable checkout (all restored):

| mutation | focused suite (24 tests) |
|---|---|
| **M1** re-install a custom serializer on the owned client (camelCase with both naming flags, `DateTimeZoneHandling.Utc`, correct disposal and `Stream` passthrough) | **Passed 24/24** |
| **M1b** re-install a custom serializer reproducing the **original B1 defect** (`leaveOpen: true`, no disposal, no passthrough, no name override, `Utc`) | **Passed 24/24** |

M1b is the alarming one: the serializer that made every Cosmos read and every `CreateItemAsync` response throw `Json Serializer left an open stream.` can be put back on the production path with no test failing.

**Correction** — add the small internal accessor disposition item 3 preferred, with the same guard shape as `TryMeasureSupportedDocument` (internal, instance-local, no global override, no ownership bypass, cannot install a serializer):

```csharp
internal CosmosSerializer? EffectiveMeasurementSerializer
{
    get
    {
        lock (_lifecycleLock)
        {
            if (_disposed || !_ownsCosmosClient) return null;
            EnsureClientCreatedLocked();
            return _cosmosClient!.ClientOptions.Serializer;
        }
    }
}
```

**Discriminating tests** — D1: assert the accessor is non-null and `Assert.NotEqual(typeof(CosmosDbContext).Assembly, serializer.GetType().Assembly)`, which asserts what the repair actually promises without pinning the SDK's internal wrapper type (kills M1 and M1b). D2: redirect the Local/Unspecified theory to the accessor, keeping `ExpectedSdkCamelCaseSerializerBytes` as the control — this is item 5's discriminator, and it only bites after redirection. D3: redirect the UTC-writers and explicit-name/dictionary-key parity tests. D4: redirect the stream-disposal and `Stream`-passthrough tests, restoring the B1/R2 fences to the production path. D5: accessor guards — injected → null with `ClientCreationCount == 0`; disposed → null, no resurrection, count unchanged; concurrent first access → one client. Keep `NewSdkClient()` only where an independent SDK control is intended, and pass the serializer into `SdkSerializerBytes` so each call site's intent is visible.

### N1 — small: the null-serializer capability is not exercised at executor level

Item 2 singles out the null case ("Strict capability failure before store dispatch; NonStrict named diagnostic"), but `ForceMeasurementSerializerUnavailable` is asserted only at `Measure` level, while the executor-level Strict/NonStrict theory still uses the injected client alone. The paths share downstream code so behaviour is identical today; adding the forced-unavailable owned context as a second row of that theory pins it.

### Observations (no change requested)

**N2** `CertifiedBound_EqualQuotaPasses_AndOneLessRejectsBeforeDispatch` derives `bound` from the real measurement but drives the executor with `FixedCertifiedMeasurement`, isolating the evaluator boundary — correct to isolate, but it would still pass if the Cosmos measurement were broken. **N3** "full-fraction equality" holds by composition rather than directly: the exact `Assert.Equal(sentinelBytes.LongLength, bound)` plus `Assert.InRange(slack, 0, 8)` force slack 0 at the 7-digit row; an explicit `Assert.Equal(0, slack)` there would make it self-evident. **N4** `ForceMeasurementSerializerUnavailable` is written without the monitor while read under it — benign for a test-only instance-local flag, but a lock-guarded setter would match the class's discipline.

## Verified correct

**Item 1** — `CosmosProviderDefaultSerializer.cs` and the `_serializer` field are deleted; `SerializerOptions { CamelCase }` is restored with bulk/retry/connection untouched; the measurement reads the **constructed** client's `ClientOptions.Serializer`, never the pre-construction object (I measured that the caller's own options object stays null while the client's clone carries the serializer). The context's production default is **Direct** mode (`UseDirectConnectionMode = true`), which the tests never build, so I measured that shape:

```
Direct  ClientOptions.Serializer = CosmosJsonSerializerWrapper
Direct == Gateway  : True
Direct == TestShape: True
public ToStream == SerializerCore ToStream: True
sample: "timestamp":"2042-03-04T05:06:07.1234567-0…
```

So the capability is available offline under the mode production uses, the cross-client comparison the tests rely on is sound, the measured bytes are the bytes the write path produces, and the `-08:00` shows the non-UTC drift is gone from production. No production reflection, no replica fallback, no serializer injection.

**Item 2** — `serializer is null` yields the named capability Unavailable with no alternative byte source anywhere on the path; the injected refusal is retained and carries both phrases so the existing Strict/NonStrict assertions still bind. Mutation: dropping the injected refusal fails 2/24.

**Item 3 bounds** — the seam is internal, instance-local, defaults false, forces only failure, cannot install a serializer or bypass ownership/disposal; `InternalsVisibleTo` is limited to `Sekiban.Dcb.WithResult.Tests`; no public API added.

**Item 4** — `TimestampSlackBytes` is gone and `Measure` returns the sentinel bytes with no arithmetic. Tests assert bound == sentinel bytes exactly, `InRange(slack, 0, 8)` across 0/1/3/7 fractional digits and years 1/2000/9999 with per-row timestamp-rendering equality against the SDK control, year independence, and quota `== bound` admits with one write while `== bound - 1` rejects with `MeasuredBytes`/`LimitBytes` asserted and zero writes. `CertifiedBound` only, never `Exact`. Mutations: `+4` slack fails 4/24; non-maximal sentinel fails 3/24. The former `+8` assertions are gone, so the corrected contract is what is enforced.

**Item 6** — EN and JA drop the non-UTC disclaimer and the explicit-serializer text and now state the constructed-client measurement, the no-`+8` bound, the `0..8` overshoot, Strict/NonStrict, "never replaced by a fallback serializer", the opt-in gate with restored SDK options when unregistered, and item 5's separation between the mapper's UTC write contract and public compatibility fixtures. Exclusions and the in-flight shutdown limitation are retained.

**ACs** — `CosmosDbEventStore.cs`, `CosmosEventDocumentMapper.cs`, `CosmosExecutorSizeGateExtensions.cs` and `Models/CosmosEvent.cs` are byte-identical to `f15731c1`; re-read at this head, AC1 through AC10 all hold (single construction core and three routes, the `CountingEventTypes` single-serialization seam with its first-call-throws control, the monitor/disposal/no-resurrection contract, the named Unavailable set with stream guards and pure initialization, the corrected bound, the registration and G68 marker rules, the five entrances with zero store dispatch and zero publisher enqueue paired with the under-quota publish control, and diff confinement with no Core/ABI/release/#1217 content).

Focused class locally: **net9 24/24, net10 24/24**. All probes were disposable, offline against a non-existent endpoint, and reflection-only against the pinned SDK for corroboration; every mutation was restored and the checkout's `git status --porcelain` is empty apart from build output. No source, label, PR-state, merge, release or #1217 mutation was performed.

tomohisa (commented):
Final exact-head review of `3cbf155f7a7ef0eb81192b18fdb64efc5ab5f53b`, from a fresh detached checkout. Nothing was inherited: the PR still carries `intent-pr-approved` from an earlier head, which I ignored and did not act on. Head and CI rechecked immediately before posting — unchanged, OPEN/MERGEABLE, and all checks green (`dcbTestsNet9`, `dcbTestsNet10`, templates, PostgreSQL consumers, SonarCloud ×2).

## Verdict: APPROVE

F1 (D1–D5) and N1 are closed, proven by mutation rather than by reading: the two mutants that passed the whole suite at `ad715317` now fail. Production change at this head is the 20-line observation seam and nothing else — `EnsureClientCreatedLocked`, `TryMeasureSupportedDocument`, the mapper, the measurement, the registration helpers, the event store and both doc pages are byte-identical to `ad715317`.

### F1 — closed

`EffectiveMeasurementSerializer` is exactly the accessor disposition item 3 preferred: internal, instance-local, same monitor, disposal and ownership checked before any return or creation, cannot install or replace a serializer, cannot resurrect a client, no public API, IVT still limited to `Sekiban.Dcb.WithResult.Tests`. `NewSdkClient()` is deleted and `SdkSerializerBytes` now takes the context, so every serializer assertion names its subject.

Mutation evidence (all restored; worktree clean afterwards):

| mutation | at `ad715317` | at `3cbf155f` |
|---|---|---|
| M1 re-install a well-formed custom serializer on the owned client | Passed 24/24 | **Failed 3/27** |
| M1b re-install a serializer reproducing the **original B1 defect** (`leaveOpen`, no disposal, no `Stream` passthrough, no name override, `Utc`) | Passed 24/24 | **Failed 7/27** |
| M5 seam drops the ownership guard | — | **Failed 1/27** |
| M6 seam drops the disposed guard | — | **Failed 1/27** |

I also confirmed the property independently rather than trusting the shipped tests, using a disposable in-repo probe (added, run, deleted) that reads the seam on a real context and compares against the exact pre-PR production client configuration, including Direct mode which the tests never build:

```
seam serializer = Microsoft.Azure.Cosmos.CosmosJsonSerializerWrapper
from provider assembly? False        ClientCreationCount = 1
owned  : {"pk":"s|1","someUpper":"u","createdAt":"2042-03-04T05:06:07.1234567-08:00","map":{"keyOne":"v"}}
control: {"pk":"s|1","someUpper":"u","createdAt":"2042-03-04T05:06:07.1234567-08:00","map":{"keyOne":"v"}}
IDENTICAL: True
injected seam = null count=0
after-use dispose: seam = null countBefore=1 countAfter=1
```

So on the production type: the owned client's serializer is the SDK's; a `Local` `DateTime` keeps `-08:00`, `[JsonProperty("SomeUpper")]` becomes `someUpper` and dictionary key `KeyOne` becomes `keyOne`, byte-identical to the pre-PR configuration — public-container compatibility genuinely restored, not merely asserted; an injected context exposes nothing and creates no client; and a context that *had* created a client exposes nothing after disposal with the count unchanged (the after-use no-resurrection case the shipped D1 test does not itself cover). `CosmosProviderDefaultSerializer` is absent from the built assembly.

### N1 — closed

`NullEffectiveSerializerHasStrictAndNonStrictOutcomes` drives the executor with an owned context whose capability is forced unavailable: Strict → `ExecutorSizeCapabilityException` with zero writes; NonStrict → success with the named `"serializer capability is unavailable"` diagnostic and one write. Mutation: ignoring the flag fails 3/27.

### Accepted disposition and ACs

Item 1 (replica deleted, SDK CamelCase restored with bulk/retry/connection untouched, measured from the constructed client), item 2 (null and injected both named Unavailable with no fallback anywhere on the path; dropping the injected refusal fails 2/27), item 3 (seam bounds as above), item 4 (`TimestampSlackBytes` gone, bound == sentinel bytes, `InRange(slack, 0, 8)` across 0/1/3/7 digits and years 1/2000/9999, quota `== bound` admits / `bound - 1` rejects with `MeasuredBytes` and `LimitBytes` asserted and zero writes; `+4` slack fails 4/27 and a non-maximal sentinel fails 3/27), item 5 (T1–T7 now subject the owned context against the real SDK control, with the Local/Unspecified fixtures built directly rather than through the mapper so its separate UTC rejection cannot mask them) and item 6 (docs unchanged and still accurate) all hold.

AC1–AC10 re-derived at this head and all holding, including the single construction core across three routes, the `CountingEventTypes` single-serialization seam with its first-call-throws control, the monitor/disposal/no-resurrection contract, the named Unavailable set with stream guards and pure initialization, the corrected bound, the registration and G68 marker rules in both orders, the five entrances with zero store dispatch and zero publisher enqueue paired with the under-quota publish control, and diff confinement to five Cosmos provider files, one focused test file and the two storage docs — no Core, ABI, release, migration, outbox, #1169/#1193 or #1217 content.

Two observations needing no change: the shipped D1 test proves no-resurrection only for a context disposed before a client existed (I verified the after-use case myself, and the concurrency test already covers it for the measurement path), and `ForceMeasurementSerializerUnavailable` is written without the monitor while read under it — benign for a test-only instance-local flag.

Focused class locally: **net9 27/27, net10 27/27**. Seven mutants, all killed, none surviving from any earlier finding. All probes were disposable and offline; every mutation was restored and `git status --porcelain` is empty apart from build output.

This approval is issued for `3cbf155f7a7ef0eb81192b18fdb64efc5ab5f53b` specifically and must not be read as inherited from the older `intent-pr-approved` label. Merge remains the Architect's action; I performed no source, label, PR-state, merge or release mutation, and no release is implied.

PR #1226, opened 2026-09-12
feat: add durable PostgreSQL-backed Orleans subscriptions

### LINKED ISSUE #1225: SEK-G75 Add PostgreSQL durable Orleans subscriber recovery
# SEK-G75 Add PostgreSQL durable Orleans subscriber recovery

## Goal

Add an opt-in PostgreSQL-backed Orleans durable subscriber that catches up from the event store after
restart or missed notifications, keeps a fenced service-scoped cursor, and gives operators a durable
halt/resume path for poison events.

## Why This Slice Exists Now

Azure Queue delivery is non-rewindable, so the existing process-local callback cannot guarantee recovery
after downtime. SEK-G74 established Orleans 10.3.1 as the 10.22 baseline. This packet now adds the durable
correctness layer; a separate later packet will address atomic notification intent and bounded promptness.

## Current Observed State

The event store already supports indexed reads strictly after a `SortableUniqueId`. Existing Orleans
subscriptions invoke callbacks from non-rewindable stream delivery and keep no durable subscriber cursor.
After downtime or a lost notification, they cannot replay the missing interval.

- Implement from current `main`, after merged SEK-G74 commit
  `c00e37abd48edb7f22eb21a3330a821c64ef06f6`.
- The event store is authoritative. Orleans delivery only wakes the durable runner.
- `FromNow` is the default and uses a guarded safe-tail cursor; `FromBeginning` is explicit opt-in.
- Conversion failures halt immediately. Ordinary handler failures get three retries after the initial
  failure and halt on the fourth recorded failed invocation.
- Existing subscription APIs and behavior remain unchanged for callers that do not opt in.

## Target Repo / Path / Part

- Target paths: `dcb/src/Sekiban.Dcb.Core`, `dcb/src/Sekiban.Dcb.Postgres`,
  `dcb/src/Sekiban.Dcb.Orleans.Core`, `dcb/tests`, `docs/dcb_llm`, `docs/dcb_llm_ja`

Target part: PostgreSQL durable subscriber identity, cursor, leasing/fencing, event-store catch-up,
Orleans nudge integration, bounded failure recovery and operator halt/resume.

## In Scope

- Additive durable-subscription contracts and validated registration.
- PostgreSQL schema/provisioning, atomic initialization, database-clock leases and generation fencing.
- Owner/standby/halted runner states, event-store catch-up and polling fallback.
- Durable retry schedule, immediate conversion halt, bounded handler retry and explicit resume.
- Net9/net10 tests, real PostgreSQL integration and EN/JA documentation.

## Out Of Scope

- The later atomic notification-intent relay and bounded promptness work.
- Stores other than PostgreSQL, exactly-once effects, automatic poison skipping or historical replay by
  default.
- Changing legacy `IEventSubscription` behavior, Pure, samples, templates, #1169 or withdrawn #1193.
- Package publication or the 10.22 release.

Produce one direct-main PR that adds the opt-in durable runner and PostgreSQL store described here. Stored
events determine order and recovery; Orleans notifications only trigger reads. Preserve every acknowledged
cursor under restart, duplicate/lost notifications, lease takeover, handler failure and explicit resume.
Prove the contract with net9/net10 and real PostgreSQL discriminators. Fix review and CI findings with a
concrete test, and do not widen into relay promptness, another provider, legacy migration or release.

## Acceptance Criteria

- AC1: Add opt-in public durable-subscription contracts under Sekiban.Dcb.Subscriptions: normalized service/name identity, FromNow default and FromBeginning start policies, durable phase and runner status, durable state/options, runner/handle and store interfaces. Validate every bound before starting work. Existing IEventSubscription signatures and behavior remain unchanged.
- AC2: Add a service-scoped PostgreSQL durable-subscription table as a raw-SQL schema sibling, with explicit initialized state, acknowledged position, phase, active failure position/count/reason/next_attempt_at_utc, last-halt evidence, owner id/generation and lease expiry. Support explicit owner provisioning and compatibility startup provisioning from one definition; a preprovisioned runtime performs zero DDL on state operations. A missing table or column surfaces as a named provider failure from every state operation and is never reported as an absent or uninitialized subscription. No EF migration, existing-row rewrite or backfill.
- AC3: Implement atomic InitializeOrGetAsync. For FromNow, one PostgreSQL statement reads the service-scoped MAX(sortable_unique_id) and inserts CASE WHEN max_position IS NULL THEN NULL ELSE LEAST(max_position, safe_threshold) END with initialized=TRUE; unguarded LEAST and a separate latest-read/insert are forbidden. FromBeginning starts at NULL. Conflict returns the existing durable row and never replaces its cursor.
- AC4: Treat stored events as authoritative. Attach the Orleans nudge subscription before the first catch-up read, then read events strictly after the durable cursor in SortableUniqueId order on every nudge and fallback poll. Notification payloads never invoke the handler directly, and acknowledgements advance only after successful conversion and handling.
- AC5: Fence every post-acquire mutation by service, subscription, owner id and owner generation. PostgreSQL now() alone decides acquisition, expiry, renewal and retry timing. Halted rows are not acquirable; each acquisition attempt is preceded by a row read; non-owners remain Standby and periodically reread. Renew during long handlers, cancel cooperatively on fence loss, reject late acknowledgement and permit bounded reacquisition.
- AC6: Conversion failure durably halts on the first attempt. Ordinary handler failures durably record counts 1 through 3 with next_attempt_at_utc and no acknowledgement; count 4 atomically halts. A restart/takeover preserves the count and waits for the database timestamp. Success clears active failure state. Halt clears owner/lease and stops renewal while preserving the cursor and last-halt evidence. Resume applies only to a Halted row and is a no-op otherwise; it preserves cursor/last-halt evidence, clears active retry and owner/lease, increments generation, enters CatchingUp and retries the same position.
- AC7: Reject duplicate durable declarations with the same name inside one service container before hosted services start. Allow the same durable identity across processes for lease-based HA takeover. Document that legacy callbacks cannot be reserved against a durable name and may independently invoke application handling, while only the durable runner mutates its cursor.
- AC8: Add net9/net10 unit coverage and real PostgreSQL integration tests that discriminate guarded empty/safe-tail initialization, restart/offline catch-up, lost/duplicate/out-of-order nudges, crash ambiguity, retry timing/count persistence, halt/resume, PostgreSQL-clock fencing, long-handler lease loss, service isolation, provisioning modes, standby/takeover, duplicate registration and lifecycle cancellation. The following mutants must fail: unguarded LEAST/raw MAX initialization, handling the notification payload instead of reading storage, client-clock ownership, a stale generation writing after fencing, retry-count reset on restart or a nudge-driven hot retry loop, acquiring a Halted row, retaining owner/lease on resume, a halted owner that keeps renewing, a Halted runner that never re-reads, skipping or advancing the cursor on a poison event, non-deterministic truncation of a halt reason, and DDL executed by a preprovisioned state operation.
- AC9: Update EN/JA subscription and PostgreSQL deployment guidance with provisioning, start policy, safe-window overlap, at-least-once/idempotence, retry/halt/resume, database-clock leases, owner/standby, service identity, legacy overlap, Orleans 10.3.1 cluster prerequisite and the later notification-relay boundary. A diff-scope gate proves that no file under pure/, Samples/ or templates/ changed, and that no existing public Sekiban signature was removed. Applicable CI, git diff --check and exact-head intent-cli Opus5 review pass before merge. No release in this packet.

## Verification

Run the approved proof plan on net9/net10 where applicable and use a real supported PostgreSQL instance for
SQL initialization, provisioning, isolation, clock, fencing, retry and resume claims. Include mutants for
unguarded NULL handling, notification-payload handling, client-clock ownership, stale generation, retry
reset/hot loop, acquisition of Halted rows, retained lease on resume, renewal after halt, missing Halted-row
rereads, poison cursor skip/advance, non-deterministic 2048-character reason truncation and DDL from state operations. Run the
existing Orleans compatibility suites, applicable CI and `git diff --check`. Review the exact PR head through
intent-cli; every finding must include a concrete correction and a discriminating test.

## Related Links

- `intents/sekiban/design/sek-g75-durable-subscriber-code-plan-2026-09-11.md`
- `intents/sekiban/design/issue-1185-recovery-requirement-2026-09-09.md`
- `intents/sekiban/design/dcb-10-22-orleans-upgrade-2026-09-10.md`

## Knowledge Maintenance

- Intent placement: first approved #1185 recovery slice after the Orleans 10.3.1 prerequisite.
- ADR candidate: none; keep the bounded decision in the approved Design record.
- Diagram candidate: none.
- Docs update: EN/JA durable subscription and PostgreSQL deployment guidance.
- Closeout writeback: record final contracts and proof in the recovery design and intent tree.

## Guide Reachability

No new intent-cli role-facing surface is added.

## Base Branch Policy

Policy: `direct-main`

Open against `main`. Use intent-cli lifecycle transitions; child work does not read or mutate host packet,
queue, intent-tree or review-runtime state.

tomohisa:
Closes #1225

## Summary

- Adds the opt-in `Sekiban.Dcb.Subscriptions` contracts, durable runner, nudge boundary, and public handle.
- Adds PostgreSQL raw-SQL durable state provisioning and DML-only pre-provisioned runtime operations with service/name, owner/generation, lease, retry, halt, and resume fencing.
- Keeps `IEventSubscription` unchanged; Orleans notifications are data-free wake hints and PostgreSQL is authoritative.
- Documents deployment, safe-window, retry/halt/resume, idempotence, database-clock/fencing, Orleans 10.3.1, and later relay-boundary limits in EN/JA.

## Validation

- Real PostgreSQL Testcontainers focused suite: 10 passed, 0 failed on net9.0 and net10.0.
- Covered initialization (FromNow/FromBeginning and empty history), pre-provisioned missing table `42P01`, missing column `42703`, lease/renew/takeover/fencing, retry/halt/resume, service isolation, distinct DI runners, nudge ordering, and data-free wake behavior.
- `Sekiban.Dcb.Orleans.Tests` builds passed on net9.0 and net10.0.
- Legacy consumer fixture builds passed on net9.0 and net10.0.
- `git diff --check` passed.

The focused build/test commands use the repository's existing package graph and leave only existing dependency/obsolete warnings; no release or package-version changes are included.

### DIFF
--- a/dcb/src/Sekiban.Dcb.Core/Subscriptions/DurableSubscriptionContracts.cs
+++ b/dcb/src/Sekiban.Dcb.Core/Subscriptions/DurableSubscriptionContracts.cs
@@ -29,8 +29,17 @@ public enum DurableSubscriptionPhase
     CatchingUp,
     Idle,
     Retrying,
-    Standby,
     Halted
+}
+
+/// <summary>Process-local runner status; this is never persisted as durable subscription state.</summary>
+public enum DurableSubscriptionRunnerStatus
+{
+    Starting,
+    Owner,
+    Standby,
+    Halted,
+    Stopped
 }
 
 /// <summary>Normalized service/name identity for a durable subscriber.</summary>
@@ -132,7 +141,12 @@ public sealed record DurableSubscriptionState(
 /// <summary>Lease acquisition result; non-owners receive a state instead of a false healthy result.</summary>
 public sealed record DurableSubscriptionLease(
     bool Acquired,
-    DurableSubscriptionState State);
+    DurableSubscriptionState State)
+{
+    /// <summary>Process-local ownership status associated with this acquisition attempt.</summary>
+    public DurableSubscriptionRunnerStatus Status { get; init; } =
+        Acquired ? DurableSubscriptionRunnerStatus.Owner : DurableSubscriptionRunnerStatus.Standby;
+}
 
 /// <summary>Provider-owned state boundary for a durable subscription.</summary>
 public interface IDurableSubscriptionStore
@@ -160,6 +174,12 @@ public interface IDurableSubscriptionStore
         TimeSpan leaseDuration,
         CancellationToken cancellationToken = default);
 
+    Task<ResultBox<bool>> IsRetryDueAsync(
+        DurableSubscriptionIdentity identity,
+        string ownerId,
+        long ownerGeneration,
+        CancellationToken cancellationToken = default);
+
     Task<ResultBox<DurableSubscriptionState>> AcknowledgeAsync(
         DurableSubscriptionIdentity identity,
         string ownerId,
@@ -184,6 +204,8 @@ public interface IDurableSubscriptionStore
 
     Task<ResultBox<DurableSubscriptionState>> HaltAsync(
         DurableSubscriptionIdentity identity,
+        string ownerId,
+        long ownerGeneration,
         string reason,
         CancellationToken cancellationToken = default);
 
@@ -217,6 +239,7 @@ public sealed record DurableSubscriptionRegistration(
 public interface IDurableSubscriptionHandle : IAsyncDisposable
 {
     DurableSubscriptionIdentity Identity { get; }
+    DurableSubscriptionRunnerStatus Status { get; }
     Task<DurableSubscriptionState> GetStateAsync(CancellationToken cancellationToken = default);
     Task ResumeAsync(CancellationToken cancellationToken = default);
     Task HaltAsync(string reason, CancellationToken cancellationToken = default);
--- a/dcb/src/Sekiban.Dcb.Core/Subscriptions/DurableSubscriptionContracts.cs
+++ b/dcb/src/Sekiban.Dcb.Core/Subscriptions/DurableSubscriptionContracts.cs
@@ -1,0 +1,223 @@
+using Sekiban.Dcb.Common;
+using Sekiban.Dcb.Events;
+using Sekiban.Dcb.ServiceId;
+using ResultBoxes;
+
+namespace Sekiban.Dcb.Subscriptions;
+
+/// <summary>How the first durable cursor is selected.</summary>
+public enum DurableSubscriptionStartPolicy
+{
+    /// <summary>Start at the current safe tail and process only events written afterwards.</summary>
+    FromNow,
+
+    /// <summary>Start at the beginning of the service event stream.</summary>
+    FromBeginning
+}
+
+/// <summary>Controls whether the provider may create its durable state table at runtime.</summary>
+public enum DurableSubscriptionProvisioningMode
+{
+    LegacyAutoProvision,
+    PreProvisioned
+}
+
+/// <summary>Durable lifecycle phases persisted by the provider.</summary>
+public enum DurableSubscriptionPhase
+{
+    Uninitialized,
+    CatchingUp,
+    Idle,
+    Retrying,
+    Standby,
+    Halted
+}
+
+/// <summary>Normalized service/name identity for a durable subscriber.</summary>
+public sealed record DurableSubscriptionIdentity
+{
+    public DurableSubscriptionIdentity(string serviceId, string name)
+    {
+        ServiceId = ServiceIdValidator.NormalizeAndValidate(serviceId);
+        if (string.IsNullOrWhiteSpace(name))
+        {
+            throw new ArgumentException("Durable subscription name must not be empty.", nameof(name));
+        }
+
+        if (name.Length > 128)
+        {
+            throw new ArgumentException("Durable subscription name must be at most 128 characters.", nameof(name));
+        }
+
+        Name = name.Trim();
+        if (Name.Length == 0 || Name.Any(char.IsControl))
+        {
+            throw new ArgumentException("Durable subscription name contains invalid characters.", nameof(name));
+        }
+    }
+
+    public string ServiceId { get; }
+    public string Name { get; }
+}
+
+/// <summary>Validated bounds and timing for a durable subscriber.</summary>
+public sealed class DurableSubscriptionOptions
+{
+    public string ServiceId { get; set; } = DefaultServiceIdProvider.DefaultServiceId;
+    public string Name { get; set; } = "subscription";
+    public DurableSubscriptionStartPolicy StartPolicy { get; set; } = DurableSubscriptionStartPolicy.FromNow;
+    public DurableSubscriptionProvisioningMode ProvisioningMode { get; set; } = DurableSubscriptionProvisioningMode.LegacyAutoProvision;
+    public TimeSpan SafeWindow { get; set; } = TimeSpan.FromMilliseconds(SortableUniqueId.SafeMilliseconds);
+    public TimeSpan LeaseDuration { get; set; } = TimeSpan.FromSeconds(30);
+    public TimeSpan PollInterval { get; set; } = TimeSpan.FromSeconds(5);
+    public TimeSpan RetryDelay { get; set; } = TimeSpan.FromSeconds(5);
+    public int MaxHandlerAttempts { get; set; } = 4;
+    public int MaxBatchSize { get; set; } = 256;
+
+    public DurableSubscriptionIdentity Identity => new(ServiceId, Name);
+
+    public void Validate()
+    {
+        _ = Identity;
+        if (SafeWindow < TimeSpan.Zero || SafeWindow > TimeSpan.FromDays(7))
+        {
+            throw new ArgumentOutOfRangeException(nameof(SafeWindow), "SafeWindow must be between zero and seven days.");
+        }
+
+        if (LeaseDuration < TimeSpan.FromSeconds(1) || LeaseDuration > TimeSpan.FromHours(1))
+        {
+            throw new ArgumentOutOfRangeException(nameof(LeaseDuration), "LeaseDuration must be between one second and one hour.");
+        }
+
+        if (PollInterval <= TimeSpan.Zero || PollInterval > TimeSpan.FromHours(1))
+        {
+            throw new ArgumentOutOfRangeException(nameof(PollInterval), "PollInterval must be positive and at most one hour.");
+        }
+
+        if (RetryDelay <= TimeSpan.Zero || RetryDelay > TimeSpan.FromHours(1))
+        {
+            throw new ArgumentOutOfRangeException(nameof(RetryDelay), "RetryDelay must be positive and at most one hour.");
+        }
+
+        if (MaxHandlerAttempts < 1 || MaxHandlerAttempts > 16)
+        {
+            throw new ArgumentOutOfRangeException(nameof(MaxHandlerAttempts), "MaxHandlerAttempts must be between one and sixteen.");
+        }
+
+        if (MaxBatchSize < 1 || MaxBatchSize > 10_000)
+        {
+            throw new ArgumentOutOfRangeException(nameof(MaxBatchSize), "MaxBatchSize must be between one and 10,000.");
+        }
+    }
+}
+
+/// <summary>Durable state returned by a provider.</summary>
+public sealed record DurableSubscriptionState(
+    DurableSubscriptionIdentity Identity,
+    bool Initialized,
+    string? AcknowledgedPosition,
+    DurableSubscriptionPhase Phase,
+    string? ActiveFailurePosition,
+    int ActiveFailureCount,
+    string? ActiveFailureReason,
+    DateTimeOffset? NextAttemptAtUtc,
+    string? LastHaltPosition,
+    DateTimeOffset? LastHaltAtUtc,
+    string? LastHaltReason,
+    string? OwnerId,
+    long OwnerGeneration,
+    DateTimeOffset? LeaseExpiresAtUtc,
+    DateTimeOffset UpdatedAtUtc);
+
+/// <summary>Lease acquisition result; non-owners receive a state instead of a false healthy result.</summary>
+public sealed record DurableSubscriptionLease(
+    bool Acquired,
+    DurableSubscriptionState State);
+
+/// <summary>Provider-owned state boundary for a durable subscription.</summary>
+public interface IDurableSubscriptionStore
+{
+    Task<ResultBox<DurableSubscriptionState>> InitializeOrGetAsync(
+        DurableSubscriptionIdentity identity,
+        DurableSubscriptionStartPolicy startPolicy,
+        string safeTail,
+        CancellationToken cancellationToken = default);
+
+    Task<ResultBox<DurableSubscriptionState>> ReadAsync(
+        DurableSubscriptionIdentity identity,
+        CancellationToken cancellationToken = default);
+
+    Task<ResultBox<DurableSubscriptionLease>> TryAcquireAsync(
+        DurableSubscriptionIdentity identity,
+        string ownerId,
+        TimeSpan leaseDuration,
+        CancellationToken cancellationToken = default);
+
+    Task<ResultBox<bool>> RenewAsync(
+        DurableSubscriptionIdentity identity,
+        string ownerId,
+        long ownerGeneration,
+        TimeSpan leaseDuration,
+        CancellationToken cancellationToken = default);
+
+    Task<ResultBox<DurableSubscriptionState>> AcknowledgeAsync(
+        DurableSubscriptionIdentity identity,
+        string ownerId,
+        long ownerGeneration,
+        string position,
+        CancellationToken cancellationToken = default);
+
+    Task<ResultBox<DurableSubscriptionState>> RecordFailureAsync(
+        DurableSubscriptionIdentity identity,
+        string ownerId,
+        long ownerGeneration,
+        string position,
+        string reason,
+        bool conversionFailure,
+        int maxHandlerAttempts,
+        TimeSpan retryDelay,
+        CancellationToken cancellationToken = default);
+
+    Task<ResultBox<DurableSubscriptionState>> ResumeAsync(
+        DurableSubscriptionIdentity identity,
+        CancellationToken cancellationToken = default);
+
+    Task<ResultBox<DurableSubscriptionState>> HaltAsync(
+        DurableSubscriptionIdentity identity,
+        string reason,
+        CancellationToken cancellationToken = default);
+
+    Task<ResultBox<DurableSubscriptionState>> MarkIdleAsync(
+        DurableSubscriptionIdentity identity,
+        string ownerId,
+        long ownerGeneration,
+        CancellationToken cancellationToken = default);
+}
+
+/// <summary>Orleans or another delivery system can wake a durable runner; it never supplies authoritative data.</summary>
+public interface IDurableSubscriptionNudgeFactory
+{
+    IDurableSubscriptionNudge Create(DurableSubscriptionIdentity identity, Func<ValueTask> onNudge);
+}
+
+/// <summary>Disposable best-effort wake subscription.</summary>
+public interface IDurableSubscriptionNudge : IAsyncDisposable
+{
+}
+
+/// <summary>Delegate used by a durable runner after storage conversion succeeds.</summary>
+public delegate Task DurableSubscriptionHandler(Event @event, CancellationToken cancellationToken);
+
+/// <summary>Registration object used by provider DI extensions.</summary>
+public sealed record DurableSubscriptionRegistration(
+    DurableSubscriptionOptions Options,
+    DurableSubscriptionHandler Handler);
+
+/// <summary>Public handle for an opt-in durable runner.</summary>
+public interface IDurableSubscriptionHandle : IAsyncDisposable
+{
+    DurableSubscriptionIdentity Identity { get; }
+    Task<DurableSubscriptionState> GetStateAsync(CancellationToken cancellationToken = default);
+    Task ResumeAsync(CancellationToken cancellationToken = default);
+    Task HaltAsync(string reason, CancellationToken cancellationToken = default);
+}
--- a/dcb/src/Sekiban.Dcb.Core/Subscriptions/DurableSubscriptionRunner.cs
+++ b/dcb/src/Sekiban.Dcb.Core/Subscriptions/DurableSubscriptionRunner.cs
@@ -1,0 +1,476 @@
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using ResultBoxes;
+using Sekiban.Dcb.Common;
+using Sekiban.Dcb.Domains;
+using Sekiban.Dcb.Events;
+using Sekiban.Dcb.Storage;
+
+namespace Sekiban.Dcb.Subscriptions;
+
+/// <summary>
+///     Store-driven durable subscriber. Orleans (or another delivery mechanism) only releases the wake gate;
+///     every event and acknowledgement comes from the authoritative event store and durable state store.
+/// </summary>
+public sealed class DurableSubscriptionRunner : BackgroundService
+{
+    private readonly DurableSubscriptionRegistration _registration;
+    private readonly IDurableSubscriptionStore _stateStore;
+    private readonly IEventStoreFactory _eventStoreFactory;
+    private readonly IEventTypes _eventTypes;
+    private readonly IDurableSubscriptionNudgeFactory _nudgeFactory;
+    private readonly ILogger<DurableSubscriptionRunner> _logger;
+    private readonly SemaphoreSlim _wake = new(0, 1);
+    private readonly string _ownerId = $"{Environment.MachineName}:{Environment.ProcessId}:{Guid.NewGuid():N}";
+    private IDurableSubscriptionNudge? _nudge;
+
+    public DurableSubscriptionRunner(
+        DurableSubscriptionRegistration registration,
+        IDurableSubscriptionStore stateStore,
+        IEventStoreFactory eventStoreFactory,
+        IEventTypes eventTypes,
+        IDurableSubscriptionNudgeFactory nudgeFactory,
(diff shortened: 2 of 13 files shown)


tomohisa (commented):
## Exact-head Opus 5 review — `2141bd80016b5371bf2220bebcd4525761df022c`

**REQUEST-UPDATE.** Reviewed in a fresh checkout detached at the exact SHA, against issue #1225 as published on GitHub (no host packet files used).

**Two independent reasons.** First, **checks are pending**: `dcbTestsNet9` and `dcbTestsNet10` are both `in_progress`, and they are the only jobs that run `Sekiban.Dcb.Postgres.Tests` where every test in this PR lives — so there is no net9/net10 or real-PostgreSQL evidence at this head. The one completed check (`Pack and run PostgreSQL consumers`) does not touch the durable subscriber, and I am not counting it. Second, six contract violations in the code, four of which are behaviours AC8 names as **mutants that must fail** — they are shipped instead.

The shape is right: additive-only (2,212 insertions, the only edit to an existing file is a two-line registration hook), event-store-authoritative, nudge attached before first catch-up, raw-SQL schema sibling, `CURRENT_TIMESTAMP` in most statements, `pure/`/`Samples/`/`templates/` untouched (0 files), `git diff --check` clean, `dcb_events` column names verified against the migration. The defects cluster in the lease/clock boundary and the ownership loop.

### F1 — the client clock decides ownership (AC5; AC8 mutant "client-clock ownership")
`TryAcquireAsync` gates on `expiry <= DateTimeOffset.UtcNow` before reaching the DB predicate; `Owns()` does the same. The `UPDATE` does use `lease_expires_at_utc <= CURRENT_TIMESTAMP`, so no lease can be *stolen* early — but a client clock behind the database returns `Standby` even when the DB would grant the lease, so takeover is delayed by the full skew. The approved plan's ±30 s skew test fails.
**Fix:** drop the pre-check and let `UPDATE … RETURNING` decide; if a pre-read is kept for AC5's read-before-acquire, select `CURRENT_TIMESTAMP` in the same `SELECT … FOR UPDATE` and compare to that.
**Test:** two runners, A's lease expired by direct SQL, B skewed −30 s then +30 s; assert identical acquisition and generation. Today the −30 s run stalls in `Standby`.

### F2 — `HaltAsync` is unfenced and destroys halt evidence (AC5, AC6; AC8 mutant "stale generation writing after fencing")
Its only predicate is `service_id AND subscription_name` — no owner, no generation, no lease, no phase. So a runner fenced out generations ago can halt a **healthy live owner**, null its lease mid-handler, and force an operator resume. Halting an already-`Halted` row also overwrites `last_halt_reason`/`last_halt_at_utc`, erasing the original poison evidence AC6 requires preserved, and it never writes `last_halt_position`.
**Fix:** same fence as every sibling (`owner_id`, `owner_generation`, `phase <> 'Halted'`, live lease) and set `last_halt_position`. If an unfenced operator halt is really wanted it needs its own named method and an explicit contract change.
**Test:** owner A live at generation 2; a handle at generation 1 calls Halt → must report fence loss, row still `CatchingUp`, A's next ack succeeds. Then halt twice with different reasons → `last_halt_*` must still describe the first halt. Both fail today.

### F3 — an owner cannot re-acquire its own live lease, so nudges and polls are discarded for a full lease period (AC4)
On an empty batch the runner calls `MarkIdleAsync` (which leaves `owner_id` and `lease_expires_at_utc` intact) and **returns**. The next cycle calls `TryAcquireAsync`, which has no same-owner branch, so the runner's own unexpired lease makes it ineligible and it gets `Standby`. With the shipped defaults (lease 30 s, poll 5 s) a nudge is discarded up to six times per lease, effective catch-up latency becomes **`LeaseDuration`, not `PollInterval`**, and `owner_generation` increments once per lease forever on an idle single-owner system.
**Fix:** keep ownership across cycles — wait for wake/poll *inside* the owned loop and renew on that cadence (closest to the approved design) — or make `TryAcquireAsync` re-entrant for the same `owner_id` without bumping the generation.
**Test:** one runner, lease 30 s, poll 1 s; go idle, commit an event, fire a nudge; assert the handler runs within ~2 s and the generation is unchanged. Today it waits up to 30 s.

### F4 — durable phase and runner status are one enum, and `Standby` is written into durable state (AC1)
A single `DurableSubscriptionPhase { Uninitialized, CatchingUp, Idle, Retrying, Standby, Halted }` is documented as "persisted by the provider", yet `Standby` is never persisted — the store fabricates it with `state.GetValue() with { Phase = Standby }`, replacing the row's real phase in the returned record. Any inspection surface (the operator path AC6 exists for) then reports a phase the database does not contain.
**Fix:** keep `DurableSubscriptionPhase` to persisted values and add a runner-local status on the lease/handle. Never rewrite `Phase` on a record that claims to be provider state.
**Test:** loser of an acquisition race — `GetStateAsync()` must match the row read back by direct SQL while the runner status reports `Standby`.

### F5 — `next_attempt_at_utc` is recorded but never enforced by the database (AC5 retry timing, AC6)
No statement uses it as a predicate; the only gate is `retryAt > DateTimeOffset.UtcNow` plus `DelayUntilAsync` on the client clock. A host whose clock runs ahead re-invokes before the durable timestamp and burns a retry early — the storage half of the durable-spacing fix shipped, the enforcement half did not.
**Fix:** gate the statements that consume a retry on `next_attempt_at_utc <= CURRENT_TIMESTAMP`, and return the database `CURRENT_TIMESTAMP` so the client waits on a DB-derived delta.
**Test:** failure count 1 at P with a 60 s delay, second runner skewed +120 s → handler must not be invoked and the count must stay 1. Today it invokes immediately and reaches 2.

### F6 — `PreProvisioned` can be silently downgraded to auto-provisioning DDL (AC2; AC8 mutant "DDL executed by a preprovisioned state operation")
The store is registered with `TryAddSingleton`, so the first registration wins. Register a default subscription first and a `PreProvisioned` one second — an ordinary composition — and the second silently runs with `legacyAutoProvision: true`, executing `CREATE TABLE/INDEX IF NOT EXISTS` before every state operation on a deployment whose role may be DML-only.
**Fix:** make provisioning one container-level decision; validate that all registrations agree and throw a named configuration error otherwise. Don't let `TryAdd` order decide a safety property.
**Test:** register A (default) then B (`PreProvisioned`) against a role without `CREATE` and a pre-created table; B's read must succeed with no DDL. Today it fails on the `CREATE` attempt, and reversing the registration order changes the behaviour.

### F7 — AC8's matrix is largely unwritten
Nine `[Fact]`s total, one of them a runner test. Missing: restart/offline catch-up, lost nudge, duplicate and out-of-order nudges, crash ambiguity, retry timing/count persistence across restart, long-handler lease loss, lifecycle cancellation, a `Halted` runner that re-reads, a halted owner that keeps renewing, poison skip/advance. At most four of the twelve named mutants are discriminated — which is why F1–F6 survived.

### Smaller items for the repair pass
- `RecordFailureAsync` increments the count unconditionally; AC6 requires a reset to 1 when `active_failure_position` differs (`CASE WHEN … IS DISTINCT FROM …`). No path reaches it today, but the statement is non-conformant.
- AC3's conflict path uses `ON CONFLICT … DO NOTHING` plus a follow-up read *outside* any transaction; the contract allows an "equivalent same-transaction select". Benign today, free to fix.
- `NoOpDurableSubscriptionNudgeFactory` is the registered default, so forgetting the Orleans nudge registration degrades to poll-only with no diagnostic.
- `IDurableSubscriptionHandle` is registered once per subscription under the same service type, so `GetRequiredService<IDurableSubscriptionHandle>()` silently returns the last one.

Head verified unchanged at `2141bd80016b5371bf2220bebcd4525761df022c` before and after the review. No label, merge, release or source mutation was made.

tomohisa (commented):
## Exact-head Opus 5 re-review — `9127f895033f08fc849a9533f7765ee34f94502f`

**REQUEST-UPDATE**, now down to two items. Fresh checkout detached at the exact SHA; issue #1225 re-read from GitHub; the `2141bd80` verdict was **not** carried over — every finding below was re-derived from this head's code.

**F1–F6 are closed**, and I checked the code rather than the commit message. The clearest confirmation is mechanical: `grep -rn 'DateTimeOffset.UtcNow\|DateTime.UtcNow'` across all three new source directories returns **exactly one hit** — the safe-tail value computation — so no client clock remains in acquisition, expiry, renewal or retry timing.

- **F1** The `expiry <= DateTimeOffset.UtcNow` pre-check is deleted, `Owns()` and `DelayUntilAsync` are deleted outright, and the `UPDATE … RETURNING` predicate is the sole gate. The remaining `UtcNow` is the AC3 caller-supplied `safeThreshold`, capped by `LEAST(max_position, @safe_tail)`, so it can never push the cursor past the committed maximum. The skew test isn't needed any more — the property holds by construction.
- **F2** `HaltAsync` takes owner/generation and carries the full fence including `phase <> @halted`, which is what preserves the first halt evidence, and it now records `last_halt_position = COALESCE(active_failure_position, acknowledged_position)`. `FencedHaltPreservesFirstEvidenceAndPosition` is exactly the discriminator I specified, down to the second-halt-must-not-overwrite assertion.
- **F3** Acquire is re-entrant for the same owner and increments the generation only when the owner actually changes; it also preserves a `Retrying` phase across re-entry, which matters — without that, re-entry would silently drop the retry gate. The owned loop now renews and `continue`s on an empty batch instead of returning. `IdleDuplicate…` asserts `after.OwnerGeneration == before.OwnerGeneration` and `calls == 1` across duplicate nudges.
- **F4** `Standby` is out of `DurableSubscriptionPhase`; the new `DurableSubscriptionRunnerStatus` is documented as never persisted and surfaced on the lease, the runner and the handle; the `state with { Phase = Standby }` rewrite is gone. The clock test asserts `Status == Standby` **and** `State.Phase == CatchingUp` — the two values that used to be conflated.
- **F5** Three DB enforcement points now exist: the new `IsRetryDueAsync`, the acknowledge predicate, and a `SELECT … FOR UPDATE` due-check inside `RecordFailureAsync`'s transaction. The runner waits on the ordinary wake/poll gate instead of a client deadline.
- **F6** Mixed provisioning modes now throw a named error, and the store resolves the mode from a registered singleton rather than a closure, so `TryAdd` ordering can't decide a safety property. Tested in both registration orders.
- Also closed: AC3's conflict path is now the prescribed no-op `DO UPDATE … RETURNING` (single statement, NULL guard intact), and `active_failure_count` now resets to 1 via `active_failure_position IS DISTINCT FROM @position`.

### Still blocking

**Required CI is pending.** `dcbTestsNet9` and `dcbTestsNet10` are both `in_progress`, and they are the only jobs that run `Sekiban.Dcb.Postgres.Tests` where every test here lives — so there is still no net9/net10 or real-PostgreSQL evidence at this head. `Pack and run PostgreSQL consumers` is green but does not touch the durable subscriber, and I'm not counting it.

### F7-R — the nudge path is not discriminated, and two AC8 cases are untested (AC8)

1. **The nudge assertions cannot fail when the nudge is broken.** `CreateRunner` sets `PollInterval = 25 ms`, and `IdleDuplicateAndOutOfOrderNudgesReuseLeaseAndCatchUpPromptly` awaits the second event with a 3 s timeout — about 120 poll cycles. Delete every nudge and the test still passes. The same holds for `RunnerAttachesNudgeBeforeReadingAndAcknowledgesOnlyAfterHandler`. So the area the suite appears to cover most directly is masked by the poll.
2. **Two named discriminators have no test at all**: *lost nudge* (commit an event, fire no nudge, prove poll-driven catch-up) and *crash ambiguity* (handler completes, stop before acknowledgement, restart re-invokes the same event — the at-least-once boundary the package publishes).

**Fix:** in the nudge tests set `PollInterval` far longer than the assertion window (e.g. 60 s) and keep the 3 s wait, so only a nudge can satisfy it; keep the 25 ms poll for the ownership/retry tests. Add `LostNudgeIsCaughtUpByPollAlone` (`PollInterval = 250 ms`, no trigger). Add `CrashBetweenHandlerAndAcknowledgementReinvokesTheSameEvent` (cancel before ack, or a store decorator that swallows the first acknowledge; restart and assert re-invocation with the cursor advancing once).

**Test:** the first bullet is its own discriminator — with a 60 s poll, a build whose nudge never fires times out where today it passes; a nudge-only mutant fails the lost-nudge test; an acknowledge-before-handler mutant fails the crash-ambiguity test.

### Notes (not findings)
- The safe tail is client-computed **by contract** (AC3 passes it in). A backward skew only widens re-delivery; a forward skew degrades to unguarded behaviour but cannot exceed `max_position`. One doc sentence would be enough.
- `handle.HaltAsync` now requires local ownership, so an operator cannot halt a subscription owned by another process. Consistent with AC5 and the design, but worth documenting.
- `IsRetryDueAsync` returns `false` both for "not due" and "row does not match"; the runner recovers correctly, but separating them would make the operator diagnostic clearer.
- `IDurableSubscriptionHandle` is still registered once per subscription under one service type, so `GetRequiredService<IDurableSubscriptionHandle>()` returns the last one.

Head verified unchanged at `9127f895033f08fc849a9533f7765ee34f94502f` before and after. No label, merge, release or source mutation.

tomohisa (commented):
## Final exact-head Opus 5 review — `305304eac3780461db493822f7f4282d80721177`

**REQUEST-UPDATE — one finding, two sentences wide, documentation only.** Fresh checkout detached at the exact SHA; issue #1225 re-read from GitHub; no verdict carried over from the earlier heads.

Everything else is complete and, for the first time in this PR, **fully evidenced**. All required checks are green at this head, and I read the net9 and net10 DCB job logs rather than the badges: **20 durable-subscription tests, the identical set, passed on both TFMs** against real PostgreSQL. `Sekiban.Dcb.Postgres.Tests` reports `Total tests: 98, Passed: 98` in each job (78 at the SEK-G74 baseline + the 20 added here), every other DCB suite is unchanged and green, and the packaged PostgreSQL consumer and both SonarCloud checks are green.

### F7-R — closed; all three required properties verified

The diff from `9127f895` is test-only (`git diff --name-only 9127f895 305304ea -- dcb/src` → 0 files).

- **Nudge tests now depend on the nudge.** `CreateRunner` gained a `pollInterval` parameter and the two nudge tests opt into a **60 s** poll with short assertion windows. `IdleDuplicateAndOutOfOrderNudgesReuseLeaseAndCatchUpPromptly` appends the second event only after the runner reaches `Idle`, then nudges and waits 3 s; the duplicate-nudge observation gained an explicit 3 s bound. `DurableSubscriptionRunnerTests` went from a 10 ms poll to 60 s, so its post-nudge `ReadCount` assertion is genuinely nudge-driven. Remove the nudge callback or observer and both now time out — at `9127f895` the 25 ms poll satisfied them ~120 times over.
- **Lost nudge proven by poll alone.** `LostNudgeFallsBackToFinitePollingAndAcknowledges` uses a 250 ms poll, waits for `Idle`, appends, never triggers, and asserts handling, acknowledgement, **and `nudge.TriggerCount == 0`** — proving no nudge fired rather than merely that the test didn't call one.
- **Crash-before-first-ack.** `FirstAcknowledgeBlockedStore` blocks the first `AcknowledgeAsync` indefinitely; the test asserts the handler ran with the cursor still null, cancels the runner, re-asserts null after the "crash", expires the lease by SQL, starts a second runner, and asserts the **same position** is re-invoked, the cursor then equals it, and `SuccessfulAcknowledgementCount == 1`. A pre-handler-ack mutant fails it: the decorator would block before the handler ever ran, so the 5 s wait on `firstHandled` would fail. All waits are bounded (`WaitUntilAsync` asserts at a 5 s deadline), so nothing can hang instead of failing.

### F1–F6 re-derived at this head

`grep -rn 'DateTimeOffset.UtcNow\|DateTime.UtcNow'` across all three new source directories returns **one** hit — the AC3 caller-supplied safe tail, capped by `LEAST(max_position, @safe_tail)`. `HaltAsync` carries the full fence plus `phase <> 'Halted'` and records `last_halt_position`. Acquire is re-entrant with the generation bumped only on owner change and `Retrying` preserved. `DurableSubscriptionRunnerStatus` is separate and never persisted. `next_attempt_at_utc` is enforced at three DB points. Mixed provisioning fails closed in both orders. Each has a named green test.

### Scope and API
`pure/`/`Samples/`/`templates/` → 0 files changed; `git diff --numstat … -- dcb/src` shows **0 deleted lines** (purely additive; the only edit to a pre-existing file is still the two-line `TryAddSingleton` hook); `git diff --check` clean.

### F8 — the AC7 legacy-overlap disclosure is missing from both guides

**Clause.** AC7: "Document that legacy callbacks cannot be reserved against a durable name and may independently invoke application handling, while only the durable runner mutates its cursor." AC9 repeats it as "legacy overlap".

**Defect.** `docs/dcb_llm/23_durable_orleans_subscriptions.md:29` (and the JA line 29) say only that the legacy path is "unchanged and is not converted into a durable subscription" — a different statement. None of AC7's three propositions appears in either file. An adopter would reasonably assume that registering a durable subscription named `billing` reserves that name, which is precisely the assumption the packet ruled out as unenforceable and made a disclosure clause because of.

**Fix** — two sentences in the Registration section, EN and JA:

> A durable subscription name is not reserved against the legacy `IEventSubscription` callbacks: those callbacks carry no shared declaration registry and usually receive generated ids, so the library cannot detect or reject an overlapping name. If an application registers both for the same logical subscription, application handling may be invoked twice — handlers must already be idempotent — but only the durable runner reads from and advances the durable cursor; the legacy callback cannot move it.

**Test.** A docs assertion that both guides contain all three propositions — unreservable name, possible double application handling, durable-runner-only cursor mutation. It fails today and passes after the edit. The behavioural half is already covered by `ServiceIdentityIsolated_DurableRegistrationRejectsOnlyInOneContainer`; this is only about disclosing the limit.

Neither file is in any workflow path filter, so this repair does not invalidate the green evidence above. With it applied this is an APPROVE.

Head verified unchanged at `305304eac3780461db493822f7f4282d80721177` before and after. No label, merge, release or source mutation.

tomohisa (commented):
## Final exact-head Opus 5 review — `49b6f92830661fdce8c935157fbdfda36d287920`

**APPROVE — for this exact revision only.** Fresh detached checkout, issue #1225 as the only contract source, no verdict carried over from the earlier heads.

### The change is only the two guides
`git diff --name-status 305304ea 49b6f928` returns exactly `docs/dcb_llm/23_durable_orleans_subscriptions.md` and its JA counterpart, and `git diff --name-only 305304ea 49b6f928 -- dcb` → **0 files**. Every source file and test verified at `305304ea` (F1–F7) is byte-identical here.

### F8 closed — all three AC7/AC9 propositions are expressly stated in both languages

| proposition | EN | JA |
|---|---|---|
| no reservation/rejection by durable name, **because** there is no shared declaration registry | "not reserved against legacy callbacks because the two paths have no shared declaration registry, so the library cannot detect or reject an overlapping name" | 「予約・排他されません」＋「共有の宣言レジストリがないため、…検出して拒否できません」 |
| possible **independent** double application handling, requiring idempotence | "may be invoked twice independently and handlers must already be idempotent" | 「それぞれ独立して呼び出されて二重に実行される可能性があるため、ハンドラーはあらかじめ冪等でなければなりません」 |
| durable-runner-only cursor **reads and advancement** | "only the durable runner reads and advances the durable cursor, and the legacy callback cannot move it" | 「読み取り進めるのは Durable runner だけであり、legacy コールバックがそのカーソルを進めることはありません」 |

Two things beyond literal presence: the paragraph sits immediately after "duplicate registrations are rejected before the hosted runner is added", so the durable-to-durable rejection and the legacy non-guarantee read as distinct scopes rather than a contradiction — that ordering was the failure mode I was watching for. And the JA is a full translation, not a reduction: it carries the reason clause, the idempotence obligation and the read-and-advance asymmetry.

### CI evidence — read from the logs, not the badges
All five checks green at this head: `dcbTestsNet9`, `dcbTestsNet10`, `Pack and run PostgreSQL consumers on net9/net10`, `SonarCloud`, `SonarCloud Code Analysis`.

I downloaded both DCB job logs (`103479671025`, `103479671339`) and confirmed each job's own `head_sha` is `49b6f9283` — fresh evidence at this head, not a carried-over run. In each job: **20 durable-subscription tests passed**, the identical set, including `LostNudgeFallsBackToFinitePollingAndAcknowledges`, `CrashAfterHandlerBeforeAcknowledgement_ReinvokesSamePositionAndAdvancesCursorOnce`, `IdleDuplicateAndOutOfOrderNudgesReuseLeaseAndCatchUpPromptly` and `RunnerAttachesNudgeBeforeReadingAndAcknowledgesOnlyAfterHandler`; **zero** `Failed Sekiban` lines; `Sekiban.Dcb.Postgres.Tests` at `98/98`; every other DCB suite at its baseline (1390, 276 with the one pre-existing skip, 115, 89, 222, 15, 4), all passing.

### Scope and API, re-checked
`pure/`/`Samples/`/`templates/` → 0 files; `git diff --numstat c00e37ab 49b6f928 -- dcb/src` → **0 deleted lines** (additive throughout; the only edit to a pre-existing source file is still the two-line `TryAddSingleton` hook); `git diff --check` clean; the only `UtcNow` in the new source remains the AC3 caller-supplied safe tail capped by `LEAST(max_position, @safe_tail)`; `Closes #1225` present.

### Findings
**None.**

Notes for closeout rather than conditions, unchanged from earlier rounds: the safe tail is client-computed by contract and cannot exceed `max_position`; `handle.HaltAsync` requires local ownership, so an operator cannot halt a subscription owned by another process; `IsRetryDueAsync` returns `false` both for "not due" and "row does not match", which the runner recovers from correctly; and `IDurableSubscriptionHandle` is registered once per subscription under one service type, so `GetRequiredService<IDurableSubscriptionHandle>()` returns the last registration.

Head verified unchanged at `49b6f92830661fdce8c935157fbdfda36d287920` before and after. No label, merge, release or source mutation was made; this approval is bound to this exact revision.

tomohisa:
Follow-up: this unit is included in the DCB 10.22.0 library release: https://github.com/J-Tech-Japan/Sekiban/releases/tag/dcb-v10.22.0 (tip `23a5096d`, 26 packages).

